PR #88, opened 2018-08-04
Resolved FrontPage error when running IIS Local

darkpsy3934:
This will resolve the Internal Server Error that happens when there is no FrontPage data to display.  Added a conditional check to see if data is returned, and if not, it returns a null value.

Changed Model.First() to Model.FirstOrDefault().

Model.FirstOrDefault() returns the null value if nothing is present in a lookup.  Then a simple if...else statement is checked for null.  if not null, it goes on to display things normally.  Otherwise, returns nothing.


darkpsy3934:
Nevermind, odd stuff happened.

PR #296, opened 2021-01-26
Use Liger.WebOptimizer.Sass to serve scss files

nattthebear:


adelikat (approved):

adelikat:
Ran it locally and it worked.  
I also tried changing bootstrap colors by putting something like this in the boostrap.scss file:
$theme-colors: ( "primary": #531f86  );

and it worked! This is exactly what I was looking for.   Just need to confirm we can get this deployed on the demo server and this will be good to go.

https://media.giphy.com/media/l0IyopK5bMZB6dGIo/giphy.gif

PR #297, opened 2021-01-27
Replace a tuple with a viewmodel

nattthebear:
This feels awful, but better than guessing what `Item1` and `Item2` are


adelikat:
This has been in a not done state, for awhile now, and highly outdated, closing

PR #299, opened 2021-01-28
use a constants class and a class attribute for mapping wiki modules …

adelikat:
…to view components instead of relying on matching text strings to view component class names

### DIFF
--- a/TASVideos.WikiEngine/WikiModules.cs
+++ b/TASVideos.WikiEngine/WikiModules.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 
@@ -50,17 +51,9 @@ namespace TASVideos.WikiEngine
 			.Select(fi => fi.GetRawConstantValue()?.ToString() ?? "")
 			.ToList();
 
-		public static bool TryGetModule(string name, out string moduleName)
+		public static bool IsModule(string name)
 		{
-			var result = Modules.FirstOrDefault(s => s == name);
-			if (result == null)
-			{
-				moduleName = "";
-				return false;
-			}
-
-			moduleName = result;
-			return true;
+			return Modules.Any(s => string.Equals(s, name, StringComparison.CurrentCultureIgnoreCase));
 		}
 	}
 }
--- a/TASVideos.WikiEngine/WikiModules.cs
+++ b/TASVideos.WikiEngine/WikiModules.cs
@@ -45,15 +45,15 @@ namespace TASVideos.WikiEngine
 		public const string WikiUsers = "wikiusers";
 		public const string Youtube = "youtube";
 
-		private static readonly IList<string> Modules = typeof(WikiModules)
+		private static readonly HashSet<string> Modules = typeof(WikiModules)
 			.GetFields(BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy)
 			.Where(fi => fi.IsLiteral && !fi.IsInitOnly)
 			.Select(fi => fi.GetRawConstantValue()?.ToString() ?? "")
-			.ToList();
+			.ToHashSet(StringComparer.InvariantCultureIgnoreCase);
 
 		public static bool IsModule(string name)
 		{
-			return Modules.Any(s => string.Equals(s, name, StringComparison.CurrentCultureIgnoreCase));
+			return Modules.Contains(name);
 		}
 	}
 }
--- a/TASVideos.WikiEngine/WikiModules.cs
+++ b/TASVideos.WikiEngine/WikiModules.cs
@@ -1,0 +1,66 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace TASVideos.WikiEngine
+{
+	public static class WikiModules
+	{
+		public const string Awards = "awards";
+		public const string BrokenLinks = "brokenlinks";
+		public const string ActiveTab = "activetab";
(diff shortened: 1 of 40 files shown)


TASVideos.WikiEngine/WikiModules.cs
```diff
@@ -0,0 +1,66 @@
...
+		public const string WikiUsers = "wikiusers";
+		public const string Youtube = "youtube";
+
+		private static readonly IList<string> Modules = typeof(WikiModules)
+			.GetFields(BindingFlags.Public | BindingFlags.Static | BindingFlags.FlattenHierarchy)
+			.Where(fi => fi.IsLiteral && !fi.IsInitOnly)
+			.Select(fi => fi.GetRawConstantValue()?.ToString() ?? "")
+			.ToList();
+
+		public static bool TryGetModule(string name, out string moduleName)
```
nattthebear, line 53: This function needs to be O(1)

TASVideos.WikiEngine/WikiModules.cs
```diff
@@ -0,0 +1,66 @@
...
+		{
+			var result = Modules.FirstOrDefault(s => s == name);
+			if (result == null)
+			{
+				moduleName = "";
+				return false;
+			}
+
+			moduleName = result;
```
nattthebear, line 62: Is it intended that `moduleName` is exactly equal to `name`?  What's the point of that mapping?

TASVideos.WikiEngine/WikiModules.cs
```diff
@@ -0,0 +1,66 @@
...
+			.Where(fi => fi.IsLiteral && !fi.IsInitOnly)
+			.Select(fi => fi.GetRawConstantValue()?.ToString() ?? "")
+			.ToList();
+
+		public static bool TryGetModule(string name, out string moduleName)
+		{
+			var result = Modules.FirstOrDefault(s => s == name);
+			if (result == null)
+			{
+				moduleName = "";
```
nattthebear, line 58: To be consistent with other C# APIs (like in `Dictionary`), should this return `null`?

TASVideos/TagHelpers/WikiMarkupTagHelper.cs
```diff
@@ -94,10 +96,22 @@ bool IWriterHelper.CheckCondition(string condition)
...
+			.ToDictionary(tkey => ((WikiModuleAttribute)tkey.GetCustomAttribute(typeof(WikiModuleAttribute))!).Name, tvalue => tvalue);
+
 		void IWriterHelper.RunViewComponent(TextWriter w, string name, string pp)
 		{
 			// TODO: Do we want to asyncify this entire thingy?
-			var content = _viewComponentHelper.InvokeAsync(name, new { pageData = PageData, pp }).Result;
+			var result = ViewComponents.TryGetValue(name, out Type? viewComponent);
+			if (!result)
+			{
+				throw new InvalidOperationException($"Unknown ViewComponent: {name}");
```
nattthebear, line 111: What happened in the old code without this check?

nattthebear (commented):
Looking reasonable so far.

TASVideos.WikiEngine/WikiModules.cs
adelikat: The point of the mapping is so that the public const string values can be used in the attribute, and you have that reference that shows which are in use.  

Since the ViewComponents are in an assembly we do not want referenced in the wiki assembly, something along these lines are necessary to tie them together with more than just magic strings.

The value must equal the markup in the code. The name of the const is arbituary.

 Though, it occurs to me that we need to make them case insensitive.  That is how it was before this refactor

TASVideos.WikiEngine/WikiModules.cs
adelikat: Bleh, fair enough.  I did that because then it complicated the calling code because things could be null, more syntax as usual

TASVideos/TagHelpers/WikiMarkupTagHelper.cs
adelikat: Good question, I'll check it out.  The old code was using the string overload where this one uses the Type overload

TASVideos.WikiEngine/WikiModules.cs
nattthebear: That all makes sense but it's still not explaining the two separate variables here.  What is `name` and what is `moduleName` and why are they separate variables if they always have the same value?

adelikat: moduleName is the text found in the wiki markup.  We match it up to a known wikiModule, and if there is a match, that module is executed

nattthebear: But they're the same, the two values are exactly the same!  This method checks if the value is a valid one, but it doesn't change it at all.

TASVideos/TagHelpers/WikiMarkupTagHelper.cs
nattthebear: This isn't hugely important, because theoretically bad modules would have been caught before they got here, but it might be worth digging into.

TASVideos/TagHelpers/WikiMarkupTagHelper.cs
```diff
@@ -94,10 +96,22 @@ bool IWriterHelper.CheckCondition(string condition)
 			return null;
 		}
 
+		private static readonly IDictionary<string, Type> ViewComponents = Assembly
```
nattthebear, line 99: Should this code, or some code somewhere, verify that each and every `public const string Foo = "bar"` actually has an implementer?  As it stands now, `WikiModules.IsModule` will return true if the field is defined, but that's no guarantee that any particular class has the attribute on it for that value.

PR #300, opened 2021-01-30
Refactor and clean up some wiki things

nattthebear:
* Move responsibility for computing style filters into the wiki engine's rendering code.
  `IWriterHelper` is meant for things that have to come from some external source.  I stuffed style filters there because I didn't have anywhere else to put them, but let's do better now.
* Move a bunch of things out of `Node.cs` into `NodeImplementations.cs`
  File too big.
* Move ParamHelper into the wiki engine project.
  The big picture for this one is that because it's parsing work, it should be the responsibility of the wiki engine, and because our long term goals include reworking or replacing the wiki syntax, it shouldn't be something individual module-rendering-components need to know about.  This is a little sloppy for now, but in the future I hope to switch modules over to using a much cleaner API to get their `pp` values, which can then abstract better over different `pp` formats.

### DIFF
--- a/TASVideos.Test/ViewComponents/ParamHelperTests.cs
+++ b/TASVideos.Test/ViewComponents/ParamHelperTests.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.Linq;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
-using TASVideos.ViewComponents;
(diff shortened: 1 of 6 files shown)


adelikat (approved):

adelikat:
Makes a lot of sense moving ParamHelper into the wiki project

PR #301, opened 2021-01-31
Wiki Module plumbing rework

nattthebear:
* The wiki parsing code itself now handles parsing out module name and parameters into a `Dictionary<string, string>`.
* Modelbinder-like responsibilities are now handled in the WikiMarkupTagHelper, which deserializes the parameters that individual view components need.


TASVideos.Test/ViewComponents/ParamHelperTests.cs
```diff
@@ -92,14 +109,12 @@ public void GetYear(string parameterStr, string param, int? expected)
...
+			var module = new Module(0, moduleString.Length, moduleString);
+			var actual = WikiMarkup.ConvertParameter<IList<int>>(module.Parameters.GetValueOrDefault(param));
 
 			Assert.IsNotNull(actual);
-			var actualList = actual.ToList();
-			foreach (var expectedVal in expected)
-			{
-				Assert.IsTrue(actualList.Contains(expectedVal));
-			}
+			Assert.IsTrue(actual.SequenceEqual(expected));
```
adelikat, line 117: Sequence equal only returns true if it is the same values in the same order.  The previous code, while more verbose, ensures that they can be in a different order and still qualify as passing the requirements.

TASVideos/TagHelpers/WikiMarkupTagHelper.cs
```diff
@@ -1,71 +1,126 @@
...
+				?? viewComponent.GetMethod("Invoke");
+
+			if (invokeMethod == null)
+			{
+				throw new InvalidOperationException($"Could not find an Invoke method on ViewComponent {viewComponent}");
+			}
+
+			var paramCandidates = invokeMethod
+				.GetParameters()
+				.Where(p => !paramObject.ContainsKey(p.Name!)); // filter out any already supplied parameters
```
adelikat, line 81: Will this be case-insensitive when matching parameter names?  We need it to be, can't count on users to match variable camelCasing

TASVideos.Test/ViewComponents/ParamHelperTests.cs
nattthebear: Yes, but the callee gets passed an `IList<int>`, so I have to put things in _some_ order.  If we want these all to be order-indifferent at the API level we'll need to switch to `ISet<int>`.  (Or maybe, some are `ISet<int>` and some are `IList<int>` and we have two different parameter binders and two different sets of tests.)

If we really do want the callee to get an `IList<int>`, then I think `SequenceEqual` is appropriate and correct due to principle of least surprise:  If the incoming string is `1,3`, as a callee I'd expect `l[0] == 1, l[1] == 3`.

TASVideos/TagHelpers/WikiMarkupTagHelper.cs
nattthebear: `paramObject` was initialized as `new Dictionary<string, object?>(StringComparer.OrdinalIgnoreCase)`, so this line is fine.

However, there's some question on the other side:  Will ASP.NET core's binder use the string equality I gave it?  I do not know, so leaving this open until I can check.

nattthebear: I changed the frames viewcomponent's parameter list to `(WikiPage pagedata` instead of `(WikiPage pageData` and it still got the pagedata, and was able to figure out the canonical NES FPS on a NES game page.  So I think that's all we needed there, right?

nattthebear: Oh, as far as the user supplied parameters go that all works too; `[module:frames|aMOUnt=400|    fPs=61.7]` is fully processed.

TASVideos.Test/ViewComponents/ParamHelperTests.cs
adelikat: Fair enough

adelikat (approved):

PR #313, opened 2021-07-12
PlayerPointsTable module

noxxaa:
PlayerPointsTable module used in /HomePages/Mothrayas/Rankings/AllTime

### DIFF
--- a/TASVideos/ViewComponents/PlayerPointsTable.cs
+++ b/TASVideos/ViewComponents/PlayerPointsTable.cs
@@ -1,0 +1,53 @@
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+using TASVideos.Core.Services;
+using TASVideos.Data;
+using TASVideos.Data.Entity;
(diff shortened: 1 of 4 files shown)


TASVideos/ViewComponents/PlayerPointsTable.cs
```diff
@@ -0,0 +1,53 @@
...
+
+		public PlayerPointsTable(ApplicationDbContext db, IPointsService pointsService)
+		{
+			_db = db;
+			_pointsService = pointsService;
+		}
+
+		public async Task<IViewComponentResult> InvokeAsync(int? count)
+		{
+			var showCount = count ?? 50;
```
adelikat, line 26: this might be avoidable by making the parameter of the invoke:  int? count = 50
Needs to be tested though, I'm not 100% sure it won't break the wiki module "model binder"

noxxaa: I did try that before, but it just passes null if the parameter isn't passed through the module. That's why I did it like this.

adelikat: Yeah, that makes sense, was just an idea

PR #314, opened 2021-07-13
Responsive Design Updates

TiKevin83:
Part 1: Tables
Starting out with a hotfix for viewing tables on mobile with a current example on the subforum pages. To make table elements mobile friendly you can wrap them in a div with overflow-x "auto" - however this is still not truly responsive. To be fully responsive I would wonder if the other devs are open to switching to flexbox instead of pure html table elements.

Part 2: CSS Compilation
Whatever SCSS/SASS compiler we use should generate files with extension .css to avoid the possibility of having to do extra server configuration for mime types. I added a commit to fix this, but it has other outstanding issues. Our CSS and JS compilers/optimizers should be capable of outputting sourcemaps, and I would even suggest having them on in production of the site is going to be open source, but ligersharks's weboptimizer.Sass has no sourcemap support at all. Are the other devs open to switching to something with sourcemap support if/when that exists?

### DIFF
--- a/TASVideos/Pages/Forum/Subforum/Index.cshtml
+++ b/TASVideos/Pages/Forum/Subforum/Index.cshtml
@@ -16,34 +16,34 @@
 </div>
 <hr />
 <partial name="_Pager" model="Model.Forum.Topics" />
-
-<table class="table table-striped table-bordered">
-	<tr>
-		<th></th>
-		<th>Topics</th>
-		<th>Replies</th>
-		<th>Author</th>
-		<td>Views</td>
-		<td>Last Post</td>
-	</tr>
-	@foreach (var topic in Model.Forum.Topics)
-	{
-		<tr>
-			<td>
-				<span condition="@(topic.Type == ForumTopicType.Sticky)" class="fa fa-sticky-note"></span>
-				<span condition="@(topic.Type == ForumTopicType.Announcement)" class="fa fa-bullhorn"></span>
-			</td>
-			<td>
-				<a asp-page="/Forum/Topics/Index" asp-route-id="@topic.Id">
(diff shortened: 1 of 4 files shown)


TASVideos/Pages/Forum/Subforum/Index.cshtml
```diff
@@ -16,34 +16,34 @@
 </div>
 <hr />
 <partial name="_Pager" model="Model.Forum.Topics" />
-
-<table class="table table-striped table-bordered">
```
adelikat, line 20: did tabs change to spaces here? if so, we want tabs, hard to tell with the indenting changing

TASVideos/Pages/Forum/Subforum/Index.cshtml
```diff
@@ -16,34 +16,34 @@
...
+    <table class="table table-striped table-bordered">
+        <tr>
+            <th></th>
+            <th>Topics</th>
+            <th>Replies</th>
+            <th>Author</th>
+            <td>Views</td>
+            <td>Last Post</td>
+        </tr>
+        @foreach (var topic in Model.Forum.Topics) {
```
adelikat, line 29: In our coding standards, curlies go on the next line for C#/razor

adelikat (changes requested):
I requested changes because of the curly positioning.  Sorry to be so nitpicky :)

TASVideos/Pages/Forum/Subforum/Index.cshtml
adelikat: Flex box instead of tables is fine with me, I use tables because it's quick and easy and I know how to do it :)  Flex box is probably a better design.  I apologize in advance, that there are a lot of tables in this code.  I would focus on any non-logged in user facing tables

PR #375, opened 2021-09-02
Change display of movie times to compacted version

Masterjun3:
TimeSpan has no way of easily create strings with optional hours or even days. So just make one.
This recreates the old display system where hours are only shown if they aren't 0. Same with days. Only minutes, seconds, and centiseconds always appear and are always two digits.
See the adjusted tests for examples.

### DIFF
--- a/TASVideos.Common/ITimeable.cs
+++ b/TASVideos.Common/ITimeable.cs
@@ -29,7 +29,22 @@ namespace TASVideos.Common
 				return TimeSpan.MaxValue;
 			}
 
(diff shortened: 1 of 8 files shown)


Masterjun3:
Oh hold on I forgot some occurrences because there are no submissions in the database.

adelikat (approved):

adelikat:
Glad we have all those unit tests and they were able to be fixed as part of this work.  Makes for an easy approval

Issue #371, opened 2021-09-02
Implement moviechangelog wiki module


EZGames69:
![image](https://user-images.githubusercontent.com/44628443/131784473-ffc7332b-b4bf-4d26-ab95-ecab2347e0b3.png)

adelikat:
module implemented

PR #392, opened 2021-09-05
MovieChangeLog module

Noxxaa:
MovieChangeLog module, as used in /NewMovies (linked from the front page)

Includes indicators for each new movie whether it is a new game, new branch, or improvement (replacing the old site's system where it shows obsoleted movies if any, which results in a more cluttered-looking list).

image


Noxxaa:
#371

PR #400, opened 2021-09-08
Fix Width of Elements on Mobile

TiKevin83:
Fixes all remaining cases of element widths being wider than the screen on mobile


adelikat (approved):
Approved, but maybe we should consider a table tag helper that generates the parent div and the table tag

PR #401, opened 2021-09-08
Add responsive pre-scales of award art

TiKevin83:
better compromise between bilinear and pixelated scaling


TASVideos/Pages/Shared/_MovieModule.cshtml
```diff
@@ -38,149 +47,152 @@
...
+				@foreach (var flag in Model.Flags.Where(f => !string.IsNullOrWhiteSpace(f.IconPath)))
+				{
+					<a href="@flag.LinkPath">
+						<img class="pixelart-image" title="@flag.Name" alt="[@flag.Name]" src="/@flag.IconPath" />
+					</a>
+				}
+			</div>
+		</row>
 	</cardheader>
-    <cardbody class="px-2 py-0">
```
adelikat, line 56: What happened here? whitespace?

TiKevin83: yeah you want everything standardized on tab delimited right?

TiKevin83:
I need to create a module/taghelper for the TierIcon side before merging this, it duplicated a lot of code.

TASVideos/Pages/Shared/_MovieModule.cshtml
adelikat: Yes please.  If there is a way to specify tabs in .cshtml files in .editorconfig, that would be ideal, so there is no need to nag

adelikat (approved):

PR #403, opened 2021-09-09
Port Bootstrap 4 spacing to bootstrap 5

TiKevin83:


Masterjun3:
Why are things like `mr-1` properly changed to `ms-1`, but then some instances remain wrong like `pr-0` not being changed to `pe-0`?
Intentional or accidental?

TiKevin83:
Oh I just forgot about 0, will add that

PR #406, opened 2021-09-10
Implement Dark Mode

### LINKED ISSUE #397: Implement Dark Mode
now that we've migrated to bootstrap 5 we can do the following:
use `media prefers-color-scheme: dark` to theme the site automatically based on the presence of dark mode
set different color variables on `:root {}` for the dark theme
add dark icon and change out light/dark icon visibility in the dark theme
add "dark mode" button in the header or footer to force toggle light/dark themes

TiKevin83:
Fixes #397 

Detects if a user's browser supports automatic dark themes and pulls down a dark mode stylesheet if the user is in dark mode.
Adds a button that will toggle the dark theme on/off to the footer to support operating systems/browsers that don't have system dark themes.
The dark mode stylesheet is intended to have few direct style rules and mainly swap out the value of CSS content properties that bootstrap uses to set colors and background colors.
Also added a folder for partials and restored the use of color variables to combine SASS with the content properties


adelikat:
For this to be ready, we need a way to persist the setting

Issue #407, opened 2021-09-11
Forum: Last Post field only includes date, and not author/post ID link


Noxxaa:
Main forum page: http://tasvideos.org/forum/ vs https://tasvideos.mistflux.net/Forum - last post field is missing entirely

![image](https://user-images.githubusercontent.com/5488449/132955300-5ef940bf-e8fb-47d3-b5e0-5695223ae458.png)

![image](https://user-images.githubusercontent.com/5488449/132955281-dc409899-863a-4bc8-8834-7ac6702f29b3.png)

Subforum page: http://tasvideos.org/forum/viewforum.php?f=2 vs https://tasvideos.mistflux.net/Forum/Subforum/2 - last post field only includes date, and does not include last poster nor a link to the last post (both highly useful for browsing the forum)

![image](https://user-images.githubusercontent.com/5488449/132955326-1acfec76-920e-4e61-a6c9-f4e5bab799f3.png)

![image](https://user-images.githubusercontent.com/5488449/132955342-34684576-a402-4266-a8f8-9c6d8592fbbd.png)

PR #414, opened 2021-09-12
Add last post link and profile links to subforum topic entries

Noxxaa:
Fixes regressions compared to the original forum. Profile links help finding user information, and latest post link icon helps with forum browsing in general.

Old:

image

New:

image

Old forum for comparison:

image

### DIFF
--- a/TASVideos/Pages/Forum/Subforum/Index.cshtml
+++ b/TASVideos/Pages/Forum/Subforum/Index.cshtml
@@ -38,12 +38,12 @@
 					</a>
 				</td>
 				<td>@Math.Max(0, topic.PostCount - 1)</td> @*Imported Submission Topics will have no replies and no original post*@
-				<td> <profile-link username="@topic.CreateUserName">@topic.CreateUserName</profile-link></td>
+				<td> <profile-link username="@topic.CreateUserName"></profile-link></td>
                 <td>
                     @if (topic.LastPost != null)
                     {
                         <timezone-convert asp-for="@topic.LastPostDateTime" /> <br />
-                        <profile-link username="@topic.LastPostUserName">@topic.LastPostUserName</profile-link>
+						<profile-link username="@topic.LastPostUserName"></profile-link>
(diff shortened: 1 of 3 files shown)


Noxxaa:
Addresses issue #407

TASVideos/Pages/Forum/Subforum/Index.cshtml
```diff
@@ -38,8 +38,15 @@
...
 				</td>
 				<td>@Math.Max(0, topic.PostCount - 1)</td> @*Imported Submission Topics will have no replies and no original post*@
-				<td>@topic.CreateUserName</td>
-				<td><timezone-convert asp-for="@topic.LastPostDateTime" /></td>
+				<td> <profile-link username="@topic.CreateUserName">@topic.CreateUserName</profile-link></td>
+                <td>
+                    @if (topic.LastPost != null)
+                    {
+                        <timezone-convert asp-for="@topic.LastPostDateTime" /> <br />
+                        <profile-link username="@topic.LastPostUserName">@topic.LastPostUserName</profile-link>
```
adelikat, line 46: This taghelper allows you to omit the value between the tags.  so you can simply do:
<profile-link username="@topic.LastPostUserName"></profile-link>

TASVideos/Pages/Forum/Subforum/Index.cshtml.cs
```diff
@@ -62,7 +62,7 @@ public async Task<IActionResult> OnGet()
 					CreateTimestamp = ft.CreateTimestamp,
 					Type = ft.Type,
 					PostCount = ft.ForumPosts.Count,
-					LastPost = ft.ForumPosts.Max(fp => (DateTime?)fp.CreateTimestamp)
+					LastPost = ft.ForumPosts.OrderBy(fp => fp.CreateTimestamp).Last()
```
adelikat, line 65: I wonder if .OrderByDescending().First() is better.  EF might be smart enough to do the same query either way, but it is worth checking

Noxxaa: Based on my understanding it is guaranteed at least not to be worse, so I've made the change in the next commit

PR #427, opened 2021-09-14
Use WebOptimizer's TagHelpers and Fix Cache

### LINKED ISSUE #424: Frontend Asset Cache Not Clearing
Local and dev site are caching frontend assets, we should set `max-age=0` and `cache-control=no-store` in HTTP headers for CSS and JS, at least in local and dev servers. This should help avoid repeated issues where CSS and JS updates appear broken until creating a new incognito window.

TiKevin83:
Fixes #424 

Allows WebOptimizer to control cache headers for its bundles and adjust cache behavior between dev and prod. Prod assets will now have query strings which will be updated on any changes to live assets to force the new assets to be used. Local and demo site CSS/JS bundles will have a no-cache header so they are never cached by the end user.

### DIFF
--- a/TASVideos/Pages/Shared/_Layout.cshtml
+++ b/TASVideos/Pages/Shared/_Layout.cshtml
@@ -6,7 +6,7 @@
 	<title>@(ViewData["Title"] != null ? ViewData["Title"] + " - " : "")TASVideos</title>
 
 	<link rel="shortcut icon" href="/favicon.ico" type="image/x-icon">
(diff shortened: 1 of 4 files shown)


adelikat:
How does this code account for production behavior?

TiKevin83:
> How does this code account for production behavior?

EnableCaching defaults to true for stylesheets loaded through the weboptimizer taghelpers, so adding the definition to prod appsettings would be unnecessary

adelikat:
Nice!

PR #437, opened 2021-09-16
Showing branches in tabularmovielist

### LINKED ISSUE #383: Support for showing Branch Labels in module:tabularmovielist
On the old site the tabularmovielist module shows branch labels alongside the game name. It would be preferred if the new site had a way to show the branch labels as well with that module as it is used in some key places (ie front page).

Masterjun3:
Resolves #383

### DIFF
--- a/TASVideos/Pages/Shared/Components/TabularMovieList/Default.cshtml
+++ b/TASVideos/Pages/Shared/Components/TabularMovieList/Default.cshtml
@@ -24,7 +24,7 @@
 					{
 						<tr>
 							<td><timezone-convert asp-for="@pub.CreateTimestamp" date-only="true" /></td>
-							<td>@pub.Game<span condition="pub.Branch is not null"> "@(pub.Branch)"</span></td>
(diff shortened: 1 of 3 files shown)


TASVideos/Pages/Shared/Components/TabularMovieList/Default.cshtml
```diff
@@ -24,7 +24,7 @@
 					{
 						<tr>
 							<td><timezone-convert asp-for="@pub.CreateTimestamp" date-only="true" /></td>
-							<td>@pub.Game</td>
+							<td>@pub.Game<span condition="pub.Branch is not null"> "@(pub.Branch)"</span></td>
```
adelikat, line 27: probably safer (more paranoid) to use !string.IsNullOrWhiteSpace(pub.Branch)

Masterjun3: *Ah*, yeah I saw it being used every now and then and never read what it actually means. Will change.

PR #442, opened 2021-09-16
Allow image resizing

Randomno:
No idea if this works to be honest

Might be preferable to add a single "size" attribute


nattthebear:
The code is fine and works, but this feature does not exist on the live site.  @adelikat are you OK with adding this as a new feature commit?  I don't have a good sense of how useful it is for editors.  Given that we've survived this long without it, it might not be important.

adelikat:
I really thought we already had that, maybe I misremembered.  Either way, it seems like a nice addition, that does not add any much risk

Randomno:
It was brought up in Discord yesterday which I why I added it.

As another note might be better to use w and h to be consistent with the YouTube module.

PR #459, opened 2021-09-22
Add forum table styling

### LINKED ISSUE #456: Feature request: borders in forum tables
Very few people know this feature exists, but when they use it, it becomes incredibly hard to figure out the columns. It'd be very helpful if there were borders around cells.

https://tasvideos.mistflux.net/Forum/Topics/21916?Highlight=500037&CurrentPage=1#500037

Old site doesn't have them either

http://tasvideos.org/forum/viewtopic.php?p=500037#500037

Randomno:
Fixes #456 

Just copied the wiki table styling, could be changed.

### DIFF
--- a/TASVideos/wwwroot/css/partials/_customizations.scss
+++ b/TASVideos/wwwroot/css/partials/_customizations.scss
@@ -45,28 +45,29 @@ body {
 	overflow-x: auto;
 }
 
-/* Wiki Tables */
-article.wiki table {
-	width: 100%;
-	margin-bottom: 1rem;
-	color: var(--bs-dark);
-	border-collapse: collapse;
+/* Wiki and Forum Tables */
+article.wiki , .postbody {
+	table {
+		width: 100%;
+		margin-bottom: 1rem;
+		color: var(--bs-dark);
+		border-collapse: collapse;
 
-	th, td {
-		padding: 0.3rem;
-		vertical-align: top;
-		border: 1px solid var(--bs-gray-600);
-	}
+		th, td {
+			padding: 0.3rem;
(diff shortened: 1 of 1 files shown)


TASVideos/wwwroot/css/partials/_customizations.scss
```diff
@@ -83,6 +83,29 @@ article.wiki table {
...
+		vertical-align: top;
+		border: 1px solid var(--bs-gray-600);
+}
+
+.postbody tbody th {
+		background-color: rgba(var(--bs-black-rgb), 0.05);
+}
+
+.postbody tbody + tbody {
+		border-top: 2px solid #dee2e6;
```
adelikat, line 105: I object to the use of a random inline style color.  Is there not a color that is "close enough" that is already defined? and if not, we need to put this into the color variables.

We want to consider color-blindness and other accessibility concerns.  Step 1 is to minimize the colors used, and Step 2 is to have them all defined in one place, so they can easily be tweaked as needed

TiKevin83, line 105: `#dee2e6` should be either a SASS variable like `$dropdown-text-dark` if not intended to change in dark mode, or a native css var if it needs to change in dark mode (good use of those otherwise though)

Randomno:
That color is used in the existing wiki table styling, which is what this copies.

If the forum tables should have the same styling I believe this could be implemented by just changing line 49 to

`article.wiki table, .postbody table {`

TiKevin83:
I'd be fine with using the new table style in the forum, even better would be
```
article.wiki, .postbody {
  table {
  }
}
```

Issue #393, opened 2021-09-05
Implement relative time display for posts (and other things)


Masterjun3:
Practically all of the modern Internet shows `1 hour ago` instead of `2021-09-06 22:33:44` for time displays. It usually shows the precise timestamp when hovering or clicking on the relative time.

Of course, the trivial solution of dynamically changing the webpage would run into problems e.g. with caching.
So I assume the solution here would be a javascript one. Just serving the date in ISO 8601 format and then running the javascript over the tags to convert it, I think.

nattthebear:
> Of course, the trivial solution of dynamically changing the webpage would run into problems e.g. with caching.

The site has a lot of dynamic content on it and there's no setup in place for making any of that caching friendly, so we should be able to just do this.

Masterjun3:
Implemented in bd46d86a8ad45db7509cdf6ff66e6a837a24f003.

PR #464, opened 2021-09-23
Displaying recent timestamps in relative time form

Masterjun3:
This is aiming to resolve my #393.

It replaces the absolute timestamp with a relative form if the timestamp is more recent than 30 days ago.
image

It also adds a `title` html attribute showing the full date to all date displays (not only relative ones).
image

### DIFF
--- a/TASVideos/Pages/Shared/Components/UserMovies/Default.cshtml
+++ b/TASVideos/Pages/Shared/Components/UserMovies/Default.cshtml
@@ -11,7 +11,7 @@
 		@foreach (var item in Model)
 		{
 			<tr>
-				<td><timezone-convert asp-for="@item.UploadTimestamp" date-only="true" relative-time="true" /></td>
+				<td><timezone-convert asp-for="@item.UploadTimestamp" date-only="true" /></td>
 				<td>@item.Author</td>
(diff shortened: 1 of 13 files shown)


Masterjun3:
Even though I suggested doing it in JavaScript to allow for better caching, this is now all being done on the server side of things.

Masterjun3:
Problem: `at 1 minute ago`
image
Maybe it would make sense to add another boolean "inline" to generate strings fitting into a sentence.

TASVideos/Pages/Shared/Components/UserMovies/Default.cshtml
```diff
@@ -11,7 +11,7 @@
 		@foreach (var item in Model)
 		{
 			<tr>
-				<td><timezone-convert asp-for="@item.UploadTimestamp" date-only="true" /></td>
+				<td><timezone-convert asp-for="@item.UploadTimestamp" date-only="true" relative-time="true" /></td>
```
adelikat, line 14: It seems you want to use relative time pretty often, maybe true should be the default and optional?

Masterjun3: Yeah, I didn't realize there were so many separate places where it's useful to have.

PR #478, opened 2021-09-24
Various small redesigns

Masterjun3:
- changes links from
`Forum/Topics/22418?Highlight=502940&CurrentPage=1#502940` to
`Forum/Topics/22418?CurrentPage=1&Highlight=502940#502940`
- changes PM icon to envelope
- makes navbar wrap less and makes some text bold
- adjust FrontpageSubmissionList
- small design fixes in the profile
- overflow scroll only on desktop


TASVideos/Pages/Shared/_NavBarPartial.cshtml
```diff
@@ -1,31 +1,31 @@
-<navbar>
-	<nav-dropdown activate="Home">
```
TiKevin83, line 2: bootstrap font-weight-bold class should be used in place of the b element

Masterjun3: Fixed with `fw-bold`.

Issue #342, opened 2021-08-30
Font of forum location is too tiny


vadosnaprimer:
https://i.imgur.com/tv9wXGT.png

Also it'd be better to have a button sending to submission page rather than a tiny text in parentheses.

adelikat:
The first one is dubious to request a change too, you have the same text in big letters right above it, it is redundant at best.  Also it is exactly how the old phpbb2 forum behaves.

For the 2nd piece, A button, I'm not sure about.  What do other people think?

vadosnaprimer:
The old site has smaller font in general. In this pace it looks tiny for no reason and kinda out of place.

The text above is the topic name but not the forum location. Though the old site doesn't have the same text in 2 places close to each other. It only shows thread name in one place.
https://i.imgur.com/0fhTWOM.png

Randomno:
https://i.imgur.com/PnERA0p.png

Thoughts?

- Navigation from <small> to <span>
- -> to → (don't think this symbol will cause issues, and -> looks particularly bad on subforums like [here](https://tasvideos.mistflux.net/Forum/Subforum/7))
- Submission header to 18px
- link text to lowercase (insignificant but thought it looked better)

Masterjun3:
Lowercase looks bad here. Inconsistent with the title case of all the other titles on the site.

Randomno:
Not sure what qualifies as a title but there seems to be several inconsistencies between title and sentence case.

Also using h5 instead of 18px would be consistent with the header on the submission page. It's a little bigger.

vadosnaprimer:
I don't really think the same exact text should be present in 2 places close to each other. Since the old forum only shows which subforum we're in, and then displays the topic title separately. So I'd just do it like this (font style is already there in subforum view)

![изображение](https://user-images.githubusercontent.com/7092625/134528228-a36e5bd5-325a-45bd-a00a-8955d99cfcdb.png)

Randomno:
Been looking into implementing this. I agree with feos's navigation change, since having the thread title in 3 different places near each other is unnecessary, but I'm not sure how best to deal with locked threads.

https://tasvideos.mistflux.net/Forum/Topics/21346

If the current thread is removed from the navigation the red border won't make sense anymore.

vadosnaprimer:
On the old forum, "Locked" replaces the "Post" button, so the same can be done here (with red BG).

Masterjun3:
I have it like this (before | after):
[...]

PR #494, opened 2021-09-26
Header and title adjustments

Masterjun3:
- Normalize document titles to
Content title - Location - TASVideos
- Center headers inside card-headers by removing margin (how bootstrap intended)
- Change -> to →
(is non-ascii a problem? a bootstrap arrow icons would work, but make copying bad)
- Hide signatures in mobile view
- Fix maximum width h1 titles by putting them into a container (how the default layout does it)
- Swaps order of Announcement and Sticky topics to fit with old site

Fixes one problem mentioned in #342.

### DIFF
--- a/TASVideos/wwwroot/css/partials/_customizations.scss
+++ b/TASVideos/wwwroot/css/partials/_customizations.scss
@@ -282,3 +282,9 @@ article.wiki, .postbody {
 .fa {
 	transform: translate(0);
 }
+
+.card-header {
+	h1, h2, h3, h4, h5, h6 {
+		margin-bottom: 0;
+	}
+}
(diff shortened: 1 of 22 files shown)


TASVideos/wwwroot/css/partials/_customizations.scss
```diff
@@ -282,3 +282,9 @@ article.wiki, .postbody {
 .fa {
 	transform: translate(0);
 }
+
+.card-header {
```
adelikat, line 286: this affects every card on the site, are you sure this makes sense?

Masterjun3: Yes!
The card-header tag itself already has `margin-bottom: 0;` and it was meant to be used as a tag for h1-h6, not as a class for a div with h1-h6 inside. They overwrite the `margin-bottom: 0;` and are not centered anymore. But since using it in a div makes stuff like buttons and badges possible, this is the workaround.

PR #514, opened 2021-09-29
fixes #510 - adds visited link style

### LINKED ISSUE #510: Visited forum links stays blue instead of purple
Repro:
1. https://tasvideos.mistflux.net/Forum/Subforum/7
2. Click on a topic
3. Go back to previous page

The link should turn purple (visited) to help users maintain their progress of visiting various topics.

TiKevin83:
fixes #510 

requesting lots of feedback before we merge this, it's a drastic change. There are potentially a lot of links that may need to change to being styled as buttons for this change to make consistent visual sense, for example the link back to the main "Forum" page on a subforum does not really make sense to be given a visited style, it doesn't help anything to the user to know whether they've already visited the main forum page

### DIFF
--- a/TASVideos/wwwroot/css/partials/_customizations.scss
+++ b/TASVideos/wwwroot/css/partials/_customizations.scss
@@ -15,6 +15,10 @@ body {
 	background: var(--tasvideos-bg) linear-gradient(var(--tasvideos-bg-gradient) 0, var(--tasvideos-bg) 250px) repeat-x;
 }
 
+a:visited {
+	color: purple;
+}
+
 .btn-separator::after {
 	content: ' ';
 	display: block;


Masterjun3:
If a visited style is ever implemented, just adding a "purple" to the visited links like in this pull request shouldn't be how it's done.
- We don't want every single link to be recolored.
For example in the forum view it makes no sense to recolor the subforums. Thus we need a system to easily stop certain links to be recolored.
- The recoloring doesn't always need to be purple.
This makes the whole idea of our primary and secondary theme colors obsolete. The forum on the old site used a lighter version of blue as a visited style for the blue forum links.
- Add a visited style to buttons.
This can be done by making their color lighter as well, but this one should be possible to be enabled manually, as most buttons don't need it.

TiKevin83:
having to create new PR, unable to push to the branch on upstream

PR #524, opened 2021-10-02
Wiki Text Renderer System with Frames module implemented

adelikat:
This PR wires up a text version of wiki modules, to be implemented as needed.  Youtube is the current integration that needs a text only version.  This implements the frames module since publication descriptions are the target use case and the frames module is frequently used here.  Most other modules would be inappropriate in this situation and are hard to find examples in real descriptions.

The goal here is to set up a system where a text version of a module can easily be implemented as needed.

### DIFF
--- a/TASVideos/Services/ModuleParamHelpers.cs
+++ b/TASVideos/Services/ModuleParamHelpers.cs
@@ -81,6 +81,7 @@ namespace TASVideos.Services
 					if (needsNonNull)
 					{
 						// TODO: Better styling, or something
+						// TODO: do not pass in TextWriter and this isn't how it should report that failure. We should return some failure state to the caller which can then handle output
 						w.Write($"MODULE ERROR for `{name}`: Missing parameter value for {paramCandidate.Name}");
 						return new Dictionary<string, object?>();
 					}
--- a/TASVideos/Services/ModuleParamHelpers.cs
+++ b/TASVideos/Services/ModuleParamHelpers.cs
@@ -1,0 +1,95 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using Namotion.Reflection;
+using TASVideos.Data.Entity;
+using TASVideos.TagHelpers;
+using TASVideos.ViewComponents;
+
+namespace TASVideos.Services
+{
+	public static class ModuleParamHelpers
+	{
+		public static readonly Dictionary<Type, IModuleParameterTypeAdapter> ParamTypeAdapters = typeof(WikiMarkup)
+			.Assembly
+			.GetTypes()
+			.Where(t => t.BaseType != null
+				&& t.BaseType.IsGenericType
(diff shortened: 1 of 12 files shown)


TASVideos/Services/ModuleParamHelpers.cs
```diff
@@ -0,0 +1,95 @@
...
+
+				if (result == null)
+				{
+					// Conversion failed.  See if the parameter type is a failable type.
+					var needsNonNull = paramType.IsValueType && doNullableWrap
+						|| !paramType.IsValueType && paramType.ToContextualType().Nullability == Nullability.NotNullable;
+					if (needsNonNull)
+					{
+						// TODO: Better styling, or something
+						w.Write($"MODULE ERROR for `{name}`: Missing parameter value for {paramCandidate.Name}");
```
nattthebear, line 84: With the way this has been refactored, this method shouldn't receive the `TextWriter` and this isn't how it should report that failure.  We should return some failure state to the caller which can then handle output.

TASVideos/Services/WikiToTextRenderer.cs
```diff
@@ -53,10 +66,34 @@ public bool CheckCondition(string condition)
...
+				if (!componentExists)
+				{
+					return;
+				}
+
+				var invokeMethod = textComponent!.GetMethod("RenderTextAsync");
+
+				if (invokeMethod == null && textComponent.GetMethod("RenderText") != null)
+				{
+					throw new NotImplementedException("Sync method not supported yet");
```
nattthebear, line 81: How strongly do we feel about also supporting a `RenderText`(Sync)?  I don't see any problem with doing just async and cutting this entirely.

TASVideos/Services/WikiToTextRenderer.cs
```diff
@@ -53,10 +66,34 @@ public bool CheckCondition(string condition)
...
+				}
+
+				if (invokeMethod == null)
+				{
+					throw new InvalidOperationException($"Could not find an RenderText method on ViewComponent {textComponent}");
```
nattthebear, line 86: I think this error is supposed to say RenderTextAsync, not RenderText.

TASVideos/Services/WikiToTextRenderer.cs
```diff
@@ -53,10 +66,34 @@ public bool CheckCondition(string condition)
...
+				}
+
+				var paramObject = ModuleParamHelpers
+					.GetParameterData(w, name, invokeMethod, _wikiPage, pp);
+
+				var module = _serviceProvider.GetRequiredService(textComponent);
+				var task = (Task)invokeMethod.Invoke(module, paramObject.Values.ToArray())!;
```
nattthebear, line 93: You should be able to just cast to `Task<string>` here and leave out the reflection bit below.

nattthebear (commented):
I don't see any major issues here, other than it's ugly, but the whole thing is ugly.

PR #549, opened 2021-10-14
Ratings: add slider, add decimal parsing, and other rating things

Masterjun3:
* add number box and slider
* parse both dot and comma separated values
* redirect to rating index
* show yourself on the rating index page even if your ratings are set to hidden

### DIFF
--- a/TASVideos/Pages/Publications/Models/PublicationRateModel.cs
+++ b/TASVideos/Pages/Publications/Models/PublicationRateModel.cs
@@ -1,17 +1,62 @@
-﻿using System.ComponentModel.DataAnnotations;
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+using System.Globalization;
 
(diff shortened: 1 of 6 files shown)


TASVideos/Pages/Publications/Models/PublicationRateModel.cs
```diff
@@ -1,17 +1,62 @@
-﻿using System.ComponentModel.DataAnnotations;
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+using System.Globalization;
 
 namespace TASVideos.Pages.Publications.Models
 {
 	public class PublicationRateModel
 	{
+		public sealed class RatingString : ValidationAttribute
```
adelikat, line 9: I've never seen a validation attribute slammed into a page model before.  I guess if the intended scope is ONLY this page, that's fine.  But we also have a folder for these things

PR #568, opened 2021-10-19
Email Confirmation Overhaul, and allow resending of email confirmation

### LINKED ISSUE #450: New site needs a way to re-send activation e-mail.
The current site doesn't appear to have a way to resend the activation e-mail, so if one doesn't get sent for some reason, that attempted forum/site account will never exist without manual intervention.

adelikat:
Do not require email confirmation to sign in, but only assign permissions when email is confirmed.  Add a warning and a resend email confirmation link on profile.

Resolves #450

Note that in dev environments, more setup will be needed to register a new user and allow them to have permissions (by default it uses EmailLogger, so the link can be grabbed from logs)

### DIFF
--- a/TASVideos/Pages/Account/Register.cshtml.cs
+++ b/TASVideos/Pages/Account/Register.cshtml.cs
@@ -133,8 +133,8 @@ namespace TASVideos.Pages.Account
 				var result = await _userManager.CreateAsync(user, Password);
 				if (result.Succeeded)
 				{
-					var code = await _userManager.GenerateEmailConfirmationTokenAsync(user);
-					var callbackUrl = Url.EmailConfirmationLink(user.Id.ToString(), code, Request.Scheme);
+					var token = await _userManager.GenerateEmailConfirmationTokenAsync(user);
+					var callbackUrl = Url.EmailConfirmationLink(user.Id.ToString(), token, Request.Scheme);
 					await _emailService.EmailConfirmation(Email, callbackUrl);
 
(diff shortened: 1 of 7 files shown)


TASVideos/Pages/Account/Register.cshtml.cs
```diff
@@ -133,48 +133,28 @@ public async Task<IActionResult> OnPost()
 				var result = await _userManager.CreateAsync(user, Password);
 				if (result.Succeeded)
 				{
-					await AddStandardRoles(user.Id);
-					await _userManager.AddUserPermissionsToClaims(user);
+					var code = await _userManager.GenerateEmailConfirmationTokenAsync(user);
```
TiKevin83, line 136: can we call this variable "token" or "confirmationToken" to be more consistent with the method being called?

TASVideos/Pages/Account/SendConfirmationEmail.cshtml.cs
```diff
@@ -0,0 +1,32 @@
...
+		public SendConfirmationEmail(UserManager userManager, IEmailService emailService)
+		{
+			_userManager = userManager;
+			_emailService = emailService;
+		}
+
+		public async Task<IActionResult> OnPost()
+		{
+			var user = await _userManager.GetUserAsync(User);
+			var code = await _userManager.GenerateEmailConfirmationTokenAsync(user);
```
TiKevin83, line 25: same here

PR #598, opened 2021-10-31
Forum font size BBCode unit-less number fix

Masterjun3:
There are a lot of instances where `[size=9]text[/size]` was used.
The old site converted it to `font-size: 9px`.
The current demo site converts it to `font-size: 9` which isn't valid.

Just adding a px string would give wrong results, since the default font size was 12px but is now 14px, and might or might not change.
This PR instead converts `[size=9]` to `font-size: 0.75em` (9px/12px = 0.75) to give a proper representation regardless of default font size.

### DIFF
--- a/TASVideos.ForumEngine/Node.cs
+++ b/TASVideos.ForumEngine/Node.cs
@@ -368,7 +368,17 @@ namespace TASVideos.ForumEngine
 					w.Write("<span style=");
 
 					// TODO: More fully featured anti-style injection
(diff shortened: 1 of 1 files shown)


nattthebear:
> The old site converted it to `font-size: 9px`.

I'm not seeing this?

image

Masterjun3:
Browsers implicitly convert decimcals to pixels, but integers give an error.

nattthebear (approved):

PR #599, opened 2021-11-02
Refactor forum engine HTML output to be less error prone

nattthebear:
Eventually, the wiki engine will use this as well.


nattthebear:
Could use some feedback here, especially on how easy (or hard) this is to understand and use for coders who are not overly familiar with the forum engine.

adelikat:
Looks readable to me

PR #601, opened 2021-11-03
Visited Links on Forum Threads and Last Posts

### LINKED ISSUE #510: Visited forum links stays blue instead of purple

TiKevin83:
fixes #510 more precisely

I still disagree with the need to use a color other than purple, the purple visited style is pretty ubiquitous, willing to change it though if there's wider disagreement.

I updated the PR to only affect visited links on the forum thread links and last post links, it looks pretty useful to my eye.

### DIFF
--- a/TASVideos/Pages/Forum/Subforum/Index.cshtml
+++ b/TASVideos/Pages/Forum/Subforum/Index.cshtml
@@ -35,7 +35,7 @@
 				<td>
 					<span condition="@(topic.Type == ForumTopicType.Announcement)" class="fw-bold"><i class="fa fa-bullhorn"></i> Announcement: </span>
 					<span condition="@(topic.Type == ForumTopicType.Sticky)" class="fw-bold"><i class="fa fa-sticky-note"></i> Sticky: </span>
-					<a asp-page="/Forum/Topics/Index" asp-route-id="@topic.Id" class="fw-bold">@topic.Title</a>
+					<a asp-page="/Forum/Topics/Index" asp-route-id="@topic.Id" class="fw-bold thread-link">@topic.Title</a>
 					<div class="ms-2">
 						@{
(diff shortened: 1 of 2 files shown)


Masterjun3:
The problem with making post links colored is they don't get colored when you browse the forum by clicking on threads and advance pages to the latest post.
I dislike how this creates the illusion that purple = seen the post, and, blue = not seen the post, which isn't true in this design.

vadosnaprimer:
Here's the current color difference (top not visited, bottom visited)

изображение

Masterjun3:
> Is it about that while the thread you already visited becomes purple (edited) **even after** a new post has been made?

No.
It's about creating the illusion that purple means seen and blue means not seen. But this isn't true with this pull request.

TiKevin83:
to be fair masterjun I'm not sure if that's a solvable problem, I think we'd be stuck between no visited link styling at all and the visited style not covering the "last post" unless you read it through that exact link.

Masterjun3:
The visited link style works for threads, since you then know what thread you were in before.
But there is no need to make post links visisted. That's not how the site used to work, and it brings new issues.
A better way to separate new posts from old posts would be to use something similar to the "Posts since last visit" feature, not a browserside visited style.

vadosnaprimer:
> The visited link style works for threads, since you then know what thread you were in before. But there is no need to make post links visisted. That's not how the site used to work, and it brings new issues. A better way to separate new posts from old posts would be to use something similar to the "Posts since last visit" feature, not a browserside visited style.

I agree with this.

TiKevin83:
Updated PR to only affect threads and not last posts

PR #614, opened 2021-11-14
Remove overflow y scrolls

### LINKED ISSUE #399: Replace all overflow-y scrolls
This is a debatable design feature. Currently, there are several elements with a maximum height, which show content in a scrollable block if the height exceeds the maximum value.

I personally agree with the feedback on the forums that this should be replaced with a different design. The block-scrolling can interrupt page-scrolling, and this is especially noticable on mobile devices where actual page-scrolling is only possible on the side edges of the screen.

For small limited blocks such as the featured movie or the news on the front page it should be asked if it's even necessary to limit the height. 
For potentially larger blocks such as the submission text, a friendlier solution would be a label at the bottom saying "Read more", which when pressing displays more of the content.

TiKevin83:
fixes #399 
Might not be amazing for long submission texts

### DIFF
--- a/TASVideos/Pages/Forum/Topics/Split.cshtml
+++ b/TASVideos/Pages/Forum/Topics/Split.cshtml
@@ -24,7 +24,7 @@
 		<input type="text" asp-for="Topic.SplitTopicName" class="form-control" />
 		<span asp-validation-for="Topic.SplitTopicName" class="text-danger"></span>
 	</form-group>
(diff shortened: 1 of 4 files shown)


adelikat:
I think overflow-y is at least needed for submission forum topics.  Take a look at 6678S

TiKevin83:
updated PR to preserve overflow scroll for the forum submission notes

PR #626, opened 2021-11-28
Monospace element background colors

### LINKED ISSUE #329: Different BG color for monospace
It'd be awesome if a different BG color was used for monospace font in the wiki (things like starting with a space and `{{these}}`), like some darker grey or maybe white? We will need A LOT of small code embeds, including inline, on pages like https://tasvideos.mistflux.net/EmulatorResources/PCem/DOS and it's very hard to spot what is regular text and what is exact internal strings. When reading such stuff, it helps a lot if code is highlighted, because that serves as attention anchors of some kind.

Wiki `> quote` syntax has some nice darker grey that could be used here.

If the line starts with a space, the entire line's BG can be changed, and if it's `{{}}` then only of the text within the tag

TiKevin83:
fix #329 
Have the wiki render `{{}}` to `<code>` instead of `<tt>` so it isn't using deprecated elements and gets picked up by bootstrap's monospace font reboot
Also adds a background color that varies for dark mode.

### DIFF
--- a/TASVideos.WikiEngine/NewParser.cs
+++ b/TASVideos.WikiEngine/NewParser.cs
@@ -535,11 +535,11 @@ namespace TASVideos.WikiEngine
 			}
 			else if (Eat("{{"))
 			{
-				Push("tt");
+				Push("code");
 			}
 			else if (Eat("}}"))
 			{
-				if (!TryPop("tt"))
+				if (!TryPop("code"))
 				{
 					AddText("}}");
 				}
--- a/TASVideos/wwwroot/css/partials/_customizations.scss
+++ b/TASVideos/wwwroot/css/partials/_customizations.scss
@@ -352,6 +352,13 @@ article.wiki, .postbody {
 	font-size: 28px;
 }
 
+tt {
+	font-family: var(--bs-font-monospace);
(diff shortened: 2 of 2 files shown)


vadosnaprimer:
Thanks! Can you post pics please?

nattthebear (approved):

TiKevin83:
NewLight
CurrentDark
CurrentLight
NewDark

adelikat (approved):

vadosnaprimer:
Looks like using bold (maybe other style too?) on inline monospace sample resets the font color. Also font color is different between inline and whole-line (I still prefer default color, black/white). And what about the forum?

TiKevin83:
Personally I like how the pre and code tags have separate colors, the bold thing can be fixed in wiki markup but we could override bootstrap if we don't like what it's doing there. Checking on the forum `tt` tag, ideally I'd want to change that to `code` too if possible

vadosnaprimer:
I guess it's fine for now. Bold/italic probably doesn't make sense on inline code so yeah we can fix the wiki instead.

vadosnaprimer (approved):

PR #631, opened 2021-11-29
Static downtime page

### LINKED ISSUE #629: Site is down for Upgrade page
We need a static html document we can put up while the site is migrating over to the new code.  It will be up for some hours.  It needs to clearly indicate the situation and be styled in a TASVideos theme of some kind.

TiKevin83:
fixes #629


Masterjun3:
Probably best to just sort information by importance.
```
TASVideos.org is temporarily down while we are upgrading the site.
This is expected to take no longer than a few hours.

Compared to the old site, the new site is [open source] and will feature:
- HTTPS
- Mobile-friendly design
- Dark mode
- ...and more!

For more information you can reach us on our chat rooms on [Discord] or [IRC].
```

Masterjun3:
I was about to approve the changes, but quite frankly, the two bullet point list changes are just bad.

Changing to "Mobile First Responsive Design" just feels like Desktop design is abandoned or second priority. Users don't need to care that Bootstrap is designed in a way where we start at mobile design and then adjust to desktop, especially since most of the wiki pages are designed Desktop First anyway (because of the old site predating mobile phones).

Changing to "Autodetected Dark Mode" instead of just "Dark Mode" undermines the whole comparison with the old site. "Compared to the old site we now have Autodetected Dark Mode" just sounds like we already had dark mode in the old site and now it's even autodetected.

I know it's tempting to list all cool new stuff, but all we need to tell is "It's secure now, you can use your phone now, and it's easy on the eyes now". For all the technical stuff we already link the github page.

PR #641, opened 2021-12-06
fix mobile screen overflow of post creation button groups

### LINKED ISSUE #546: BBCode buttons clip through the sides when creating new posts or topics on mobile
Example page (view on mobile/small widths): https://demo.tasvideos.org/Forum/Posts/Create/22808

TiKevin83:
fix #546

### DIFF
--- a/TASVideos/Pages/Forum/_CreatePostHelper.cshtml
+++ b/TASVideos/Pages/Forum/_CreatePostHelper.cshtml
@@ -10,17 +10,17 @@
 	<button type="button" class="btn bg-info text-dark btn-sm" tabindex="-1" data-fmt="[sup],[/sup]" data-skey="+"><sup>sup</sup></button>
 </div>
 
-<div class="btn-group mt-1 mb-1">
-	<button type="button" class="btn bg-info text-dark btn-sm" tabindex="-1" data-fmt="[quote],[/quote]" data-akey=q>quote</button>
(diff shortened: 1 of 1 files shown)


Masterjun3:
This change currently doesn't really look good:
image
Maybe make it stay at a size fitting for the text and put outlines around buttons.

Masterjun3 (changes requested):
This change currently doesn't really look good:
image
Maybe make it stay at a size fitting for the text and put outlines around buttons.

TiKevin83:
I've made it so the buttons no longer grow to fit the full width when wrapping and I've given them a border so they should be more easily distinguishable

Masterjun3:
Nice, looks good. Any reason why the button groups aren't consistent? Some have the flex-wrap, others don't.

TiKevin83:
Those groups shouldn't be affected by this problem except in unusably tiny screen widths so I think it's better to leave them

Masterjun3:
Well, or if they have an increased font size, or if we decide to add more buttons. Can you explain why it's "better" to leave it?
Feels like a weird inconsistency.

TiKevin83:
I could go either way on it but to me there's a design difference where some of these button groups are more like a button toolbar and the order and how close they are to each other doesn't matter, and the ones with fewer buttons are more tightly semantically connected and you wouldn't want a line to wrap with only one of them at a time

Masterjun3 (approved):

PR #642, opened 2021-12-06
Standardize site on ES6

TiKevin83:
Standardizes inline javascript resources on ES6+ language features

Drops `toArray()` helper in favor of ES6 `Array.from()`
Drops polyfills of ES6 functions (all 95%+ browser support)
Replaces uses of var with let and const
Per above, drops IIFEs in favor of block scoping
Simplifies many anonymous functions with arrow functions


nattthebear (approved):
I especially like getting rid of the non-standard builtin object extensions

Issue #646, opened 2021-12-08
Publication view: replace torrent links with direct links


vadosnaprimer:
![изображение](https://user-images.githubusercontent.com/7092625/145250577-71853105-46f2-479d-add9-7d64efb0a116.png)

This thread reminds us why torrents as we use them right now are useless:
http://tasvideos.org/forum/viewtopic.php?t=22931

Direct links on the other hand are not useless and will be relevant for a good while. But the publication module only has torrent links and not direct download links.

ThunderAxe31:
Just to give a bit more context: from looking at the source code of `_MovieModule.cshtml` it seems there is already a logic for including the direct urls to the available mirrors, but for some reason it's not working:
```
<a condition="@(!string.IsNullOrWhiteSpace(Model.MirrorSiteUrl))" href="@Model.MirrorSiteUrl" class="ms-2">A/V file via Mirror<br /></a>
```
Instead, the logic for showing the Torrent downloads is working, even though I think it should be removed (topic that feos linked) due to their uselessness:
```
<span condition="Model.TorrentLinks.Any()">
	@foreach (var torrent in Model.TorrentLinks)
	{
		<a href="~/torrent/@torrent.Path" class="ms-2">A/V file via BitTorrent @TorrentRemark(torrent.Path, torrent.Id)</a>
		<br />
	}
</span>
```

Masterjun3:
I'm not even sure what the purpose of a MirrorSiteUrl even is (It's not multiple, only a single string). Submissions also have a MirrorSiteUrl field and it gets transferred to the Publication on publishing. It's next to the submission Online-watching URL, so I suppose it's for secondary streaming encodes, but the publication OnlineWatchingUrls exists for that purpose.

The actual direct download links are in a different database table called PublicationUrls.

PR #647, opened 2021-12-08
Prototype improvement for watching options

ThunderAxe31:
Hopefully I didn't submit a completely nonsensical code... here goes nothing. Note that the aesthetic and the logic will need to be properly adjusted.

As promised, I've made a very rough draft of my [idea](http://tasvideos.org/forum/viewtopic.php?p=509570#509570) of grouping all the available watching options under the movie screenshot and making the wording more intuitive. More in detail:
- Putting a link to the relative emulator page by the side of the movie url (logic not coded yet)
- Putting the javascript button for enabling the YouTube embed (logic not coded yet) before the external link
- Putting the javascript button for enabling the Archive.org embed (logic not coded yet) before the external link (currently [broken](https://github.com/TASVideos/tasvideos/issues/646))

I've also moved the link to Submission, Discussion, Ratings, to the upper right. This is just temporary, I just needed to move away those buttons from the watching options; feel free to adjust these to your liking.
Oh, and I also moved the three wiki edit-related buttons under the movie description, instead than at the bottom left. (The "Catalog" button didn't appear because I don't have the privilege)

Have here a comparison through screenshots. The first image is a screen of the current page at https://staging.tasvideos.org/4271M while the second image is a rough mockup I made in Paint:

(Please ignore the fact that the screenshot is distorted in the second screenshot, I have no idea how but something happened in Opera while I was playing around with the debugger.)

### DIFF
--- a/TASVideos/Pages/Shared/_MovieModule.cshtml
+++ b/TASVideos/Pages/Shared/_MovieModule.cshtml
@@ -85,10 +85,10 @@
 				<div>
 					@foreach (var url in Model.OnlineWatchingUrls)
 					{
-						YouTube: <a>Embed here</a><a href="@url" class="btn btn-primary btn-sm mt-1" target="_blank"><i class="fa fa-external-link"></i> New tab</a>
+						YouTube: <a>Watch here</a><a href="@url" class="btn btn-primary btn-sm mt-1" target="_blank"><i class="fa fa-external-link"></i> New tab</a>
 					}
 				</div>
-				<div>Video .mp4: <a>Embed here</a><a condition="@(!string.IsNullOrWhiteSpace(Model.MirrorSiteUrl))" href="@Model.MirrorSiteUrl" class="ms-2"> Video file via Mirror<br /></a>
+				<div>Video .mp4: <a>Watch here</a><a condition="@(!string.IsNullOrWhiteSpace(Model.MirrorSiteUrl))" href="@Model.MirrorSiteUrl" class="ms-2"> Video file via Mirror<br /></a>
 				</div>
 				@foreach (var award in publicationAwards.OrderByDescending(a => a.Year))
 				{
--- a/TASVideos/Pages/Shared/_MovieModule.cshtml
+++ b/TASVideos/Pages/Shared/_MovieModule.cshtml
@@ -1,4 +1,4 @@
-﻿@model TASVideos.Pages.Publications.Models.PublicationDisplayModel
+@model TASVideos.Pages.Publications.Models.PublicationDisplayModel
 @{
 	var classIconPath2X = "";
 	var classIconPath4X = "";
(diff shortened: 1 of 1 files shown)


Masterjun3:
For now, this adds too many new buttons. The goal seems to be more clarification for newbies. In that case, "less is more" applies. It also makes things more convenient for mobile design, since space is very limited.

The main draw of the publication page is obviously the publication image. (The movie description naturally takes a lot of space so its prominence speaks for itself.)
The next thing that should be immediately visible is a way to *watch* the movie with the *movie rating* nearby. "Watching options" isn't necessary if we replace it with something known like "Watch on YouTube".
Next priority would be both the *submission* (more information) and *discussions*. Those are currently a bit too far away from the main point of focus (which are always image and video).

After this the other elements such as movie file, emulator, additional watching options, publication information, game information, administrative tools, etc. can be displayed in a way to not distract from the main priorities. There are too many distracting buttons that either have no use (emulator downloads for mobile viewers), or don't need the prominent spot (.mp4 download).

YoshiRulz:
Could make it `Watch | ▾` with the right half being a dropdown. https://getbootstrap.com/docs/5.0/components/button-group/#nesting

ThunderAxe31:
> For now, this adds too many new buttons. The goal seems to be more clarification for newbies. In that case, "less is more" applies. Which also makes things more convenient for mobile design, since space is very limited.

I agree. I see that the sheer amount of options can be overwhelming, so even if the text "Watching options" and "YouTube:" make it clear to a newbie where can find what they need the most, that doesn't mean that they would instantly understand that's what they want to click. In that case, let's just yeet the redundant stuff. Let's use YouTube for both embed and external link, and .mp4 for download from Archive.org. Let's also remove the mention to the .mp4 extension, as it's the most common and compatible format nowadays. In fact, I can watch the .mp4 encodes just fine in my smartphone's preinstalled player.

I've made more mockup screenshots in Paint. First, there is how my new and improved idea would look on PC. Further below, there is a comparison between the current smartphone version and how my new idea would look like on the same device (Samsung Galaxy S5 Mini, resolution: 720 x 1280). Note that I didn't update the pr code yet.


Here is a list of changes:

- As I said, I left only the 3 watching options mentioned above.
- I moved the ratings on the side of the movie title, because it makes sense while saving space.
- I moved "Submission text" and "Discuss" to the bottom right of the description box, because it makes sense to be interested about these extra informations after having read the movie description, not before that.
- I renamed "Submission text" to "Author notes" for intuitiveness.
- I moved the publication date on bottom of the movie text, because it's not the most important thing at it was wasting little but precious space.
- I moved the download links for emulator and movie file along with other technical publication info.
- I changed "Emulator version:" to just "Emulator:" because it's not just the emulator version that we're pointing out, since there are other emulators beside BizHawk.

> Could make it `Watch | ▾` with the right half being a dropdown. https://getbootstrap.com/docs/5.0/components/button-group/#nesting

I think that would not make it simpler, moreover after having decreased to 3 watching options. I may be wrong, but anyway I think that the problem is already solved at this point.

nattthebear (changes requested):
Cherry pick this to be against the new default branch `name`.

PR #663, opened 2021-12-16
fix #660 link to forum from news post headings

### LINKED ISSUE #660: News links on index page do not really behave like links
These links toggle content dynamically on click:
image

But opening them in a new tab/window (e.g. via the context menu or middle mouse click) just opens the same page, without the contents for the entry being expanded.

TiKevin83:
fixes #660 
Points news post header links to the forum post while preserving toggle behavior for basic clicks
Wires script to auto-open dropdowns for any other cases where this is used
Adds back link to author in news post

### DIFF
--- a/TASVideos/wwwroot/js/site.js
+++ b/TASVideos/wwwroot/js/site.js
@@ -74,7 +74,13 @@ if (localStorage.getItem("style-dark") !== null) {
 	}
 }
 
-if (location.hash != null && location.hash != "") {
-	document.querySelector(`[href='${location.hash}']`).setAttribute("aria-expanded", "true");
-	document.querySelector(location.hash + '.collapse').classList.add("show");
+if (location.hash) {
+	let expandButton = document.querySelector(`[href='${location.hash}']`);
+	if (expandButton) {
+		expandButton.setAttribute("aria-expanded", "true");
+	}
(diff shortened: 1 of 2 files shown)


meshuggahtas:
Tried to test it locally, but I have zero news article with Demo and Postgres Sample.

TASVideos/wwwroot/js/site.js
```diff
@@ -73,3 +73,8 @@ if (localStorage.getItem("style-dark") !== null) {
 		forceDarkMode();
 	}
 }
+
+if (location.hash != null && location.hash != "") {
```
adelikat, line 77: if (location.hash) is cleaner here

brunnerh: I do not know the policy for browser support is here, but if only support for recent browsers is required that could also be simplified using optional chaining (`?.` operator). [(Support Info)](https://caniuse.com/mdn-javascript_operators_optional_chaining)

TiKevin83: We're targeting ES6 minimum support, optional chaining is a tad too new, looks like it was part of ES2020. We should probably implement babel+uglify at some point to let us code with newer features while targeting ES6, but at the same time ES6 is pretty clean already

PR #681, opened 2021-12-21
Recategorize Homepage Nav Menu

### LINKED ISSUE #680: Restore the "About" option in the navigation
On the old page it's a tab linking our information page "WelcomeToTASVideos".
This option is currently hidden in the "Home" dropdown (which very much shouldn't be a dropdown, but that can be a different issue), and is poorly titled "Welcome". It should be "About" like on the old page and it's own option, preferably at the very right end of the list.

We do have the Welcome banner, but it only appears on the frontpage. A lot of links are to publications where the banner doesn't appear.

This tab is incomparably important to absolute first timers to our page. So important in fact that I personally count it as go-live critical.

TiKevin83:
fixes #680

### DIFF
--- a/TASVideos/Pages/Shared/_NavBarPartial.cshtml
+++ b/TASVideos/Pages/Shared/_NavBarPartial.cshtml
@@ -1,29 +1,25 @@
 <navbar class="flex-wrap me-2 me-md-0 fw-bold">
-	<nav-dropdown id="HomeDropdown" activate="Home">
-		<a id="Home" class="dropdown-item" href="/">Home</a>
-		<a id="Welcome" class="dropdown-item" href="/WelcomeToTASVideos">Welcome</a>
-		<a id="News" class="dropdown-item" href="/News">News</a>
+	<nav-item activate="Home">
+		<a id="Home" class="nav-link" href="/">Home</a>
+	</nav-item>
+	<nav-dropdown id="MoviesDropdown" activate="Movies">
+		<a id="Publications" class="dropdown-item" href="/Movies">Publications</a>
+		<a id="Submissions" class="dropdown-item" asp-page="/Submissions/Index">Submissions</a>
+		<a id="Userfiles" class="dropdown-item" href="/Userfiles">User Files</a>
 	</nav-dropdown>
-	<nav-item activate="Movies">
-		<a id="Movies" class="nav-link" href="/Movies">Movies</a>
-	</nav-item>
-	<nav-item activate="Game Resources">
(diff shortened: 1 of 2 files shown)


TiKevin83:
image

ThunderAxe31:
Looks excellent. My only objection is for using the vague term "chat". People may mistakenly expect it to be a site built-in chat, and thus be discouraged from joining, for various reasons. I think it would be much better to either write "Discord" or put the Discord icon (with varying size depending on client system, like the rest). https://discord.com/assets/847541504914fd33810e70a0ea73177e.ico

vadosnaprimer:
> Looks excellent. My only objection is for using the vague term "chat". People may mistakenly expect it to be a site built-in chat, and thus be discouraged from joining, for various reasons. I think it would be much better to either write "Discord" or put the Discord icon (with varying size depending on client system, like the rest). https://discord.com/assets/847541504914fd33810e70a0ea73177e.ico

What about renaming it to Chats? We still host IRC and stuff.

meshuggahtas:
The legacy site has **Chat** in the menu and **Live Chat** after clicking it. I think chat is a known word internationally (because of chat clients, chat rooms, etc) and describes it's intended feature enough. 

Personally, **Chats** sounds bizarre. Just googled it and there is a "The Chats" artist, Viber Chats and another one. Every other result is Chat for me.

meshuggahtas:
I have a very beginning draft of important links, it looks like this (ignore the order of submenu):
WelcomeToTASVideos (About @ menu)
- Community
- Users
- SiteHistory
- Helping

Rules and guidelines
- SiteRules
- TextFormattingRules
- MovieRules

Game informations
- Games
- GameResources

From http://tasvideos.org/MESHUGGAH/CheatSheet.html

ThunderAxe31:
> What about renaming it to Chats? We still host IRC and stuff.

IRC is dying, while Discord is growing. Most people that still hang around on our IRC channel (which are few and decreasing) are also present in the Discord server.

But hey, if you still think that IRC needs to be mentioned, then let's display like this in the navbar: 847541504914fd33810e70a0ea73177e-000/IRC
This way we can have both hyperlinks while saving even more space!

Masterjun3 (approved):
This is a good and important change. It should be merged before going live. More navbar changes can come later after more discussions. Ah, and this resolves #680.

vadosnaprimer:
> But hey, if you still think that IRC needs to be mentioned, then let's display like this in the navbar: 847541504914fd33810e70a0ea73177e-000/IRC This way we can have both hyperlinks while saving even more space!

IMO it visually feels worse than the word Chat.

vadosnaprimer (approved):

PR #747, opened 2022-01-03
parse VS system

CasualPokePlayer:
Leaving this as draft as I don't know if it's settled on whether to marked these as VS or Arcade. Also a db update would be needed for a VS id it seems.


vadosnaprimer:
What needs to be decided is whether we make VS a new system on the site, or just a game version. Since VS ROMs seems to be just NES roms (and it decides how to load them using iNES header?), leaving them as game versions seems logical to me.

CasualPokePlayer:
> What needs to be decided is whether we make VS a new system on the site, or just a game version. Since VS ROMs seems to be just NES roms (and it decides how to load them using iNES header?), leaving them as game versions seems logical to me.

What makes an NES ROM an NES ROM is that it is a ROM image that came from an NES cart (or could be put on an NES cart and loaded correctly). VS ROMs are definitely not that. iNES header just has info like PRG and CHR size which are needed to be provided for the emulator to properly use the ROM (which otherwise need some gamedb to provide them), and are identical to the "Arcade ROMs" after just removing the header.

CasualPokePlayer:
So here's an interesting monkey wrench for this debate;
image
As the VS system is an Arcade you could use MAME-RR for TASing games on that system. I don't know why you would, but you can. And whenever MAMEHawk is finished that will be an option too. Perhaps just marking VS TASes on NESHawk as Arcade is the best way to go here.

meshuggahtas:
- Should distinguish VS and PC10 (Playchoice-10) from NES and Arcade
- I advise adding new platforms from the parent platform. NES-VS and NES-PC10. Might be useful for other platforms as well. This also makes writing rules easier to refelct on a group of NES games that aren't NES games (referring to arcade rules that doesn't happens on home console NES games)
- Obvious but here it is, VS and PC10 should be tased with NESHawk. In sense of TASing, those VS and PC10 ROM structures and technical details matches NES better than MAMEHawk.

vadosnaprimer:
Okay yes, VS. games work in MAME just fine. Some have dual screen (Baseball and Tennis). It's true that VS games have VS in their name already. If somebody plays them in MAME-RR or MAMEHawk, the system will be parsed as Arcade, and I don't think this should be treated as wrong.

Dang every option feels right to some degree. @adelikat do you have a preference?

adelikat:
We need to add the VS system Id first.
Need Unit Tests to approve.

SamsaraTAS:
My only major preference here is that VS TASes done on any emulator (NESHawk, FCEUX, MAMEHawk, etc) should always have the same ID, whatever it ends up being. I can go either way on that, with a slight preference for VS or NES-VS, though I'd imagine that would lead to a lot of fixing "NES-VS VS Super Mario Bros" in submissions. Occam's razor here is probably to parse it as Arcade, and I wouldn't argue against that if there's more of a consensus for it.

vadosnaprimer:
Yeah I feel if we don't put it as Arcade and somebody TASes it on MAMEHawk, we'd have to tweak the parser as well as MAMEHawk movie header logic quite a bit, which in turn feels like adding complexity where it's not required.

CasualPokePlayer:
There seems to be releative agreement on using Arcade for VS TASes? Which case this PR can continue (tests are in always, and VS system ID doesn't need to be added).

vadosnaprimer (approved):

vadosnaprimer:
This should be good to merge I think.

adelikat:
It isn't good to merge because we don't have a VS system id, so this would return an unrecognized code that would then crash the submission/userfile upload process

CasualPokePlayer:
> 

It's set to change the system ID to Arcade and override framerate with NES framerate. A test is included for this case.

PR #760, opened 2022-01-03
Add dropdown icons to SubPage listings and collapsible TopicFeeds

Masterjun3:
Nobody asked for this but me.

Before | After
image


vadosnaprimer:
Good idea! It doesn't act the same as other textual links, but rather like an expandable spoiler, and those already on the site have this icon.

adelikat (approved):

PR #778, opened 2022-01-04
Restore Author Style to "X, Y & Z" and display additional authors

### LINKED ISSUE #495: Authors without an account are not listed in tables
Example: https://tasvideos.mistflux.net/Submissions/Index?Search.StatusFilter=6&Search.Years=2021&Search.User=mkdasher

Somewhat related to #431

### LINKED ISSUE #773: Change Author display from "X, Y, Z" to "X, Y & Z"
Like on the old site.
I'm currently implementing this.

Masterjun3:
Resolves #495.
Resolves #773.

Publications:
image
Table view:
image


Masterjun3:
Because the publication and submission titles are stored in the database, they will only be affected if they're regenerated.
This can be done with a simple edit or similar, but technically needs to be done to every single one if they're to be fixed.

TASVideos.Common/Extensions/StringExtensions.cs
```diff
@@ -188,5 +188,17 @@ public static string ToSnakeCase(this string str)
...
+		public static string LastCommaToAmpersand(this string commaString)
+		{
+			int lastComma = commaString.LastIndexOf(",");
+
+			if (lastComma == -1)
+			{
+				return commaString;
+			}
+
+			return commaString[..lastComma] + " &" + commaString[(lastComma + 1)..];
```
TiKevin83, line 201: I know adelikat mentioned he can clean this up but this really should be a helper method elsewhere that does this in the same stroke as the join operation and probably using string remove and insert instead of the array ranges with inferred start/end

TiKevin83:
Love the visual improvement, the code for the helper extension could use some cleanup but it's a great improvement

TASVideos.Common/Extensions/StringExtensions.cs
YoshiRulz: Code suggestion (untested):
```csharp
public static string JoinListWithEndingAmpersand(this IEnumerable<string> list) {
	if (!list.TryGetNonEnumeratedCount(out var count)) return list.ToList().JoinListWithEndingAmpersand();
	switch (count) {
		case 1:
			return list.First();
		case 2:
			return $"{list.First()} {list.First()}";
		default:
			var head = list.Take(count - 1);
			return $"{string.Join(", ", head)}, and {list.First()}";
	}
}
```

Invariel:
I appreciate the Oxford comma, I disapprove of functions that don't do what they're named.

vadosnaprimer:
> I appreciate the Oxford comma.

The consensus for the old site was that oxford comma only looks good when you have "and" before it, but looks weird if it's ", &".

TASVideos/Pages/Shared/Components/FrontpageSubmissionList/Default.cshtml
```diff
@@ -14,7 +14,7 @@
 				<td><timezone-convert asp-for="@item.Submitted" date-only="true" /></td>
 				<td>@item.System @item.GameName @(!string.IsNullOrWhiteSpace(item.Branch) ? $"\"{item.Branch}\"" : "")</td>
 				<td><sub-link id="item.Id">@(item.Time.ToStringWithOptionalDaysAndHours())</sub-link></td>
-				<td>@item.Author</td>
+				<td>@(string.Join(", ",(item.Authors ?? Enumerable.Empty<string>()).Concat((item.AdditionalAuthors ?? "").SplitWithEmpty(","))).LastCommaToAmpersand())</td>
```
adelikat, line 17: All of this logic is complicated and repeated, I'd like to see it in a method somewhere

PR #780, opened 2022-01-04
adjust contrast of anchors to meet AA WCAG

### LINKED ISSUE #762: Forum thread titles and page numbers can be hard to read in dark mode
This might be a nitpick, but I find that the dark blue text can be hard to read in the dark theme. Forum titles are generally okay but the page numbers are a bit hard for me to read sometimes. I find that making the text brighter and a bit more saturated makes a big difference in how readable it is, without messing up the site's overall theme.

Here is a before and after example:
image

TiKevin83:
fixes #762


vadosnaprimer:
Pics?

PR #782, opened 2022-01-04
use bootstrap's font stack

### LINKED ISSUE #691: Font stack isn't being used
I noticed the new website shows up as using Helvetica for me personally (this is the default sans-serif font on macOS). However, when I checked the CSS I saw that there's a different font stack defined:

--bs-font-sans-serif: system-ui,-apple-system,"Segoe UI",Roboto, "Helvetica Neue", Arial,"Noto Sans","Liberation Sans",sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol","Noto Color Emoji"

I think this is a really nice font stack, but it doesn't get used because it gets overwritten by another rule:

Screen Shot 2021-12-29 at 13 58 17

When I disable the `body: sans-serif;` rule the font stack comes through.

TiKevin83:
fixes #691 

As far as I can tell reading around, bootstrap's font stack is much better for accessibility than browser defaults, and can be overridden if really necessary

### DIFF
--- a/TASVideos/wwwroot/css/partials/_customizations.scss
+++ b/TASVideos/wwwroot/css/partials/_customizations.scss
@@ -10,7 +10,6 @@
 }
 
 body {
-	font-family: sans-serif;
 	font-size: 14px;
 	background: var(--tasvideos-bg) linear-gradient(var(--tasvideos-bg-gradient) 0, var(--tasvideos-bg) 250px) repeat-x;
 }


TiKevin83:
Before:
image

After (Chrome Win10):
image

adelikat:
Accessibility is the priority here.  I'll trust your reading that BS 5 is better for accessibility (which is a sad statement for modern browsers, but believable)

PR #784, opened 2022-01-04
fix primary and secondary button hover box shadows

### LINKED ISSUE #592: Bootstrap secondary colors are staying gray
The `box-shadow` property of the secondary buttons is still gray (from the default secondary color). Other buttons have their shadow appropriately colored.

image

Also, giving an object the `border-secondary` class will give a gray border when it should be recolored as well.

TiKevin83:
fixes #592 
I'm leaving the border part as a wontfix unless we feel that's really necessary


vadosnaprimer:
Pics?

TiKevin83:
compare to image in issue
image

vadosnaprimer:
Ah looks great!

PR #785, opened 2022-01-04
Fix table border color

### LINKED ISSUE #659: Tone down table borders?
The borders seem unnecessarily salient. For the light theme I would go from gray-600 at least to gray-500:

Before:
image

After:
image

For the dark theme I would go from gray-600 to gray-700. (Using a custom property and just reassigning that for the dark scheme might be the easiest here)

Before:
image

After:
image

Personally, I would also just drop the vertical lines completely because they do not contribute much (it is easy to see which cell contents belong to which column).

image

In general any additional line distracts from the contents and contributes visual noise. Even the horizontal lines are arguable because of the existing alternating color scheme which does most the work of separating the rows.

TiKevin83:
fixes #659

### DIFF
--- a/TASVideos/wwwroot/css/partials/_customizations.scss
+++ b/TASVideos/wwwroot/css/partials/_customizations.scss
@@ -56,7 +56,7 @@ article.wiki, .postbody {
 		th, td {
 			padding: 0.3rem;
 			vertical-align: top;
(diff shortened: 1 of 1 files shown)


vadosnaprimer:
Pics?

TiKevin83:
Before:
image

After:
image

Before Light:
image

After Light:
image

PR #796, opened 2022-01-04
Show active navbar item

Masterjun3:
image

### DIFF
--- a/TASVideos/TagHelpers/NavbarTagHelpers.cs
+++ b/TASVideos/TagHelpers/NavbarTagHelpers.cs
@@ -102,7 +102,7 @@ namespace TASVideos.TagHelpers
 
 			switch (Activate)
 			{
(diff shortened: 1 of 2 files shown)


vadosnaprimer:
Does it make sense to make the rest also a bit lighter? I have this complaint since the very beginning but I don't remember if I told anyone :D

TiKevin83:
I'll check it against accessibility

TiKevin83:
Can you add this to `bootstrap-overrides.scss` to hit WCAG AAA on the contrast in the navbar?
```
.navbar-dark .navbar-nav .nav-link {
	color: RGBA(255,255,255,.6);
}
```

Masterjun3:
Now looks like this:
image

TiKevin83 (approved):

adelikat (approved):

Issue #793, opened 2022-01-04
Disable URL-autolinkification within [url=URL] BBCode tags


SandTAS:
The BBCode parser should not auto-link URLs when they are inside [url=URL][/url] tags. Otherwise, you get nested `a` HTML elements, which is [against spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) ("there must be no ... `a` element descendant").

For example, the BBCode

```
[url=https://outer.example.com/]hello https://inner.example.com/ world[/url]
```

should become

```
<a href="https://outer.example.com/">hello https://inner.example.com/ world</a>
```

not

```
<a href="https://outer.example.com/">hello <a href="https://inner.example.com/">https://inner.example.com/</a> world</a>
```

See https://tasvideos.org/Forum/Posts/501179#501179 for an example of where this goes wrong. The old forum did not have URL auto-linkification, so if one wanted a URL that was also a link, one had to do something like `[url=URL]URL[/url]`. The example in this post is

```
[url=https://pad.riseup.net/p/u2s2Ko2T7_xD3aBW_N8w]https://pad.riseup.net/p/u2s2Ko2T7_xD3aBW_N8w[/url]
```

which renders as

```
<a href="https://pad.riseup.net/p/u2s2Ko2T7_xD3aBW_N8w"><a href="https://pad.riseup.net/p/u2s2Ko2T7_xD3aBW_N8w">https://pad.riseup.net/p/u2s2Ko2T7_xD3aBW_N8w</a></a>
```
```

**Comment by nattthebear** [comment]
> The old forum did not have URL auto-linkification

Yes, it did.

**Comment by nattthebear** [comment]
This also affects explicit urlification, such as
```
1[url=https://outer.example.com/]A[url=https://inner.example.com/]B[/url]C[/url]2

SandTAS:
> > The old forum did not have URL auto-linkification
> 
> Yes, it did.

My mistake; I guess you are right. I found https://web.archive.org/web/20170903052220/http://tasvideos.org/forum/viewtopic.php?t=7860#275096 on the old forum that appears to have auto-linked URLs. I don't know why I thought I had to use `[url]` then.

nattthebear:
> I don't know why I thought I had to use [url] then.

It may have been broken on some inputs, or may have not always existed as a feature in the old system.  That code base was Something Special.

PR #802, opened 2022-01-04
Fix URL nesting in forum parser

nattthebear:
Fixes https://github.com/TASVideos/tasvideos/issues/793.

`1[url=https://outer.example.com/]A[url=https://inner.example.com/]B[/url]C[/url]2` is easily rejected by using the same nesting logic as we were already using for some other tags, but I needed to refactor because `private enum ParseState` was becoming unwieldly as it was essentially two separate enums crossed together and it would have needed a new variant for this.

`[url=https://outer.example.com/]hello https://inner.example.com/ world[/url]` is rejected by adding code specific to it.  Since URL auto-linkification is a one-off thing and doesn't follow the normal flow, I thought it was appropriate to fix in the one-off flow.  We're probably not adding any more implicit tags...

### DIFF
--- a/TASVideos.ForumEngine/BbParser.cs
+++ b/TASVideos.ForumEngine/BbParser.cs
@@ -68,86 +68,93 @@ namespace TASVideos.ForumEngine
 
 		private static readonly Regex HtmlVoid = new (@"\G\s*([a-zA-Z]+)\s*\/?\s*>");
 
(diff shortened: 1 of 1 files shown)


nattthebear:
Could use some eyes on this, don't have much time to test it right now.

adelikat (approved):

PR #812, opened 2022-01-04
Add checkbox for watching/unwatching topics when creating posts or topics

### LINKED ISSUE #766: Topic watch is always set upon replying to a topic
This was an optional feature in the old site code, and could be turned on or off in your user profile - not everyone likes receiving emails any time somebody posts in a topic that they posted in.

Currently, this cannot be turned off, which means that any time you make a post, you _will_ receive an email if someone else posts in that topic - unless you manually delete the topic watch in your profile settings every time after making a post, which is a lot of extra hassle.

Masterjun3:
Resolves #766.


adelikat:
I was imagining this being a user setting that you can just set once

Masterjun3:
This wasn't a user setting on the old site, and it would be very restrictive to either always add them, or never.

adelikat (approved):

PR #815, opened 2022-01-05
Rename list of rom hashes to "Game Versions"

### LINKED ISSUE #800: "Roms" header on the publication page should be renamed to "Game Versions" or something similar.
When you search for a game and click on it, you're taken to this page, which has a section for "Roms".

image

I've been told this is only meant to show ROM hashes, which is fine, but I think the name is confusing. We don't want TASVideos to be associated with piracy and we don't want people thinking that ROMs are being distributed here. Having a list of ROM hashes works fine when the header is called "Game Versions" or something similar.

Masterjun3:
Resolves #800.

### DIFF
--- a/TASVideos/Pages/Games/Index.cshtml
+++ b/TASVideos/Pages/Games/Index.cshtml
@@ -86,7 +86,7 @@
 </row>
 <hr />
 <h4>
-	Roms
+	Game Versions
 	<a permission="CatalogMovies"
 	   asp-page="/Games/Roms/List"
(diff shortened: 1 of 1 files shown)


Masterjun3:
I'm open for other suggestions, but in my opinion the Game pages are one of our "main" pages, so it shouldn't mention "ROM" at all, even if other places and a lot of the code calls them Roms.

TiKevin83 (approved):

PR #817, opened 2022-01-05
add a global filter that attempts to debounce duplicate POST requests

adelikat:
This should help prevent double clicks as well as any mysterious duplicate posts that seem to be happening when the server is under high load and the post either takes too long or times out


TiKevin83:
Can you also have it ensure the user hasn't sent an edit previously within the last 20-30 seconds?

Masterjun3:
Does this use the same `__RequestVerificationToken` like the one injected into the site? I browsed the site for a bit but that specific token never changed. If it doesn't change, wouldn't that mean you're limited to 1 action per minute with this?

Masterjun3 (approved):

adelikat:
> 

It caches for 1 minute, at which point that value would be "unblocked"

But if you mean other edits that aren't this one, it isn't too relevant.  The forgery token will be a different value for each edit

PR #822, opened 2022-01-05
Fix JS crash in forum highlighted posts

### LINKED ISSUE #810: Javascript error with link to post
I visited this url https://tasvideos.org/Forum/Topics/23005?CurrentPage=1&Highlight=511306#511306

Then I opened the browser debugger and I noticed an error:
```
Uncaught DOMException: Failed to execute 'querySelector' on 'Document': '#511306.collapse' is not a valid selector.
    at https://tasvideos.org/js/site.js:24:33
```
Which points to line 24, column 33: `let expandedContent = document.querySelector(location.hash + '.collapse');`

I'm using Opera 82.0.4227.43, Windows 10.

TiKevin83:
fixes #810 
validates that the location hash is an id which can be selected before moving on to trying to ensure it is shown if collapsed.

### DIFF
--- a/TASVideos/wwwroot/js/site.js
+++ b/TASVideos/wwwroot/js/site.js
@@ -21,8 +21,11 @@ if (location.hash) {
 	if (expandButton) {
 		expandButton.setAttribute("aria-expanded", "true");
 	}
-	let expandedContent = document.querySelector(location.hash + '.collapse');
-	if (expandedContent) {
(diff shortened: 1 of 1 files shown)


nattthebear (changes requested):
`location.hash` always starts with a `#`

nattthebear (approved):

PR #835, opened 2022-01-06
fix saving user settings with the same email

### LINKED ISSUE #832: Unable to change profile settings - This email cannot be used.
https://tasvideos.org/Profile/Settings

I believe the profile settings page tries to save every data but a user's current email will be already in use by themselves.

TiKevin83:
fixes #832


meshuggahtas (approved):
Tested with Dev Postgres Sample and works, the code also looks legit (acceptable to merge and close pr) but I let someone else more experienced to do it confidentally.

PR #857, opened 2022-01-07
Fix minimovie height on homepage

### LINKED ISSUE #794: Latest Publications on the home page shrinks if the featured publication is too short
If the featured publication is too short, the header for "Latest Publications" will be higher up than it normally is, which makes the table and screenshot smaller than they should be.

image

Compare that with how it looks when the featured publication is just a tiny bit taller.
image

I'm told this is by design but I think the latest publications header should always be like it is in the 2nd screenshot. When it's scrunched up on the left side, it ruins the table and the screenshot is too small to see well.

TiKevin83:
(it makes the height of the element contain the height of the floated element eg the news)

fixes #794


Masterjun3 (approved):

vadosnaprimer (approved):
If it's what I think it is then

PR #858, opened 2022-01-07
Use the Mood Avatar System for Open Graph Embeds

### LINKED ISSUE #852: Make discord embeds display mood avatar of post rather than default avatar
Instead of using a forum account’s main avatar, use the avatar that they’re currently using for that post.

TiKevin83:
fixes #852


adelikat:
Yaaaaaaaassss

PR #871, opened 2022-01-08
Change latest posts page to show the past 3 days worth of posts

Masterjun3:
Shows 3 days worth of posts instead of the latest 100 posts.
I'm not sure if this optimization works, but instead of sorting ~500000 posts and taking the first 100, it now first takes ~100 posts and then sorts only those.


adelikat:
I'm pretty sure this isn't faster.  Ordering by id is fast because it is an index, and taking x records is fast.  Comparing dates is not fast.

However, this is better functionality, and the query speed should be plenty fast either way, so it is a solid improvement.

PR #872, opened 2022-01-08
RejectedSubmissions module: sort list and restore total counts

### LINKED ISSUE #805: RejectedSubmissions module orders submission in a funny way
On old site, they used to be ordered from the smallest submission #number to the biggest one.

The affected wiki module is placed by writing `[module:RejectedSubmissions]`

You can see it in https://tasvideos.org/RejectedSubmissions

Masterjun3:
Resolves #805.

Restores total counts:
image

### DIFF
--- a/TASVideos/Pages/Shared/Components/RejectedSubmissions/Default.cshtml
+++ b/TASVideos/Pages/Shared/Components/RejectedSubmissions/Default.cshtml
@@ -1,4 +1,4 @@
-﻿@model IEnumerable<RejectedSubmission>
+﻿@model ICollection<RejectedSubmission>
 @{ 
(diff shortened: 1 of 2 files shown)


TASVideos/Pages/Shared/Components/RejectedSubmissions/Default.cshtml
```diff
@@ -1,4 +1,7 @@
 ﻿@model IEnumerable<RejectedSubmission>
```
adelikat, line 1: Change this to ICollection, and use the .Count property instead of method.

Since you are enumerating multiple times (foreach and Count()), it is better to use a contract that ensures this is safe

adelikat (changes requested):
See comment

Masterjun3:
Alright, made the change.

adelikat (approved):

PR #877, opened 2022-01-08
Make posts catch email notify exceptions

Masterjun3:
Posts should not rely on the email notification to succeed to finish processing the request.

Email exceptions could have been a reason for the weird posting behavior users were having, resulting in double posts and blank pages.


Masterjun3:
Dear Pull Request
Please update, I pushed to the branch but it isn't showing up here.
Thanks.

PR #896, opened 2022-01-09
Show total userfile count next to author

### LINKED ISSUE #895: Put user files amount and link in a user file page
Just as it used to be on old site. Example: https://tasvideos.org/UserFiles/ForUser/ThunderAxe31

Current:
image

Desired:
image

Note that you can see files amount and url from the [Profile page](https://tasvideos.org/Profile), but I'm too much noob to make the PR myself, sorry.
image

Masterjun3:
Resolves #895.

### DIFF
--- a/TASVideos/MappingProfile.cs
+++ b/TASVideos/MappingProfile.cs
@@ -71,7 +71,7 @@ namespace TASVideos
 
 			CreateMap<UserFile, UserFileModel>()
 				.ForMember(dest => dest.Author, opt => opt.MapFrom(src => src.Author!.UserName))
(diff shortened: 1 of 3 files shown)


Masterjun3:
Hold on!
This shows all userfiles, instead of just non-hidden ones like on the profile.

Masterjun3:
It now counts only userfiles that aren't hidden. This requires calling Count() which is probably slower.

adelikat:
It is calling Count() at the IQueryable level, which is fine, it will just figure out some sql that will include it in the query at a minor speed cost

adelikat (approved):

Masterjun3:
These changes were made in pull request #898.

PR #901, opened 2022-01-10
Add "Show more" taghelper and use it for Author Notes in submission topics

Masterjun3:
Having long blocks of content has disadvantages when browsing pages.
This is why those blocks are usually limited in height, and their content inside scrollable.
However, this nested scrolling is very mobile-unfriendly and even desktop users can have problems dealing with multiple scrollbars.

This "Show more" solution is often used to avoid long blocks but also make it mobile-friendly.

https://user-images.githubusercontent.com/22375320/148708862-06369a00-d6c7-4194-876b-42aa1c97838b.mp4

The TagHelper features:
- custom maximum height
- automatic detection if a "Show more" is needed or if the full content fits inside
- page position adjustment when clicking "Hide"
- old scrollbar method if javascript is disabled

### DIFF
--- a/TASVideos/Pages/Forum/Topics/Index.cshtml
+++ b/TASVideos/Pages/Forum/Topics/Index.cshtml
@@ -105,9 +105,11 @@
 				<sub-link class="col-auto btn btn-secondary btn-sm" id="@SubmissionHelper.SubmissionId(Model.WikiPage.PageName)"><i class="fa fa-info-circle"></i> Full Submission Page</sub-link>
 			</row>
 		</cardheader>
-		<cardbody style="max-height: 600px; overflow-y: scroll">
-			<wiki-markup markup="@Model.WikiPage.Markup" page-data="Model.WikiPage"></wiki-markup>
-		</cardbody>
+		<show-more max-height="38rem">
+			<cardbody>
+				<wiki-markup markup="@Model.WikiPage.Markup" page-data="Model.WikiPage"></wiki-markup>
+			</cardbody>
+		</show-more>
 	</card>
 }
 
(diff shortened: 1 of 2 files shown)


TiKevin83:
https://github.com/TiKevin83/tasvideos/tree/showmore-taghelper
I did a small refactor to start integrating the concept with bootstrap's built-in collapse instead of adding javascript. Still need to add the custom height setting for the taghelper back in and autodetect whether show/hide is necessary (which I think should be done server side)

TiKevin83:
The code in my refactor should be at feature parity with Masterjun's now, integrates much more cleanly with the native bootstrap collapse system.

Masterjun3:
> The code in my refactor should be at feature parity with Masterjun's now

Keeping the same scrolling position when "hiding" doesn't work. You're just thrown to the end of the page or pretty far down.

TiKevin83:
ah I see, this is a case where I would want anchor over button because the anchor reference can fix the position to the top of the expanded content on each toggle.

TiKevin83 (changes requested):
Requesting pulling the refactored code that integrates with our existing collapse system

Masterjun3:
The behavior is really weird in the version in your branch. We probably don't want to have a feature with flashing as bad as this.

https://user-images.githubusercontent.com/22375320/149029441-f82ed590-1f4b-4cd8-a70a-11c770768a6e.mp4

TiKevin83:
I'm not seeing that flickering behavior on my branch, but I did have to make another change to get collapsing and showing to navigate to the top of the collapsed content, and it adds even more JS code so I'm less happy with integrating with Bootstrap's collapse.

TiKevin83 (approved):

PR #904, opened 2022-01-10
Small redesign of Game page

### LINKED ISSUE #903: Game page bugs
1. "Screenshot Url" does not seem to do anything. Screenshots were added for [SMB](https://tasvideos.org/1G) and [Bio Hazard](https://tasvideos.org/2300G) but do not appear on the Game page on mobile or desktop.
2. In the Game Versions/Roms section, setting the type to anything except Unknown changes to Unknown anyway when you click Update (to save the page).
3. Back on the main Game page, the Roms section has multiple Roms stacking horizontally. Notice the "Pocket Monsters" column on the right. This is especially hard to read on mobile. Is this intended or a bug?
image

Masterjun3:
Also fully resolves #903.

- Shows MiniMovie of all current publications of the game in tabs
- Show screenshot. If no screenshot is set show screenshot of a current publication as a fallback
- Fix game version table
- Actually save game version type (by renaming RomType to Type so automatic mapping works)

Compare with https://tasvideos.org/99G
image

### DIFF
--- a/TASVideos/Pages/Games/Index.cshtml
+++ b/TASVideos/Pages/Games/Index.cshtml
@@ -80,7 +80,7 @@
 			@foreach (var movie in Model.Movies)
 			{
 				<li class="nav-item">
-					<a class="nav-link@(movie==Model.Movies.First() ? " active" : "")" href="#tab-@(movie.Id)M" data-bs-toggle="tab">@(string.IsNullOrWhiteSpace(movie.Branch) ? "(branchless)" : $@"""{movie.Branch}""")</a>
+					<a class="nav-link@(movie==Model.Movies.First() ? " active" : "")@(string.IsNullOrWhiteSpace(movie.Branch) ? "" : " fw-bold")" href="#tab-@(movie.Id)M" data-bs-toggle="tab">@(string.IsNullOrWhiteSpace(movie.Branch) ? "(baseline)" : $@"""{movie.Branch}""")</a>
 				</li>
 			}
 		</ul>
@@ -93,6 +93,7 @@
 			</div>
 		}
 	</div>
+	<small condition="@string.IsNullOrWhiteSpace(Model.Movies.First().Branch)" class="text-muted">The baseline tab shows the default movie beating the game as fast as possible without any special conditions.</small>
 </div>
 <hr />
 <row>
--- a/TASVideos/Pages/Games/Index.cshtml
+++ b/TASVideos/Pages/Games/Index.cshtml
@@ -93,7 +93,7 @@
 			</div>
 		}
 	</div>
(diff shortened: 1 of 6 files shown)


vadosnaprimer:
How do these 2 look?

https://tasvideos.org/121G (blank branch, like in a LOT of games)
https://tasvideos.org/1466G (JPN and USA published alongside)

Masterjun3:
image

image

vadosnaprimer:
Yeah. When current branches are displayed as a regular movie list, or as a list of movie names, it leaves it visible what is branchless and what is different versions.

And I don't think it's feasible to get rid of the concept of branchless, because it doesn't explicitly mean anything specific: it's just the most common set of goals available in a given game. It's not any%,

I would understand tabs if there were (potentially) a lot of entries in each tab, like on speedrun.com. But we have publication history for that. In general I think it's practically better to show current branches as a list of pub modules, like on the old site.

adelikat (approved):

Masterjun3:
Fixed merge conflict.

vadosnaprimer:
@adelikat asked for a review, and the problem is highlighted on the 2 screenshots.

If we put branch labels in tabs, it becomes impossible to see why we have 2 movies that belong to different regions but both have a blank branch. Even if region is in the publication name, it won't go into tab name. And if it goes into tab name for those exceptional cases, it needs to be not in quotes. Even then it may look confusing.

The bugger problem IMO is `(branchless)`. Branchless is a casual term, it's not accurate because such a movie is still a branch. It's technically labelless, but that sounds even worse. A workaround would be calling it `trunk`, but nobody knows that as a term.

I think the old system where current movies are just listed like this
https://tasvideos.org/Movies-3273M-3653M-3768M-4010M
is better. That way you can also see which of them is Fastest Completion, which you can't see by looking at branch tabs. I would personally not mind just having a list of pub names that work as lings to actual pubs, but that's up to a discussion.

And I'm not requesting changes because **this needs more opinions**.

Masterjun3:
This isn't supposed to be the final redesign of the game page anyway.
That sort of discussion should go into a separate issue about the Game page design as a whole, and what the page is meant to be.

Reminder that this pull request fixes simple things like the table rows, showing the screenshot, and other small improvements, *that are still broken on the site*. Whether "(branchless)" is the one truly correct term or tabs are the correct way of displaying movies shouldn't be the thing to hold up this small simple pull request.

TiKevin83:
I would be ok with "baseline branch" or "default branch" but yeah this is fine without that change either way

vadosnaprimer:
Hey `baseline` without quotes (not to be confused with an actual branch label) sounds quite nice!

ThunderAxe31:
In English, "branchless" means "without branches", which makes no sense. Either call it (no label) or (Trunk).

Masterjun3:
Changed branchless to baseline and added short explanation.

PR #919, opened 2022-01-11
do not bake in "by user" to external media posts. 

adelikat:
1) Do not bake in "by user" into messages
2) In Irc- add use name in parenthesis at the beginning of messages

Discord already puts the user on relevant messages so the by user was redundant there.


TiKevin83 (approved):

vadosnaprimer (approved):
Let's see what this does.

PR #928, opened 2022-01-12
Refactor the topicpost partial

### LINKED ISSUE #927: Use Responsive Design in TopicPost Partial
The current TopicPost partial repeats some of the data and uses display classes to show/hide for mobile and desktop widths. This should be refactored to use one set of markup but retain the existing layout style.

TiKevin83:
fixes #927 

Aims to retain the new layout while using a single set of markup instead of adaptive design with two sets of markup for mobile vs desktop

Current Mobile:
Current-Post-Mobile
New Mobile:
Post-Mobile

Current Desktop:
Current-Post-Desktop
New Desktop:
Post-Desktop

### DIFF
--- a/TASVideos/Pages/Forum/Topics/_TopicPost.cshtml
+++ b/TASVideos/Pages/Forum/Topics/_TopicPost.cshtml
@@ -48,60 +48,44 @@
 			</div>
 		</row>
 	</cardheader>
-	<cardbody class="row p-2 d-md-none align-items-center mx-0 gx-2" style="background-color: var(--bs-gray-200)">
-		<div class="col-2 float-left text-center p-0">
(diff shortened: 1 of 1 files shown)


Masterjun3:
Can you move the pronouns to the right again?

TiKevin83:
fixed that, checked signatures and that appears to just be local sample data.

Masterjun3 (changes requested):
Signatures are located directly under the post content. They should be fixed to the bottom.
Also, images don't scale down responsively to the width anymore.
The usernames aren't aligned when someone doesn't have an avatar.

TiKevin83:
Those should be fixed now as well

Masterjun3 (approved):
One small thing, I'd love it if there were a `my-auto` on the div surrounding the profile picture, so that it's centered vertically on small widths.

Heh, very good, I'm genuinely surprised you made this happen and it works as well as it does. I approve.

PR #930, opened 2022-01-12
New Pager design

Masterjun3:
(before | after)
image

Dropdowns:
image

### DIFF
--- a/TASVideos/Pages/Shared/_Pager.cshtml
+++ b/TASVideos/Pages/Shared/_Pager.cshtml
@@ -8,7 +8,7 @@
 			<i class="fa fa-sort"> <i class="fa fa-times-circle"></i></i>
 		</a>
 	</div>
-	<div class="col-12 text-center">
+	<div condition="Model.LastPage() > 1" class="col-12 text-center">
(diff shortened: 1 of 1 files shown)


Masterjun3:
I tried to change the style of the buttons so that the shadow isn't there when it's active, but I have a hard time understanding all of those css customizations, so that can be a separate thing.

Masterjun3:
- First & Second are always visible
- Second-to-last & Last are always visible
- Previous page & Current page & Next page are always visible
- [...] gets expanded when it has only 1 page

Masterjun3:
Sort reset.
In case you sorted alphabetically or by author or others and want to return to default sort and don't want to remove all sort tags from the query in the url.

vadosnaprimer:
I feel it would be better to have a text on it saying what it does.

Masterjun3:
image

ThunderAxe31:
Can we please put everything in a single row? Just to save vertical space.

Masterjun3:
I tried, doesn't work well. I want the page list to be on the bottom always. If someone with more bootstrap skills can manage that, please pr or tell me.

PR #940, opened 2022-01-12
Avoid unnecessary blank line under + button in forum index

ThunderAxe31:
Before:
image

After:
image

### DIFF
--- a/TASVideos/Pages/Forum/Subforum/Index.cshtml
+++ b/TASVideos/Pages/Forum/Subforum/Index.cshtml
@@ -11,8 +11,12 @@
 }
 
 <partial Name="_ForumHeader" />
-<a permission="EditForums" asp-page="Edit" asp-route-id="@Model.Forum.Id" class="float-end btn btn-primary btn-sm border border-warning"><i class="fa fa-pencil"></i> Edit</a>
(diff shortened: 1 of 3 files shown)


TASVideos/Pages/Forum/Subforum/Index.cshtml
```diff
@@ -11,8 +11,12 @@
 }
 
 <partial Name="_ForumHeader" />
-<a permission="EditForums" asp-page="Edit" asp-route-id="@Model.Forum.Id" class="float-end btn btn-primary btn-sm border border-warning"><i class="fa fa-pencil"></i> Edit</a>
-<a permission="CreateForumTopics" asp-page="/Forum/Topics/Create" asp-route-forumId="@Model.Forum.Id" class="me-2 float-end btn btn-primary btn-sm"><span class="fa fa-plus"></span> New Topic</a>
+<div class="row">
```
adelikat, line 14: you can use <full-row> and it does both these row and col div classes for you

adelikat (changes requested):
a small simplification to do

TASVideos/Pages/Forum/Subforum/Index.cshtml
ThunderAxe31: use what? I don't get what you mean

adelikat (approved):

PR #942, opened 2022-01-13
Move checkboxes and text fields down to submit button

### LINKED ISSUE #938: Move "minor edit" and "edit comments" close to the Save button
The editing field is to huge now, and if you're doing a lot of edits, it becomes incredibly annoying to jump back and forth with the mouse, hitting things on the opposite sides of the screen, and in my case I even have to scroll up and down (900p monitor).

Masterjun3:
Resolves #938.

Changes include minor edits box, watch topic box, and revision message field.

(before | after)
image

### DIFF
--- a/TASVideos/Pages/Forum/Posts/Create.cshtml
+++ b/TASVideos/Pages/Forum/Posts/Create.cshtml
@@ -21,16 +21,16 @@
 		</column>
 	</row>
 	<fullrow class="mt-3">
-		<input asp-for="WatchTopic" />
-		<label asp-for="WatchTopic"></label>
-	</fullrow>
-	<fullrow class="mt-3">
 		<partial name="_CreatePostHelper" model="@("Post_Text")" />
 	</fullrow>
 	<form-group>
 		<textarea asp-for="Post.Text" class="form-control" rows="20"></textarea>
 		<span asp-validation-for="Post.Text" class="text-danger"></span>
 	</form-group>
+	<fullrow class="mt-3">
+		<input asp-for="WatchTopic" />
+		<label asp-for="WatchTopic"></label>
+	</fullrow>
 	<form-button-bar>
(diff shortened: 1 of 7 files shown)


TASVideos/Pages/Forum/Posts/Create.cshtml
```diff
@@ -20,17 +20,17 @@
 			<span asp-validation-for="Post.Mood" class="text-danger"></span>
 		</column>
 	</row>
-	<fullrow class="mt-3">
```
adelikat, line 23: why did this need to change?

TiKevin83:
I like the use of the form grouping, it looks cleaner to me that way though in that screenshot something is off with the vertical centering of the checkbox

Masterjun3:
Yeah apparently [bootstrap itself](https://getbootstrap.com/docs/5.1/forms/input-group/#checkboxes-and-radios) suggests adding  `mt-0`.

vadosnaprimer:
Sometimes when making [important changes](https://tasvideos.org/Wiki/PageHistory?path=MovieRules%2FLegacy) to the wiki, I need to write elaborate edit notes. So having more space in "edit comments" is important. Can the width of that field be made to fill all available width?

Masterjun3:
They go to full width when you make your width smaller. :) :)

Masterjun3:
Real answer, it doesn't look good at full width in large screens.
Ideally we would have automatically resizing text fields, giving you multiple lines when you write a lot. But I think that's another unsolved html problem and javascript might be needed again.

vadosnaprimer:
In terms of taste, it would just be as wide as the main edit field, I don't see that as a problem. I think being functionally helpful is more important here than looking nice and tiny.

TiKevin83:
I think a good compromise would be `col-xl-8 col-xxl-6` instead of `col-lg-6`

Masterjun3:
It's now full width.

vadosnaprimer (approved):
Thank you!

PR #953, opened 2022-01-13
Send email when password is changed

### LINKED ISSUE #713: Site Should Alert Users via Email When Their Password Changes
see title

adelikat:
send email when password is changed, also redirect from password to profile page, fixes #713


Invariel (changes requested):
Please include a URL to the password change link, with the text, "If you have received this message in error, you can reset your password and reclaim your account here."

Invariel (approved):

PR #955, opened 2022-01-13
redis - catch connection and timeout on set commands too

adelikat:


Invariel (approved):
I generally don't like the inner "if (ex is type or othertype)" style, since it's possible to do something like this: https://stackoverflow.com/a/51286385 and thus leave a generic Exception ex block for the end.  Regardless, this looks like it meets the requirement of catching and handling exceptions from SET timeouts.

PR #959, opened 2022-01-14
Close button fixes

### LINKED ISSUE #956: Make button "forum-header-dismiss" white in dark mode
Look, I've tried coding it myself, but I couldn't figure out how the dark mode works.

Current vs desired:
image image

Masterjun3:
Resolves #956.
Also fixes a leftover from bootstrap 4.

### DIFF
--- a/TASVideos/TagHelpers/BootstrapTagHelpers.cs
+++ b/TASVideos/TagHelpers/BootstrapTagHelpers.cs
@@ -223,7 +223,7 @@ $@"<button type=""button"" class=""btn-close float-end"" data-bs-dismiss=""alert
 		<div class='modal-content'>
 			<div class='modal-header'>
 				<h5 class='modal-title text-danger'>Delete Warning!</h5>
-				<button type='button' class='btn-close float-end' data-bs-dismiss='modal'><span aria-hidden='true'>&times;</span></button>
+				<button type='button' class='btn-close' data-bs-dismiss='modal'></button>
 			</div>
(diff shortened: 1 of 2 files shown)


TASVideos/TagHelpers/BootstrapTagHelpers.cs
```diff
@@ -223,7 +223,7 @@ public override async Task ProcessAsync(TagHelperContext context, TagHelperOutpu
 		<div class='modal-content'>
 			<div class='modal-header'>
 				<h5 class='modal-title text-danger'>Delete Warning!</h5>
-				<button type='button' class='btn-close float-end' data-bs-dismiss='modal'><span aria-hidden='true'>&times;</span></button>
+				<button type='button' class='btn-close' data-bs-dismiss='modal'></button>
```
ThunderAxe31, line 226: What about simply replacing it with:
`<i id="forum-header-dismiss" class="float-end fa fa-close" style="cursor: pointer"></i>`
Wouldn't be the same? I didn't make a PR because I couldn't test it myself, so I wonder it that would work.
image also, I couldn't set the margin on the left.

Masterjun3: No.
This is a bootstrap function that dismisses the modal that has been opened. Read this: https://getbootstrap.com/docs/5.1/components/modal/

The forum header dismiss sets a localstorage value so that future visits to the site don't show the header.

PR #961, opened 2022-01-14
Add colors to Announcement and Sticky topics

Masterjun3:
image

This worked out nicely for dark mode, but for light mode the colors needed to be adjusted.
It turns down the brightness of the warning yellow in light mode and uses the dark mode info color for light mode as well.
image


vadosnaprimer (approved):

vadosnaprimer:
Apparently this changed button color in a bunch of places, and with black text it's harder to read now.

изображение

изображение

PR #966, opened 2022-01-14
Show previous posts when creating a new post

### LINKED ISSUE #545: Add display of previous posts while making a new post
Example link: https://tasvideos.org/Forum/Posts/Create/9624

Adding a display of the previous posts (like the old site) is helpful to reference and reread relevant conversations.

Masterjun3:
Resolves #545.

(before | after)

https://user-images.githubusercontent.com/22375320/149519951-69c76c40-700a-49c5-8cc6-b2538364e4d9.mp4

This is a draft for now because it's probably good to add it to edits as well.
Also, the design of the posts themselves is probably a little *too* simplistic.
And because I had to pick a number, it now shows 10 posts. I don't know how many the old site showed.

### DIFF
--- a/TASVideos/Pages/Forum/Posts/Create.cshtml
+++ b/TASVideos/Pages/Forum/Posts/Create.cshtml
@@ -3,9 +3,19 @@
 @{
 	ViewData["Title"] = $"Posting to Topic: {Model.Post.TopicTitle}";
 }
-<partial Name="_ForumHeader" />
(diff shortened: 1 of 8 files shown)


Masterjun3:
I think the old site showed previous posts below the textfield, right? I'm not sure if it should stay that way or not.
To me, showing them at the top feels like the natural position, so that's what I implemented here.

adelikat (approved):

PR #977, opened 2022-01-15
Implement a localstorage cache for major textareas

### LINKED ISSUE #212: Backup typed text
If you accidentally close a tab (or the whole window) where you've been writing a super elaborate post, or a wiki edit, it's a very big loss, almost like losing your progress in a TAS. So it would be very nice if the site tracked what is entered in the most important text fields, backing it up and restoring upon demand.

TiKevin83:
fixes #212

### DIFF
--- a/TASVideos/Pages/Shared/_TextareaCacheScriptPartial.cshtml
+++ b/TASVideos/Pages/Shared/_TextareaCacheScriptPartial.cshtml
@@ -1,19 +1,20 @@
 ﻿<script>
-    const textarea = document.querySelector('textarea.wiki-edit');
-    const textareaCacheKey = textarea.getAttribute('cacheKey');
-    const revertButton = document.getElementById('revert-button');
-    if (localStorage.getItem(textareaCacheKey) !== null) {
-        textarea.value = localStorage.getItem(textareaCacheKey);
-    }
-    function cacheTextarea() {
-        localStorage.setItem(textareaCacheKey,textarea.value);
-    }
-    setInterval(cacheTextarea, 5000);
-    function clearTextareaCache() {
-        localStorage.removeItem(textareaCacheKey);
-        location.reload();
-    }
-    if (revertButton) {
-        revertButton.addEventListener('click', clearTextareaCache);
+	const textarea = document.querySelector('textarea.wiki-edit');
+	const textareaCacheKey = textarea.getAttribute('cacheKey');
+	const restoreButton = document.getElementById('restore-button');
+
+	function cacheTextarea() {
+		localStorage.setItem(textareaCacheKey,textarea.value);
+		localStorage.setItem(textareaCacheKey + "-time", Date.now())
+	}
+	setInterval(cacheTextarea, 5000);
+	function restoreCachedTextarea() {
+		textarea.value = localStorage.getItem(textareaCacheKey);
+	}
+	if (restoreButton && localStorage.getItem(textareaCacheKey) !== null) {
+		restoreButton.addEventListener('click', restoreCachedTextarea);
+		const storedTime = new Date(parseInt(localStorage.getItem(textareaCacheKey + "-time"))).toLocaleDateString();
+		restoreButton.innerText += ` from ${storedTime}`;
+		restoreButton.classList.remove("d-none");
     }
 </script>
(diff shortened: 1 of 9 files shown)


adelikat:
So this "autosaves" every 30 seconds?

TiKevin83:
yeah it caches wiki edit and submission create/edit client side while editing every 30 seconds and clears the cache if saved.

Masterjun3:
I'm not sure if this approach is good.
What if someone accidentally or intentionally clears most of textfield (without ever intending to click submit), and the cache stores this. How do they get it back? From what I can tell from the code, reloading or going back and clicking "Edit" again will overwrite it with the stored value.
What if a user edits, but discard their edit, and then a later judge added their "Judging underway"? Then the next time the user edits it would remove that "Judging underway", no?

TiKevin83:
If you feel it's necessary we could clear the cache when canceling an edit as well

Masterjun3:
The Cancel button? I wouldn't rely on people ever clicking that.

TiKevin83:
I guess we could also add keying the cache by latest revision?

Masterjun3:
What about the first issue I mentioned? How to restore the data if it always gets overridden?

TiKevin83:
For that we could add a button to the edit page "revert" that reverts the text to the most recent revision

Masterjun3:
In the first place, you would also need to notify the users that the field isn't the most recent one.
I personally think we need to go much conservative with the implementation of this feature. Don't override anything automatically, not without user promt. Make the user understand what is happening.
We've had a lot of cases of users editing and editing, and I think this feature as it is now would just increase the rate of them accidentally removing "judging underway" or similar edits.

TiKevin83:
as is the situation of one user editing repeatedly is already solved by the cache clearing on save, it just needs the keying off revision id to fix cases where someone else edits while you're editing and the cancel/revert problem, but I also agree that the browser cache would be better if fixing that header solves the case of closing a tab stopping the page from being cached.

Masterjun3:
Wait what does the Revert button actually do?

Masterjun3:
Oh, apparently it resets the cache and reloads. Uh oh. Why doesn't it just save the content when initially loading and then replaces the content and the cache when the user presses the button? By reloading, all other changes to the submission or similar are lost.

TiKevin83:
The revert button doesn't say "revert just the text" and isn't placed near it so I can imagine the alternative might be more confusing? Also yeah then you have to store the original markup somewhere in the DOM

Masterjun3:
Can't you just store the initial data in a javascript variable?

Invariel (commented):
My only question about this is: does each text area have a unique cacheKey so that I don't write half a page on Super Mario Bros., jump to the Super Metroid wiki page, and have the field pre-filled with a lot of irrelevant SMB text?

TiKevin83:
yep it's cached by page and by last update timestamp

Masterjun3:
Like I said in this comment https://github.com/TASVideos/tasvideos/issues/212#issuecomment-1014696526 I'd be fine with this approach on pages where the user *creates* something, like a post, submission, wiki page, etc. Because in this case the textfields should be empty so it's self-evident that it being filled with content means it's a saved version.
But this implementation does the same thing when *editing* something, where the textfield being filled is expected, so the user is unaware that its content was replaced for them.

If I go back to my homepage after 2 months and want to quickly edit one typo for example, but easily overlook my cancelled edits being inserted and submitted and now I have content on my page I didn't want.
The "Revert" button being present just means people would have to *always* click it to be absolutely sure nothing gets submitted they don't want.

All in all, my proposal is to *never automatically replace content if the user already expects it to be filled*.
If you want to support saves in those cases, make the code check the cache and show a button "Restore saved version from xyz time ago" or similar.

TiKevin83:
Updated per comments from masterjun

Masterjun3 (changes requested):
Currently, the Restore button becomes useless after 5 seconds. If you visit the site and don't click it within 5 seconds your cached changes are gone.

Masterjun3:
Also, a slightly more difficult problem to solve is that the javascript "ToLocaleString" is using a different way to tell locale than the server, so it shows a different formatting. This can get confusing if both dd/MM/yyyy and MM/dd/yyyy are used. Maybe it's better to use relative minutes/hours/days there, or use some other approach.

Masterjun3:
Especially annoying here is how your localstorage gets filled up without even changing the textfield, like if you're just replacing movie files or change branches of submissions. People will see a "Restore Edited Text" on every single submission even if they didn't even touch the textfield.
The caching should only start if someone actually edits the textfield.

TiKevin83:
Closing this for now, a lot of problems have been identified with various strategies for this using localstorage and it might make sense to rework it with a library of some sort to not reinvent the wheel and hit all the same problems.

PR #982, opened 2022-01-15
Disable antiforgery token for logout button

Masterjun3:
This fixes all pages having a token and thus disabling the response cache.
For logged in users, this should speed up browsing considerably.

### DIFF
--- a/TASVideos/Pages/Account/Logout.cshtml.cs
+++ b/TASVideos/Pages/Account/Logout.cshtml.cs
@@ -6,6 +6,7 @@ using TASVideos.Core.Services;
 namespace TASVideos.Pages.Account
 {
 	[Authorize]
+	[IgnoreAntiforgeryToken]
 	public class LogoutModel : BasePageModel
 	{
 		private readonly SignInManager _signInManager;
--- a/TASVideos/Pages/Shared/_LoginPartial.cshtml
+++ b/TASVideos/Pages/Shared/_LoginPartial.cshtml
@@ -12,7 +12,7 @@
 <div class="d-md-flex">
 	@if (isSignedIn)
 	{
-		<form asp-page="/Account/Logout" method="post" id="logoutForm">
+		<form asp-page="/Account/Logout" method="post" id="logoutForm" asp-antiforgery="false">
 			<navbar>
 				<nav-item activate="Messages">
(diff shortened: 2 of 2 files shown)


TiKevin83 (approved):

nattthebear:
Does this mean you can CSRF a logoout now?

adelikat:
https://owasp.org/www-community/attacks/csrf
Should we not be concerned about this for logout?

Masterjun3:
It means all a malicious third party can do is log you out. All other forms still have the token, after all.

nattthebear:
A forced logout isn't a direct vulnerability, but it's a weakness, as you can both deny service and force the user to login additional times.  Each login attempt is then something that can be attacked by other angles:  Fake login pages on a typo named site, keyloggers, etc.

If we want to stop cachebusting the logout button, we could instead have it run javascript that clears the login cookie.  This would require removing HTTPOnly from that cookie, opening up a possible XSS vector.  Perhaps that's better than a CSRF vector?  For XSS, we have to make additional mistakes to have any actual vulnerability; the HTTPOnly header is only a defense in depth.  For CSRF, we're immediately vulnerable when we stop using an antiforgery token.

adelikat:
https://stackoverflow.com/questions/17966421/is-the-anti-forgery-token-for-logout-necessary

The only consequence here is that someone could make an "annoyance" attack.  However, the alternative is that we can't utilize response caching at all, since this is on every page.  

Is this more annoying? or a server that can't handle the load of the site, that gives periodic long delays, 504 responses, and failed requests.  (Which is our current struggle, that response caching would help a lot with).

Surely there is a compromise here where something can be browser cached, without opening up a "weakness"

nattthebear:
> Surely there is a compromise here where something can be browser cached, without opening up a "weakness"

Basic forms are broken by design and always have been.  🤷   I recommend removing the XSS defense in depth instead.

Masterjun3:
Okay, you can't actually log out anymore with this change. Oops.
For some reason it still wants a token even if it's disabled, so the logout fails.

Masterjun3:
Okay, I just forgot a thing. Ignore me.

adelikat:
Merging this for now.  The need for some response caching out weighs the risk of an "annoyance" attack for now.  I'd be open to a JS solution here, if we can have a graceful fallback for noscripters (they should at least be able to log out as basic functionality that we support)

Issue #476, opened 2021-09-23
Drop Query Params for Highlighting Posts


SandTAS:
This is a minor enhancement suggestion. Currently, direct links to forum posts include a `Highlight` query parameter that causes the selected post to be highlighted:
https://tasvideos.mistflux.net/Forum/Topics/22808?Highlight=508259&CurrentPage=1#508259

It works by, for each post, checking if `Highlight` is equal to the post `Id`, and if so setting a boolean, which if true causes the CSS classes `forum-post-highlight border-warning border-2` to be applied to the post.
https://github.com/adelikat/tasvideos/blob/bbe2ce2b051f613ea1aacb68aeefd0b47706866a/TASVideos/Pages/Forum/Topics/Index.cshtml.cs#L140-L147
https://github.com/adelikat/tasvideos/blob/cdceb92343a3bddbffedd25df4e4a679d00cc81f/TASVideos/Pages/Forum/Topics/_TopicPost.cshtml#L2

It should be possible to accomplish the same effect using a selector for the [`:target-within`](https://developer.mozilla.org/en-US/docs/Web/CSS/:target-within) pseudo-class on every post, containing the same rules as `forum-post-highlight border-warning border-2`. This would remove the need for a `Highlight` query parameter in the URL. Alternatively, use [`:target`](https://developer.mozilla.org/en-US/docs/Web/CSS/:target) and set the post `id` directly on the containing `div`, rather than on an element contained within it.

adelikat:
I like this idea.  I would encourage someone interested in tinkering with the UI to try this suggestion

TiKevin83:
:target-within has no support in any released browser, so I think we need to leave this as is for now.

SandTAS:
Ah, that's my bad. I should have checked the browser support before suggesting it.

For what it's worth, when looking at the HTML I expected the `id` to be set on the `<div class="card">` element, not on an `<a>` element nested within it. In that case `:target` would work.

Feel free to close the issue—I didn't mean to drop an unactionable open issue on you :)

TiKevin83:
I do think that could be a potential solution, there's also the question of whether Chrome's link to text feature will ever catch on with other browsers, but that wouldn't exactly give the style desired here.

SandTAS:
> For what it's worth, when looking at the HTML I expected the `id` to be set on the `<div class="card">` element, not on an `<a>` element nested within it. In that case `:target` would work.

#885 does this sub-task.

[...]

PR #985, opened 2022-01-15
Simplify Post links

Masterjun3:
Currently, all Post links like
`https://tasvideos.org/Forum/Posts/511157` turn into
-> `https://tasvideos.org/Forum/Topics/23000?CurrentPage=1&Highlight=511157`
And
`https://tasvideos.org/Forum/Posts/511157#511157` turn into
-> `https://tasvideos.org/Forum/Topics/23000?CurrentPage=1&Highlight=511157#511157`
The difference is the ` #511157` (called a uri fragment).

This pull request changes the behavior so that even post links without the fragment add (or overwrite!) the fragment to the post.
I.e. posts like this
`https://tasvideos.org/Forum/Posts/511157` and this
`https://tasvideos.org/Forum/Posts/511157#511157` and even this
`https://tasvideos.org/Forum/Posts/511157#1234` all get converted into the proper
-> `https://tasvideos.org/Forum/Topics/23000?CurrentPage=1&Highlight=511157#511157`.

<hr>

This simplifies a lot of links, makes them look nicer, and makes the Discord embeds work (Discord strips all fragments away).

**But we lose one feature** with this: The ability to highlight one post but anchor something different (like a different post).
Because the fragments are never sent to the server we can't do logic depending on them. We either always replace it on posts, or never.

### DIFF
--- a/TASVideos.Core/Services/Email/EmailService.cs
+++ b/TASVideos.Core/Services/Email/EmailService.cs
@@ -80,7 +80,7 @@ namespace TASVideos.Core.Services.Email
 
 You are receiving this email because you are watching the topic, ""{template.TopicTitle}"" at {siteName}. This topic has received a reply since your last visit. You can use the following link to view the replies made, no more notifications will be sent until you visit the topic.
 
-{template.BaseUrl}/Forum/Posts/{template.PostId}#{template.PostId}
+{template.BaseUrl}/Forum/Posts/{template.PostId}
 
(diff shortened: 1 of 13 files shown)


adelikat:
I'm happy to lose the "ability" to highlight a different post than we anchor, that isn't functionality we intend to have

adelikat (approved):

TiKevin83:
fixed #476

Issue #902, opened 2022-01-10
Suggestion: Forum Markup: ignore one newline if it comes after a `[/quote]` quote end tag


Masterjun3:
Not sure how possible this suggestion is, but this is a gripe I've had with the old site as well, but the new site has elevated it even further.

Currently, a post like this

```
[quote]Hello this is the quote[/quote]
And this is immediate text under it.

And this is far off text.
```
shows up like this
![image](https://user-images.githubusercontent.com/22375320/148711299-5de04435-f303-4be9-981c-a1c78dfbecea.png)

The amount of whitespace between the quote and the following text is so big not only because the quote block has a margin, but because the quote-end already implies a newline by design, as it is a separate block. And then the newline after the `[/quote]` is parsed and another newline is added.

If one newline would be ignored, it would result in a nicer post, closer representing the source.
![image](https://user-images.githubusercontent.com/22375320/148711438-b2aad528-3a21-472b-b5aa-3c06092a3ec6.png)

nattthebear:
> but the new site has elevated it even further.

Why is that?  Different styling?

Masterjun3:
The empty space between the quote and text is even further apart now than the old site, so it's looking like the quote and the text are unrelated.

Masterjun3:
See for example this post https://tasvideos.org/Forum/Posts/509803#509803 (or other posts in that thread).
The source code of the post has properly separated quote+quotereply, like in my example post above. But the forum makes it appear perfectly in the middle of the previous and the next quote.

The idea of ignoring a newline can be extended to every markup block with an implied newline, as that creates a disconnect between source code and render.

vadosnaprimer:
The same is true for code samples and especially lists

![изображение](https://user-images.githubusercontent.com/7092625/148737134-e9e49337-14e0-461a-9d5d-7298635fc7e3.png)

nattthebear:
I can make changes to the parser, but this feels more like a styling problem.  I checked on the old site and the behavior is very similar.  The old site inserts `<br/>` where the new site uses `white-space: pre-wrap`, but the result is the same.  Given that we have this consistent behavior between the old and new parsers, why did we choose to style things in way that doesn't work well with it?

PR #991, opened 2022-01-15
Trim whitespace from block forum tags

nattthebear:
Sort of does what was asked for in #902


nattthebear:
Could use some review here; not just for bugs, but decisions since there's no one right answer here.

vadosnaprimer:
Can you post some pics please?

Masterjun3:
I tried checking this out but for some reason it only works when clicking "Preview" and doesn't work in the actual post. Is that even possible or is my setup broken?

adelikat:
I can confirm that there seems to be trimmed whitespace in the preview but not when you actually post

image

Actual post:
image

nattthebear:
I see.  The problem seems to be that when creating a post, some part of the create process is mangling `\n` into `\r\n`, but no such braindead translation is happening in post preview.  Browsers universally never send `\r` as a part of any textarea content unless the user explicitly asks for it, so I think the best thing to do here is to fix the braindead mangling in post create.  Does anyone know what is doing this?

adelikat:
All the logic is in Pages/Forums/Post/Create.cshtml.cs

Masterjun3:
According to my browser log for the POST request, it does send \r\n.

image

nattthebear:
Then why doesn't it do it on preview?  Guess I'm off to MDN to learn about some more random shit that IE4 did, so now everything does...

nattthebear:
Ready for review again.  A textarea's `.value` always contains just `\n` and not `\r`.  The preview path takes the textarea's `.value` and manually assembles it as the sole body of a `POST`, so doesn't have a `\r`.  On the other hand, the post creation path uses 'multipart/form-data' which always represents newlines as `\r\n`.  One more reason to avoid that awful format, I suppose?

Anyway, for this ticket, I made post creation trim the `\r` since we won't ever need it, and made parsing resilient to the `\r` so we don't have to migrate existing posts.

Everything should be working now.

adelikat (approved):

PR #994, opened 2022-01-16
Adjust splitcamelcase behavior

TiKevin83:
Passed the test revisions I was hoping for, fixes "P Cem" and "TAS Videos" in some wiki page titles. The code is terrifying though, I'm terrible at regex.


Masterjun3:
For anyone wondering, this changes the behavior from the old site "Split `AAa -> A Aa` and `aA -> a A`" to a new behavior where the former is removed.
This fixes things like above, but of course breaks some other things.

Link | Old site | This PR | Possible Fix
--- | --- | --- | ---
SNESAccuracyTests | SNES Accuracy Tests | SNESAccuracy Tests | Snes Accuracy Tests
Samsara/AFunnyJoke | Samsara / A Funny Joke | Samsara / AFunny Joke |
LEGOStarWarsIITheOriginalTrilogy | LEGO Star Wars II The Original Trilogy | LEGOStar Wars IIThe Original Trilogy |
MightAndMagic8DayOfTheDestroyer | Might And Magic 8 Day Of The Destroyer | Might And Magic8 Day Of The Destroyer

TiKevin83:
Yeah since I don't think we can account for both of those scenarios I think we need to have a separate editable display title for wiki pages with a fallback to one of the above problems

vadosnaprimer:
I personally prefer gluing all words together. With camel case it's still possible to understand what is separate words, and non-standard capitalization won't look broken.

Masterjun3:
For reference, to recreate this change with regex, you can replace the slash splitting and loop with
`return Regex.Replace(str, @"(\/)|(\p{Ll})(?=[\p{Lu}\p{Nd}])|(\p{Nd})(?=[\p{Lu}])|([\p{L}\p{Nd}])(?=[^\p{L}\p{Nd}])|([^\p{L}\p{Nd}])(?=[\p{L}\p{Nd}])", "$1$2$3$4$5 ");`.
This is just one big OR regex with 5 possibilites and it puts a space after it if it matches:
- `(\/)` a slash,
- `(\p{Ll})(?=[\p{Lu}\p{Nd}])` lowercase followed by uppercaseOrNumber,
- `(\p{Nd})(?=[\p{Lu}])` number followed by uppercase
- `([\p{L}\p{Nd}])(?=[^\p{L}\p{Nd}])` letterOrNumber followed by non-letterOrNumber,
- `([^\p{L}\p{Nd}])(?=[\p{L}\p{Nd}])` non-letterOrNumber followed by letterOrNumber.

This passes all the tests, but gives numbers a better representation:
Site | This PR | My Regex
--- | --- | ---
Kingdom Hearts 358 Days | Kingdom Hearts358 Days | Kingdom Hearts 358 Days
N64 | N64 | N64
Star Trek 2 5th Anniversary | Star Trek25th Anniversary | Star Trek 25th Anniversary

Not saying that regex is the best solution, it does look a bit crazy. Maybe it's even slow because of the positive lookahead.
The parsing can have many solutions, but I think numbers can be represented slight better than in this PR.

TiKevin83:
updated per above

Masterjun3 (approved):

PR #995, opened 2022-01-16
highlight terms with a background color

### LINKED ISSUE #974: Term definition misses BG color
https://tasvideos.org/TextFormattingRules#Lists

изображение

Old site

изображение

Also explanation used to be shifted from the term.

TiKevin83:
fixes #974
dtStyleDark
dtStyleLight

### DIFF
--- a/TASVideos/wwwroot/css/partials/_customizations.scss
+++ b/TASVideos/wwwroot/css/partials/_customizations.scss
@@ -374,3 +374,10 @@ tt, code, kbd, pre, samp {
 pre code {
 	padding: 0;
 }
+
+dt {
+	background: var(--cardsecondary);
+	border-radius: 5px;
+	padding: 0.125rem 0.375rem;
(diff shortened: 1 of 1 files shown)


TiKevin83:
I tried making the highlight more separate from the background and it looks pretty bad, I think it looks fine with the slight background difference

vadosnaprimer:
It's bad even with this?

изображение

Masterjun3 (approved):
I didn't even see the colors on my phone, but on my desktop monitor it looks fine. I'd personally separate them even more from the background, but if you say you tried it and it looked bad, then it's fine.
It's certainly better than what we had before.

TiKevin83:
> It's bad even with this BG?
> 
> изображение

Yeah the cardprimary color for some reason doesn't look great to me on light mode in that context
image

vadosnaprimer:
IMO it looks awesome :D

PR #996, opened 2022-01-16
give blockquote a background and italics

### LINKED ISSUE #973: Indented quote > doesn't work
Current site does nothing if the line starts with `>`

https://tasvideos.org/TextFormattingRules#Emphasis

изображение

Old site did this

изображение

This feature is much more handy than full-blown quote block that you have to remember how to type
https://tasvideos.org/TextFormattingRules#Quotations

TiKevin83:
fixes #973 
image

gives blockquote the same background as preformatted/code samples with italics to match the old site and differentiate it

### DIFF
--- a/TASVideos/wwwroot/css/partials/_customizations.scss
+++ b/TASVideos/wwwroot/css/partials/_customizations.scss
@@ -364,11 +364,15 @@ tt {
 	unicode-bidi: bidi-override;
 }
 
-tt, code, kbd, pre, samp {
+tt, code, kbd, pre, samp, blockquote {
 	background: var(--bs-gray-400);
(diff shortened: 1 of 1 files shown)


Masterjun3 (approved):

vadosnaprimer (approved):

vadosnaprimer:
Sorry I didn't realize this reduces font size. Can it be set to 100% to be consistent with forum quotes and `%%QUOTE`?

TiKevin83:
Yeah I think it's fine if you want blockquote to be 100%, the 85% is mainly to make the more inline bits have the same size as the surrounding text

PR #999, opened 2022-01-16
Add search to game list

### LINKED ISSUE #750: Games/List - Add Search
particularly for systems like NES: https://tasvideos.org/Games/List?SystemCode=NES

We need a way to search game names here.  Finding something in the middle letters of the alphabet is a drag

Masterjun3:
Resolves #750.

Not really happy with this. Technically I don't need to create an own "SearchTerms" variable in /Games/List.cshtml.cs, the query automatically takes it to the "Search" and into the "(Search.)SearchTerms". But now the URL will say "Search.SearchTerms" which breaks things, so it's possible to overwrite the "name" attribute inside the .cshtml. However, this causes the validation to not work, because it relies on the attribute.

In short we either have no automatic validation, or an extra variable in the root like this.

### DIFF
--- a/TASVideos/Pages/Games/List.cshtml.cs
+++ b/TASVideos/Pages/Games/List.cshtml.cs
@@ -1,4 +1,6 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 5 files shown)


TASVideos/Pages/Games/List.cshtml.cs
```diff
@@ -34,7 +41,12 @@ public class ListModel : BasePageModel
 
 		public async Task OnGet()
 		{
-			Games = await GetPageOfGames(Search);
+			if (ModelState.IsValid)
```
adelikat, line 44: Why are we checking the modelstate on the get method?

Masterjun3: The ModelState is invalid when the search terms of the url query aren't satisfying the conditions (which are: string must be empty or minimum of 5 characters)

adelikat (approved):

TASVideos/Pages/Games/List.cshtml.cs
adelikat: That makes sense, just an unusual pattern, but I think this is the simplest way to handle it

PR #1005, opened 2022-01-18
refactor quote markup

### LINKED ISSUE #1004: Refactor quote element usage in the forum
When someone quotes a previous comment in the forum, the quote and the author are labeled using `blockquote` and `cite`, but cite is intended for titles of works and not authors, instead `figcaption` is supposed to be used with the `figure` element wrapping the blockquote and figcaption instead of `div class="quotecontainer`.
https://getbootstrap.com/docs/5.0/content/typography/#naming-a-source
Need to reassign those elements, style the background for figure and then exclude blockquote from having the background style needed for the wiki when it's inside a figure.

TiKevin83:
fixes #1004 - uses figure and figcaption per newer HTML spec where `cite` is intended for the title of works, also integrates with bootstrap's default styling using blockquote-footer for the author and fixes the background conflict in the forum where blockquote was used for quotes.

### DIFF
--- a/TASVideos.ForumEngine/Node.cs
+++ b/TASVideos.ForumEngine/Node.cs
@@ -203,17 +203,17 @@ namespace TASVideos.ForumEngine
 					break;
 				case "quote":
 					w.OpenTag("figure");
-					w.OpenTag("blockquote");
-					await WriteChildren(w, h);
-					w.CloseTag("blockquote");
 					if (Options != "")
 					{
 						w.OpenTag("figcaption");
-						w.Attribute("class", "blockquote-footer");
 						await BbParser.Parse(Options, false, true).WriteHtml(w, h);
+						w.Text(" wrote:");
 						w.CloseTag("figcaption");
 					}
 
+					w.OpenTag("blockquote");
+					await WriteChildren(w, h);
+					w.CloseTag("blockquote");
 					w.CloseTag("figure");
 					break;
 				case "code":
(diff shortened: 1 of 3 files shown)


vadosnaprimer:
Will it affect the look of posts like this?

изображение

adelikat:
@vadosnaprimer : please list the post id of that post

Masterjun3:
https://tasvideos.org/Forum/Posts/511719

Masterjun3:
This PR definitely needs images, here is how it would look:

image

vadosnaprimer:
Yeah this style of quoting feels like it's for epic memorable notable quotable, like biblical stuff: https://tasvideos.org/GameResources/CommonTricks

If it's just a casual talk, you need to know whom are you quoting first, and we often turn author's name [into URL of some kind:](https://tasvideos.org/Forum/Posts/505652)

изображение

What I like in Masterjun's screenshot:
- Consistent BG color for quotes
- Font doesn't get smaller (and smaller (and smaller (with every nested quote)))
- Font remains non-italic which means we can apply our own highlighting to it, and it's just generally easier to read than italic.

nattthebear (commented):
The code side of this looks fine:  I have no objection to how the forum engine and the wiki engine have been modified.

I'm not involved much in the styling side so I'll let you all duke it out, but two things in mind:

1. There may be some interaction between this and https://github.com/TASVideos/tasvideos/pull/991 in terms of overall spacing
2. Please thoroughly regression test this against all forum and wiki markup.  Regardless of how beautiful it did or didn't look, the styling we had in this codebase one year ago was functionally more complete than what we ultimately went to production with:  there have been a number of regressions with things being styled very poorly.

TiKevin83:
reverted to author above quote block
image
image

vadosnaprimer (approved):

PR #1016, opened 2022-01-19
Don't call `string.Split` w/ a single-char string

YoshiRulz:
JavaScript habits die hard I guess


TiKevin83:
oh do the single quotes let you use a different overload that splits on a char in C# ?

PR #1036, opened 2022-01-22
Round timespans to nearest hundredth for display

### LINKED ISSUE #1025: Timing difference between module:frames and Submissions
From #1013 ,  which is "Timing difference between submission and UserFile"

It happens with different systems than Genesis for example NES as well
https://tasvideos.org/3595M - 34:52.44 - [module:frames|amount=125753|fps=60.0988138974405] 34:52.43 - **0.01 difference**
https://tasvideos.org/7296S - 1:47:01.42 384789 - [module:frames|amount=384789|fps=59.922751013550524] - 1:47:01.41 - **0.01 difference**
https://tasvideos.org/7290S - 09:32.65 34315 - [module:frames|amount=34315|fps=59.922751013550524] - 9:32.65 - **no difference**
https://tasvideos.org/7283S - 04:54.51 17648 - [module:frames|amount=17648|fps=59.922751013550524] - 4:54.51 **same**

nattthebear:
We already had code for this, we just weren't calling it on this codepath.

Fixes #1025


Masterjun3:
Why not make the frames module a ITimeable and then use the Time() that already has this logic and accounts for other edge cases?
https://github.com/TASVideos/tasvideos/blob/f6d934624d42fb56fa245d030635cf56e70cba7a/TASVideos.Common/ITimeable.cs#L19-L33

nattthebear:
Didn't remember that piece of code at the time.  Seems like we have a lot of ways to do things; all of the FramesModule should be blown away and replaced by ITimeable.

nattthebear:
Does anyone know why `TimeableExtensions.ToStringWithOptionalDaysAndHours` always includes at least minutes but `TimeExtensions.ToCondensedString` goes down to a seconds display?  I want to blow away the latter, but they have that difference in behavior @adelikat

Masterjun3:
Not including minutes when unnecessary probably makes sense for its usage as an inline element, wouldn't want to say a trick saves a time of 00:00.67 and better just say 0.67.

adelikat:
@nattthebear : I honestly don't know any more.  I made some things, people didn't like it and made some things.  I'm okay with doing one consistent thing and way for people to yell

nattthebear:
@adelikat Thanks for the feedback :+1:  I'll make the change, and if any other minor issues come out, we'll find and fix them.

nattthebear:
OK, it's using timeable now and the numbers seem to make sense

adelikat (approved):

PR #1041, opened 2022-01-23
Make the next button only show if needed

### LINKED ISSUE #1031: Search has Next button even on single-page results
For search terms with few results, like "[Anodyne](https://tasvideos.org/Search?SearchTerms=Anodyne)", there is still a Next button even when [there's nothing on page 2](https://tasvideos.org/Search?searchterms=Anodyne&pagenumber=2). The Next button should not appear unless it's needed for [multi-page search results](https://tasvideos.org/Search?searchterms=Kirby%2064).

TiKevin83:
fixes #1031 
I tried to implement this in a way that minimizes excess querying of the database


adelikat (approved):
Clever way to handle this

PR #1042, opened 2022-01-23
Add Permission To Login

adelikat:
Checks for this permission on sign-in, if not present, redirects to AccessDenied.  One thought I have, is we could redirect to a wiki page that explains the situation . I have no thoughts as to the name of this page, but if someone has one and wants this, I can add that in.

As part of this deployment, Default User and Site Admin will need this permission added.
Since this checks the role_permissions table, it will see all eligible users as having this permissions, so there should not be any login complications for existing users.


TiKevin83:
looks good to me as long as there's a way to add the login permission when no one will have it at first when the code is launched.

TiKevin83 (approved):

PR #1047, opened 2022-01-24
Net6.0

### LINKED ISSUE #883: Update to .NET 6
We need to do this, putting in a 2.4 milestone to signify that it needs to be done as soon as things have settled down

TiKevin83:
fixes #883 
Upgrades package references for .net 6 and a random recaptcha library update. CSS wasn't loading at first but the weboptimizer library just needed an upgrade for compatibility and everything seems to work fine otherwise.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,3 @@
+{
+	"recommendations": ["ms-dotnettools.csharp"]
+}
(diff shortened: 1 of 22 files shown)


nattthebear:
Could file-scoped namespaces be moved to a separate PR after we're on net 6?  It's very noisy and makes it difficult to track what actually changed here

adelikat:
Yes, let's just upgrade, and then do the C#10-isms in separate commits later

PR #1068, opened 2022-01-31
Ability to remove someone's claim to a submission

adelikat:
image

Adds a new permission: RemoveQueueClaim
If a user has it, and they view a submission that is Judging Underway to Publication Underway, they will see a Remove Claim button that will reset the status, remove the judge or publish, and add a status message.

image


vadosnaprimer:
When is this needed?

adelikat:
Wow, I totally missed that permissions, it is OverrideSubmissionStatus, so the RemoveQueueClaim permission is redundant.

@vadosnaprimer makes a good point too that it is too rare to warrant a handy button.

Given these circumstances, this PR is not needed

PR #1090, opened 2022-02-09
Movie module - Show RomHash Region, if not null

adelikat:
Conditionally shows the region, if the rom hash entry is added
The value needs to not be in the title so that it can be styled independently, so I put it all the way to the left.  I also only show it if it is relevant, which currently are these values:
J => (JPN),
E => (Europe),
EU => (USA/Europe)

We can easily configure these values if want different examples.

image

image

### DIFF
--- a/TASVideos/Extensions/EntityExtensions.cs
+++ b/TASVideos/Extensions/EntityExtensions.cs
@@ -181,7 +181,8 @@ public static class EntityExtensions
 				OverallRating = p.PublicationRatings
 					.Where(pr => !pr.Publication!.Authors.Select(a => a.UserId).Contains(pr.UserId))
 					.Where(pr => pr.User!.UseRatings)
-					.Average(pr => pr.Value)
+					.Average(pr => pr.Value),
+				Region = p.Rom != null ? p.Rom.Region : null
 			});
 
 		if (ratingSort)
--- a/TASVideos/Extensions/EntityExtensions.cs
+++ b/TASVideos/Extensions/EntityExtensions.cs
@@ -182,7 +182,8 @@ public static class EntityExtensions
 					.Where(pr => !pr.Publication!.Authors.Select(a => a.UserId).Contains(pr.UserId))
 					.Where(pr => pr.User!.UseRatings)
 					.Average(pr => pr.Value),
-				Region = p.Rom != null ? p.Rom.Region : null
+				Region = p.Rom != null ? p.Rom.Region : null,
+				RomVersion = p.Rom != null ? p.Rom.Version : null
 			});
 
 		if (ratingSort)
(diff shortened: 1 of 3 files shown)


meshuggahtas:
I think (U) USA should be added too

vadosnaprimer:
I think `JPN` should be turned into `Japan`. Alpha-3 codes are used for countries, but we also have `Europe` and `Asia` which don't have official alpha-3 codes. So instead of inventing codes for those, I think it would be neater to spell out Japan instead.

Also IMO there should be `World` and it should be default and hidden, just like `any` was on the old site. Finally, there should probably also be `unknown`.

Need opinions from @fsvgm777 and @ThunderAxe31

adelikat:
I dont' know what you mean about hidden.  The RomHash table has all those things, we are just deciding which to show, and what display value to show.

So it sounds like the only change you want is JPN to Japan?

vadosnaprimer:
Hidden I mean not displayed in publication titles.

Also in the case of [VVVVVV](https://tasvideos.org/Forum/Posts/511442) and possibly other PC games, we need to display numeric game version instead of the region, since region is kinda pointless on PC.

adelikat:
image

Masterjun3 (changes requested):

vadosnaprimer:
For cases when the same game works in all regions, we need to have `World`. `USA/Europe` doesn't include Japan and PAL Asia. PAL Asia games are rare, but if somebody submits a NES game made there and it uses Dendy mode, that's when we used the `Asia` region, to indicate it's running in hybrid mode. This is not critical tho.

On the matter of display, I think it's ok to put region and version within the same pair of parentheses.

ThunderAxe31:
I agree that `JPN` should be `Japan`, and that `any` should be `World`.

I don't like the style of `(Japan) (v1.4)`, as in my opinion it makes it unclear that both things are referring to the game version. I suggest instead using `(Japan v1.4)`.

Also, I want to note that there are more combo cases, like Pokémon Mystery Dungeon, in for which a version is known as `USA/Australia`.

adelikat:
JPN has been changed to Japan already.
and neither any nor world show.

fsvgm777:
Okay, kinda late to the party, but I agree with hiding any/World, especially for PC games (DOS, Windows, Linux), where region doesn't really matter.

> Also, I want to note that there are more combo cases, like Pokémon Mystery Dungeon, in for which a version is known as USA/Australia.

I think that can be put as Other or just USA (if GBA)? Same deal with games released exclusively in Germany (Hugo 2 1/2) or France (no movie examples, but I know of some (e.g. Les Visiteurs)).

> I don't like the style of (Japan) (v1.4), as in my opinion it makes it unclear that both things are referring to the game version. I suggest instead using (Japan v1.4).

Agreed.

adelikat:
image

vadosnaprimer (approved):

PR #1092, opened 2022-02-09
Restrict judgement edit permissions

TiKevin83:
Leaving draft as I haven't figured out how to test the judgement process locally

### DIFF
--- a/TASVideos/Pages/Submissions/Edit.cshtml
+++ b/TASVideos/Pages/Submissions/Edit.cshtml
@@ -14,19 +14,19 @@
 		<div class="col-lg-6">
 			<form-group>
 				<label asp-for="Submission.Status" class="form-control-label"></label>
-				<select class="form-control"
+				<select disabled="@(!User.Has(PermissionTo.JudgeSubmissions))" class="form-control"
 						asp-for="Submission.Status"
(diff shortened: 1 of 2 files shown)


TiKevin83:
Right now, submission authors can modify the rejection reason on a submission and the status. There seem to be better solutions to this with the queue service but I wanted to fix the immediate concern of that being possible

Masterjun3:
> Right now, submission authors can modify the rejection reason on a submission and the status because of needing to set to cancelled.

Well, the current situation is not quite as bad as you make it out here. If a movie is rejected a user can't change its status at all. In fact, if it's not rejected a user can't do a lot except cancelling it.
It's all handeled in the QueueService so most of the changes of this PR currently just doubles the logic that was already working quite well.

The issue that came up of course was that users could change the rejection reason (and maybe the intended tier?), so if it's intended for a quick fix you can just prevent it server-side by surrounding the lines where they're set with a judge permission condition.

adelikat:
This just isn't the correct approach to this problem.  There is a method in queue service that is in charge of determining what statuses are allow for what permissions and it is well unit tested, and deals with a lot of complicated scenarios.  The work needs to be there

PR #1099, opened 2022-02-12
When a user successfully resets their password, mark their email confirmed

adelikat:
I think this makes sense.  If they got this far, they got an email and clicked a link so we know it is verified.

Maybe this can't happen in a normal workflow but with legacy users, they may have unconfirmed emails and no password from the password wipe.


meshuggahtas:
I think this change is OK.

Is there anything that could negatively affect legacy users / Are these scenarios will work when clicking password reset?
- Legacy user with no password
- Legacy user with "upgraded password storage"

Masterjun3 (approved):

PR #1100, opened 2022-02-13
Add Topic Activity Indicator

### LINKED ISSUE #709: "New Post" Indicator for Boards and Topics
Old feature from phpBB: When there was a new post made in a board since you last visited, the icon for the board in the forum overview, and the icon for the topic in the topic list would light up.

I found this an immensely useful feature for keeping track of activity and would like to have some kind of notification icon for new posts again. 

This would probably also necessitate a "Mark All Unread As Read" button somewhere.

Masterjun3:
Resolves #709.

This readds a highly requested feature of showing an indicator when a topic has posts the viewer hasn't read yet. It uses the viewer's local storage to keep track of which topics the user visited and when. This prevents the viewer from seeing activity indicators for new posts they already read.

It currently tracks 14 days of activity before a topic stops being marked as having activity. This could be easily changed to track a longer period. The code should easily scale upwards. But increasing the timespan does mean a bit more processing for the server and client. (But probably still not as much as all the javascript libraries we're including on each page.)

image
image


Masterjun3:
Converting to draft because topic moves are not handled.

TASVideos.Core/Services/ForumService.cs
```diff
@@ -97,11 +107,44 @@ public void CacheLatestPost(int forumId, int topicId, LatestPost post)
...
+		if (_cacheService.TryGetValue(TopicActivityJsonCacheKey, out Dictionary<int, string> jsonForumActivity))
+		{
+			var dictSubforumActivity = new Dictionary<int, string>();
+			jsonForumActivity.TryGetValue(forumId, out string? jsonSubforumActivity);
+			if (jsonSubforumActivity != null)
+			{
+				dictSubforumActivity = JsonSerializer.Deserialize<Dictionary<int, string>>(jsonSubforumActivity) ?? new Dictionary<int, string>();
+			}
+			dictSubforumActivity[topicId] = CreateTimestamp.UnixTimestamp().ToString();
+			jsonForumActivity[forumId] = JsonSerializer.Serialize(dictSubforumActivity);
```
adelikat, line 132: Is there any reason you are serializing before caching? You get that for free from Redis Cache, and for memory cache it can just be the native class

Masterjun3: I'm not serializing for the cache and later deserializing. I *need the json itself* to embed the data into html and send it javascript.

adelikat (approved):

PR #1104, opened 2022-02-13
When serving userfiles, don't uncompress+recompress.

nattthebear:


adelikat (commented):
I feel like this is moving code into the wrong place.  IActionResult is a presentation concern that the core layer shouldn't understand.  Maybe just a private method on the PageModel is good enough

nattthebear:
> I feel like this is moving code into the wrong place. IActionResult is a presentation concern that the core layer shouldn't understand. Maybe just a private method on the PageModel is good enough

Sure, I'll put it on the pagemodel

PR #1108, opened 2022-02-19
Add DSi support with Bizhawk 2.8

Fortranm:
It seems Bizhawk 2.8, which adds support for DSi, is about to be released. This commit aims to make the site able to parse DSi movie files as a separate platform.

Q: Why a separate platform?
A: DSi is compatible with all DS games, some DS games have enhancements while running on DSi, and there are DSi exclusive applications that can't be run on DS. The situation here is directly comparable to that of GB and GBC.

Q: Why not parsing DSiWare separately as well?
A: DSiWare is a line of software products, not a platform. It's not like the site lists WiiWare separately. (The first game to get a published movie on the site, Mega Man 10, is a WiiWare title, btw.)


adelikat (approved):
Approving conditionally.  The site needs to support this system first.  But the code is a good approach

adelikat (approved):

PR #1114, opened 2022-02-22
Add CTM Movie File Parser

benstephens56:
This pull request is to add a parser for Citra movie files. I had a substantial amount of trouble testing locally, but I believe this is functioning. Code is based on my findings on the header for CTM files: https://pastebin.com/8w3ned8i

CTM Header

Is there anything else that I need to do here to do with this? I'm new at this and probably missed something.


TASVideos.Parsers/Parsers/Ctm.cs
```diff
@@ -0,0 +1,37 @@
...
+			return new ErrorResult("Invalid file format, does not seem to be a .ctm");
+		}
+
+		br.ReadUInt64(); // Title ID
+		br.ReadBytes(20); // Git hash of Citra revision
+		br.ReadUInt64(); // Init time of system clock
+		br.ReadUInt64(); // Movie ID
+		br.ReadChars(32); // Author
+		result.RerecordCount = br.ReadInt32();
+		result.Frames = (int)Math.Round(br.ReadInt64() * InputCountConvert);
```
nattthebear, line 33: What's the significance of this constant here?  If the system is truly sampling at 234hz and marking frames at that, then the frame count should just be `br.ReadInt64()` and the framerate should be 234.

CasualPokePlayer: Citra doesn't have a frame count, only an input count, and polls inputs at 234 Hz, so this is to convert it to the 3DS' framerate.

Although this should be Ceiling rather than a round following the precedent of other parsers (see https://github.com/TASVideos/tasvideos/blob/edafc0228488caa7185fe07d48fad09a32e217f8/TASVideos.Parsers/Parsers/Dtm.cs#L73)

TASVideos.Parsers/SystemCodes.cs
```diff
@@ -23,6 +23,7 @@ internal class SystemCodes
 	public const string Lynx = "lynx";
 	public const string Msx = "msx";
 	public const string Ngb = "ngb";
+	public const string N3ds = "3ds";
```
CasualPokePlayer, line 26: N3ds is before Ngb alphabetically, not after.

CasualPokePlayer (changes requested):

TASVideos.Parsers/Parsers/Ctm.cs
benstephens56: > What's the significance of this constant here? If the system is truly sampling at 234hz and marking frames at that, then the frame count should just be `br.ReadInt64()` and the framerate should be 234.

Great question. So it's true that Citra does sample inputs 234 times a second, and that is how the final time of the movie file is determined. Sampling rate, however, does not directly affect frame count. For instance the system can poll for inputs either 3 or 4 times every frame. On top of that, Citra framerate is not exactly 60 fps, my testing tells me it's at 59.831246111702 fps, so 234 cannot be used to directly convert input count to frames.


> Citra doesn't have a frame count in the movie file, only an input count, and polls inputs at 234 Hz, so this is to convert it to the 3DS' framerate.
> 


Citra does not have a frame count currently, but I'm in the process of getting the method I used here merged into the main Citra repo currently. 

> Although this should be Ceiling rather than a Round following the precedent of other parsers: see
> 
> https://github.com/TASVideos/tasvideos/blob/edafc0228488caa7185fe07d48fad09a32e217f8/TASVideos.Parsers/Parsers/Dtm.cs#L73

Using ceiling or floor would almost never cause a problem. Unfortunately, due to the slight inconsistencies with how many times Citra samples inputs during any given frame, rounding up or down exclusively will cause the frame count to not tick up or jump up by two  every 800 frames or so, but even that interval is inconsistent. Perhaps someone smarter than me could figure our a relationship that would work with this rounding method, but I was not able to do it. What I can tell you is that I am certain my method will not cause any duplicate or skipped frames counts between adjacent frames based on my testing with about 25 different games of varying complexity running movie files that were approximately 10 hours long.

I can go into more details, on how I went about coming up with this, but it's a little long winded, honestly.

CasualPokePlayer: That issue with rounding isn't really exclusive to Citra really. The main reason for rounding is you in practical effect have a subframe time, which if that is being converted for the nominal framerate, we are going to assume the subframe is a full frame and you aren't going to be saving time due to rounding down.

benstephens56: > That issue with rounding isn't really exclusive to Citra really. The main reason for rounding is you in practical effect have a subframe time, which if that is being converted for the nominal framerate, we are going to assume the subframe is a full frame and you aren't going to be saving time due to rounding down.

That's fair, it doesn't make all that much a difference to me either way, so if that's the precedent here, I'm fine with it. I will go ahead and change my constant to incorporate the cycle rate and frame rate to make things less ambiguous, as well as change my alphabetical error (oops).

CasualPokePlayer: https://github.com/citra-emu/citra/blob/a2f34ea82b5a31a7e842d0099921b85b8bce403f/src/core/hw/gpu.h#L27
On a note about framerate.

nattthebear: > Sampling rate, however, does not directly affect frame count.

"Frame count" here should mean input frames, not video frames.  That's how all of the other parsers define it.  So the framerate is 234fps and the framecount is just `br.ReadInt64()` with no conversion.

benstephens56: Updated everything to match the the dtm file parser more closely. Is there anything else you think should be changed at this point?

CasualPokePlayer: In most cases input frames and video frames mean the same thing. But not always, some emulators may allow changing input multiple times in a frame.

https://github.com/TASVideos/tasvideos/blob/edafc0228488caa7185fe07d48fad09a32e217f8/TASVideos.Parsers/Parsers/Omr.cs#L60
https://github.com/TASVideos/tasvideos/blob/edafc0228488caa7185fe07d48fad09a32e217f8/TASVideos.Parsers/Parsers/Dtm.cs#L73

Parsers don't particular agree here on this matter, OMR and DTM parsers calc the time of the movie and make a frame count from that.

Not only that, some parsers just take a frame count directly, which may or may not represent input frames (most probably do, tbf)
https://github.com/TASVideos/tasvideos/blob/edafc0228488caa7185fe07d48fad09a32e217f8/TASVideos.Parsers/Parsers/Ltm.cs#L50
https://github.com/TASVideos/tasvideos/blob/edafc0228488caa7185fe07d48fad09a32e217f8/TASVideos.Parsers/Parsers/Mar.cs#L34
https://github.com/TASVideos/tasvideos/blob/edafc0228488caa7185fe07d48fad09a32e217f8/TASVideos.Parsers/Parsers/Fbm.cs#L46

BK2s tho seem to do a different approach to cycle count cores, where it will instead just use input count like the non cycle count cores, when create a managled framerate from that to represent the input frame and thus the correct time
https://github.com/TASVideos/tasvideos/blob/edafc0228488caa7185fe07d48fad09a32e217f8/TASVideos.Parsers/Parsers/Bk2.cs#L126-L130

The "make frame count match video frames" approach seems to be more prevalent to deal with mismatches, but just slapping a framerate override with 234 probably would work too? Don't particularly care here anyways, whatever works works.

CasualPokePlayer: When I brought up the CycleCount field in Dolphin that was just an example on how to deal with getting a frame count out from a calculated time. The input count is not (usually) a cycle count, if anything frame count * 4481136 is the cycle count (going off Citra code), but this shouldn't be stored here.

benstephens56: So are you thinking I just shouldn't include cycle count at all? That's fine with me, just wanted to keep stuff consistent.

benstephens56: > When I brought up the CycleCount field in Dolphin that was just an example on how to deal with getting a frame count out from a calculated time. The input count is not (usually) a cycle count, if anything frame count * 4481136 is the cycle count (going off Citra code), but this shouldn't be stored here. You can just name the constant InputRate and calc it as Inputs * InputRate / FrameRate = Frames (which is essentially Seconds / FrameRate = Frames, the same as DTMs do).
> 
> Granted at this point you could just slap 234 to the framerate override and put the input count into the frame rate field, probably would make the code simpler anyways.

Understood, I simplified that bit and changed the const name from CycleRate to InputRate.

TASVideos.Parsers/Parsers/Ctm.cs
```diff
@@ -0,0 +1,38 @@
+using TASVideos.MovieParsers.Result;
+
+namespace TASVideos.MovieParsers.Parsers;
+
+[FileExtension("ctm")]
+internal class Ctm : ParserBase, IParser
+{
+	private const decimal FrameRate = 59.83122493939037M;
```
CasualPokePlayer, line 8: Perhaps instead do `268111856.0M / 4481136.0M` and link to https://github.com/citra-emu/citra/blob/a2f34ea82b5a31a7e842d0099921b85b8bce403f/src/core/hw/gpu.h#L27 in a comment.

CasualPokePlayer: And don't use a url shortener please...

CasualPokePlayer (dismissed):
Seems good enough imo. The site will need to add support for the 3DS system code (cc @adelikat), but otherwise this is fine.

CasualPokePlayer:
Actually retracting this approval, some parser unit tests should be included in here too (I can go whip some up if you don't feel like doing so).

benstephens56:
Honestly, it would be a huge help if you could do that! Most of this stuff is not exactly in my wheelhouse as you have probably gathered by this point. I really do appreciate your patience with me up until this point.

CasualPokePlayer (approved):

CasualPokePlayer:
After using Citra, I'm realizing it's "frame advance" simply advances a video frame. So really it's just the movie file format being kinda weird by reporting a input count instead of just having some frame count. Oh well, might as well do this.

benstephens56:
Yeah it's definitely not the best system ever, but anyone with the knowledge and know-how to rework it has lost interest unfortunately. What's there works well enough, but it definitely requires working around it a little, currently.

adelikat:
Will merge this soon, probably this weekend.  This will need to be deployed with some kind of announcement, so I want to be ready and available

adelikat:
Ok, someone has time now, so merging

PR #1121, opened 2022-02-23
Don't round userfile times to the nearest second

### LINKED ISSUE #652: New GameCube/Wii submissions only timed to full second
To reproduce: Submit a dtm

### LINKED ISSUE #1013: Genesis movie gives different timing between submission and userfile
Yes, it's the very same movie file.

https://tasvideos.org/7283S
image

https://tasvideos.org/userfiles/info/637781221746390722
image

CasualPokePlayer:
fix #652
fix #1013


Masterjun3:
Does this also affect #1013?

CasualPokePlayer:
Ah, that should be fixed too.

adelikat (approved):

Issue #1030, opened 2022-01-21
Rom hashes can't be reused and SGB platform redundancy


coolkirby2000:
TASVideos' [Rom list](https://tasvideos.org/Games/1473/Roms/List) feature is ​useful for authors of submissions to ensure they're using a verified good ROM. These lists will become more important in the future [when they're reflected on publication pages](https://github.com/TASVideos/tasvideos/issues/906#issuecomment-1008992699).

This may be by design, but the site prohibits adding a second ROM with the same MD5 or SHA1 hash, either to the same Game or a different one.

This is only an issue with GB and SGB Game entries for the same game which use the same ROM, like [GB Donkey Kong](https://tasvideos.org/2241G) and its [SGB copy](https://tasvideos.org/1026G). [Here is a full list](https://tasvideos.org/Forum/Topics/22854) of these pairs of Games. As console verification (via GBP/GBI) requires that GB games be played in GBA mode and parsed as GB, the GB version of a game is preferred for submission. So hashes for its Roms will likely be filled first.

I see three options going forward:
1. Change nothing. SGB submissions of Donkey Kong, etc. will not be able to choose a Rom if the GB game already uses it. On the other hand, all Rom lists will be safe from typos allowing a hash to be accidentally added twice.
2. Allow hashes to be reused. This will allow GB and SGB entries to use the same Rom, but is a workaround that's only useful for this one platform redundancy and removes the protection against typos.
3. Merge SGB Game entries/movies into the GB entries. This will let a GB game in any form use the one entry for each Rom hash, and avoid a blank Rom field, which would cause issues for submissions and publications. It also standardizes the nomenclature for these games, as GB, GB-in-GBA and GB-in-SGB are all just the GB ROM anyway. If we do this, we would probably need to relabel all SGB games as GB to be consistent, removing this redundant platform from the site.

What are everyone's thoughts on this issue?

adelikat:
The unique constraint on GameVersion (formerly known as Rom) is gone so I'm closing this ticket.  As for how to handle this situation, it is a matter of staff decisions, but whatever needs to be done can be now

Issue #332, opened 2021-08-28
Wrong Framerate used for BizHawk's SameBoy Core


CasualPokePlayer:
The SameBoy core in BizHawk is set to run at 2147727/35112 FPS, but it seems the site parser just uses the SNES framerate (which is correct for LLE cores like BSNES as inputs are locked to SNES VBlank like with a real SGB, but the HLE cores like SameBoy have no concept of this FPS and only use the internal framerate of the SGB).

adelikat:
Our site policy is to not care about what a core or emulator does, and focus on what it _should_ do.  We don't want to be dependent on the details of what a core does, or will do in the future.  Our fps values are based on theoretical hardware, and not specific emulator behavior.

So our SGB value is based on a LLE value, and I believe it is correct to do so.

nattthebear:
Bizhawk Sameboy SGB and Bizhawk Bsnes SGB run at the exact same speed (excepting the boot up difference, and whatever lag differences there are).  The difference here is really a frame timing thing.  A real SNES runs at 60.09 FPS fixed framerate.  A real SGB inside the SNES runs at a 61.17 FPS nominal framerate; it will be exactly that framerate except when the virtual LCD "display" is turned off and on.

The SGB hardware handles the difference by dropping frames once in a while when outputting to the SNES side.  I honestly don't know what real SGB hardware does to handle the input timing difference; maybe it polls on vblank for a steady 60.09 input samples per second and then just duplicates some of them to the SGB side?  Or maybe it has irregular polls and an emulator that included the SNES side would need subframe input to recover it all.  In any event, it does something to reconcile the two.

Bizhawk Sameboy SGB is counting the 61.17 fps frames as its notion of actual elapsed time; Bizhawk Bsnes SGB is counting the 60.09 fps frames as its notion of actual elapsed time.  They're counting time from different measurements, and "one frame" is not equal to "one frame" because they're one frame of different things.

There are similar problems with various notions of timing for GB and A2600, among others; "equal length frames" vs otherwise.

I don't have an opinion on what TASVideos does, but I think we should have all the facts.  Reopening this for now so we can make sure all the needed discussion happens.

CasualPokePlayer:
[...]

PR #1125, opened 2022-02-25
parse SGB movie files to use the GB label

Fortranm:
Related to #1030.
It doesn't make much sense to use SGB as a separate system on the site when GB movies done on GBC mode are still labeled as GB (the same applies to GBC movies on GBA, GBA movies on GBP, etc).
The use of SGB as a separate system is even causing issues like [this movie](https://tasvideos.org/1840M) being unintentionally un-obsoleted.
This commit aims to make the parsing system label future SGB submissions as GB. SGB is still kept as a platform on the site and in edge cases, a movie can be manually changed to be under that as a system.


CasualPokePlayer:
I personally don't particularly care with the policy here, other staff can discuss that. My only concern to it would be if someone decided to TAS that Space Invaders game that uploads SNES code and has that executed or some ACE playaround (e.g. Pokemon Plays Twitch) doing that. Someone should figure out how to resolve that.

That aside, there is a slight issue with this. Namely, framerates. This is a bit of a complicated issue, namely due to the multiple ways "framerate" can be defined.

The internal GameBoy speed on the SGB1 is derived from the SNES master clock. This is lazily divided by 5 to get the clock rate, which after that dividing by 70224 will give the internal nominal framerate, ~61.167 FPS on NTSC (21.477Mhz / 5 / 70224), and ~60.61 FPS on PAL (21.28137Mhz / 5 / 70224). On the SGB2, a separate clock is used for the internal GameBoy speed, making the internal speed and nominal framerate identical as the GameBoy regardless of region (~59.7275 FPS)

However, these internal nominal framerates do not mean so much in LLE cores (i.e. lsnes, both old and new BizHawk BSNES) and on the actual system, which would have actual frames displayed and inputs sent at the SNES' framerate (~60.099 FPS for NTSC, ~50.007 FPS for PAL), dropping/duping GB frames to maintain that framerate. As inputs are sent at this rate, these framerates would used for LLE cores (which depending on region, with SGB 1 vs 2 playing no choice here).

For HLE cores however (Gambatte in BizHawk), the SNES framerate would not be used, rather the internal nominal GB framerate, as the actual SNES is not emulated as the commands sent to the SNES are HLE'd.

VBA is also an HLE core, although that's 60 FPS and even then you still have problems with its timing and then you have the detail that SGB is banned for VBA (although GB/C VBA is likely going to be put on that ban list soon, so eh).

And then even so, Gambatte still gives a variable framerate due to the fact the GB framerate is a nominal framerate rather than an absolute framerate, so cycle count would need to be used. As Gambatte only supports the SGB2 and as established the SGB2 internal timing is identical to the GB ends up working out (so as is it just works).

The old SGB SameBoy core in BizHawk (removed in 2.7) (see #332) should be noted to run at the nominal SGB1 NTSC framerate instead, and it is forced to run with Equal Length Frames, so cycle count isn't needed there but just that framerate.

So tl;dr framerate is wrong here for LLE SGB emus.

Fortranm:
Added frame rate override settings for bSNES based SGB movie files.
As a note, it seems there is currently no way to record SGB movies in PAL mode on BizHawk as of now, but that scenario is still taken into consideration for future-proofing.

adelikat:
This was discussed briefly in discord.  But I the idea was that we do default parsing to GB. Then in the rare instances where SGB makes sense (such as a TAS that actually utilizes SGB/SNES hardware, or a game that utilizes its capabilities) we can just manually change it to SGB.  These would be rare enough that the manual work is fine

vadosnaprimer:
> It doesn't make much sense to use SGB as a separate system on the site when GB movies done on GBC mode are still labeled as GB (the same applies to GBC movies on GBA, GBA movies on GBP, etc).

Is SGB just a mode or are there games that actively utilize its features? Are there SGB-only games?

I agree that if someone makes an SGB only movie, removing SGB entirely solves zero problems but adds new ones like having to classify it wrong.

ThunderAxe31:
> This was discussed briefly in discord. But I the idea was that we do default parsing to GB. Then in the rare instances where SGB makes sense (such as a TAS that actually utilizes SGB/SNES hardware, or a game that utilizes its capabilities) we can just manually change it to SGB. These would be rare enough that the manual work is fine

Please be more specific. Which ones of these features do you think would be worth of using the SGB system name?

- Animated borders ([video](https://www.youtube.com/watch?v=6ArYj4-Ajs4))
- Wav sounds ([video](https://www.youtube.com/watch?v=ksHirSih4XU))
- Multiplayer mode (there are [31 titles](https://en.wikipedia.org/wiki/List_of_Super_Game_Boy_games) featuring it)
- Space Invaders, which is the only SGB game that actually plays as a full-screen SNES game ([video](https://www.youtube.com/watch?v=QKsJokBb9bw))

Fortranm:
> Is SGB just a mode or are there games that actively utilize its features? Are there SGB-only games?
Quoting Wikipedia: "The Super Game Boy is a peripheral that allows Game Boy cartridges to be played on a Super Nintendo Entertainment System console. The Super [Game Boy](https://en.wikipedia.org/wiki/Game_Boy) is compatible with the same cartridges as the original Game Boy: original Game Boy cartridges, the [Game Boy Camera](https://en.wikipedia.org/wiki/Game_Boy_Camera), and dual-mode [Game Boy Color](https://en.wikipedia.org/wiki/Game_Boy_Color) cartridges (in Game Boy-mode)."
It ultimately is an adapter that allows GB games to be played on an SNES. There is no SGB-only game. IMO this is similar to playing GB games on GBC, GB/C games on GBA, and GB/C/A games on GBP.

> I agree that if someone makes an SGB only movie, removing SGB entirely solves zero problems but adds new ones like having to classify it wrong.
The latest commit in this branch keeps SGB as a platform on the site and only makes SGB submissions parsed as GB by default.

> Which ones of these features do you think would be worth of using the SGB system name?
Yeah, it is important to draw an actual line for "SNES features", because if we take it in the literal sense, even simple display of colors beyond monochrome can be considered "SNES/GBC/GBA/GBP feature".

> Space Invaders, which is the only SGB game that actually plays as a full-screen SNES game
This is often used as an example on this topic, but it seems the SNES mode from that game is identical to the SNES version of the game (https://tcrf.net/Space_Invaders_(SNES)) so I doubt there is much practical reason for someone to make a submission of the SNES mode of that game unless it's about ACE demonstration, just saying.

CasualPokePlayer (approved):

vadosnaprimer:
> The latest commit in this branch keeps SGB as a platform on the site and only makes SGB submissions parsed as GB by default.

There should be some rewording of what this PR ends up doing, because people read it, see that it removes SGB as a platform, and get confused.

PR #1126, opened 2022-02-25
Improve diff system

### LINKED ISSUE #987: Highlight inline (word) difference for wiki edits
The old site used to show inline diff only, which was very helpful when checking what was edited and making sure the page can serve as official documentation:

изображение

New site shows it like the entire paragraph was changed, not just a single word:

https://tasvideos.org/Wiki/PageHistory?path=HomePages%2Ffeos

изображение

So when verifying the new wording it may be extremely difficult to figure out what was actually changed. If the paragraph is huge, it gets even harder. You have to read both old and new paragraphs word by word to find what was changed, and then keep them both in mind in full to determine if the change is good. Bonus points if an edit introduced an important spelling mistake that you couldn't spot.

Both "Inline" and "Side by side" view options have this problem.

nattthebear:
Replace diff library with a smarter one, and create a simple view layer for it.  Closes #987.


TiKevin83:
I don't see any handling of dark mode, also the css colors should come from imports from color variables. I'm glad to help with those as needed

nattthebear:
@TiKevin83 Thanks, that would be great.  I've done some rudimentary styling, but feedback and tweaks are appreciated.

nattthebear:
A note on the colors:  The old dark mode background colors all failed WCAG, so I darkened them to about 10:1 each.  The old light mode colors passed WCAG, but I found the red much harder to read than the green.  It was at a lower contrast level so I lightened it up until it had the same ratio as the green.  The default monospace font for my browser is rather faint.

TiKevin83:
I like this a lot but I would pull up the lightness of the dark mode background diff colors a tad to give them more differentiation from the grey background, #125E38 for green and #AA132A for red are just over the 7:1 AAA cutoff but I'd be fine with closer to 8.5:1 too, the 10:1 is just too close to the grey IMO.

TiKevin83 (approved):

nattthebear:
OK, I'll tweak the dark mode colors a bit tomorrow and then this will be ready to merge.

nattthebear:
Yes, I think those 7:1 colors will do fine.

adelikat (approved):

PR #1127, opened 2022-02-25
Userfiles: Content-Length is always physical length

Masterjun3:
Fixes the userfile downloads and Content-Length mismatches in the log.


nattthebear:
This is correct (assuming that file.PhysicalLength always exactly matches the length of the stored byte array).  My understanding of the Content-Length header was incorrect; when transfer encoding is used it should contain the compressed size.

nattthebear (approved):

PR #1151, opened 2022-03-13
Add context size to diff viewer

nattthebear:
I'm not too sold on the UI here so if anyone wants to tweak it, go ahead


vadosnaprimer (approved):
Checked locally, looks great!

PR #1174, opened 2022-03-19
Revert "Revert "parse SGB movie files to use the GB label (#1125)" (#…

Fortranm:
This reverts commit 06d0050bab96629e70edca2654c6e9c2c1680a50.
Related to https://github.com/TASVideos/tasvideos/issues/1030.

There was recently an attempt to implement code changes on the site that would parse all future SGB submissions to have the label GB instead. It was merged at one point but reverted because a formal consensus on the matter hasn't been reached.

Regarding this change itself, there are two questions that need to be answered IMO:
1. Are all currently published SGB movies better label as GB instead of SGB?
The exact line before GB and SGB labels can be up to debate later but all of those VBA ones literally couldn't have access to SNES functionality.

2. Is it still necessary to actively discourage people from doing GB movies in SGB mode even if it’s on the latest cores on Bizhawk?
The initial reason from a decade ago, if I understood correctly, was that VBA was trash on this regard and BizHawk didn't have up-to-standard SGB emulation either. The latter is no longer true. IMO, there is no reason to deliberately avoid using SGB mode unless the palette on GBC looks much better or the movie is specifically aiming for console verification.

If your answer to 1 is Yes and to 2 is No, then that commit most likely makes things better based on your stance.
Also keep in mind that even if we decide to automatically parse everything SGB related as SGB, there will still need to be site code changes because BizHawk bSNES submissions, the ones that actually can utilize SNES hardware in some way, would currently be parsed as SNES.


adelikat:
@MemoryTAS : can we get a verdict here? Are we accepting or rejecting this change?

vadosnaprimer:
> Are all currently published SGB movies better label as GB instead of SGB?
> The exact line before GB and SGB labels can be up to debate later but all of those VBA ones literally couldn't have access to SNES functionality.

With other meta-GB modes we don't specify the mode as a platform, even if it's emulated to run inside GBA or GameCube, we only specify the game type. Making this mode an exception may need some good reasoning.

I think we can say a game is targeting SGB explicitly if it doesn't run (or isn't playable) on regular GB. If it runs fine without the SNES, changing the platform to SGB if **the movie** uses the SNES may be desired by some people, but I'm not sure. Looks like we need a site-wide poll about this, to avoid mindreading.

> Is it still necessary to actively discourage people from doing GB movies in SGB mode even if it’s on the latest cores on Bizhawk?

No.

vadosnaprimer:
I think [the community](https://tasvideos.org/Forum/Topics/22854) is in agreement here

изображение

Fortranm:
> Tho is this made irrelevant after the [platform approach revamp](https://tasvideos.org/Forum/Posts/514397)?
"Also keep in mind that even if we decide to automatically parse everything SGB related as SGB, there will still need to be site code changes because BizHawk bSNES submissions, the ones that actually can utilize SNES hardware in some way, would currently be parsed as SNES."
This PR itself is more about changing the labeling for submissions, which isn't affected by the change in question at the time being from what I understand.

vadosnaprimer (approved):
Masterjun told me this PR doesn't conflict with the new game system, and we have a community agreement in the thread, so I guess it's safe to merge. @MemoryTAS?

MemoryTAS:
Seems fine I guess, idk if it will remain relevant in the future tho

Masterjun3:
After the Game System Revamp mentioned above the obvious justification for this change is gone. The discussion topic focuses on publications and game pages, which this PR doesn't affect.
After a lot of discussion it became apparent that there are new things to be decided related to the topic of Systems, what to do with them, and how to display them.
It's a policy change instead of a bugfix or enhancement. The question of whether to merge the PR or not would only complicate matters that are already complicated.

So I'm closing this.

However, please make a new PR with the SNES bugfix in here. It's a separate thing and ideally shouldn't have been part of the PR in the first place.

PR #1179, opened 2022-03-20
Compress userfiles on upload

nattthebear:
Avoids problems with postbody size limits in certain cases


TASVideos/Pages/UserFiles/Upload.cshtml.cs
```diff
@@ -86,6 +86,16 @@ public async Task<IActionResult> OnPost()
 			return Page();
 		}
 
+		byte[] actualFileData;
+
+		{
```
adelikat, line 91: These curlies seem unnecessary, also stylecop nags

nattthebear: `tempStream` and `gzip` only have meaning within this block; the local scope keeps them from confusing the rest of the method.  I don't care about stylecop; any formatting nag that doesn't just automatically do the formatting for you is a complete waste of developer time.

adelikat (approved):

PR #1180, opened 2022-03-21
Only load prism.js on pages needing it

### LINKED ISSUE #1130: Script prism.js hurts page load times significantly
Currently, prism.js is loaded on every page. It takes up a significant portion of the load time.
Blocking prism.js in my adblocker reduces loading time for forum page from 260ms to 190ms.

Example on /Forum page.
image

Masterjun3:
This uses the selector used in prism.js itself to check whether loading it is needed.
Resolves #1130.

Before | After:
image

### DIFF
--- a/TASVideos/Pages/Shared/_Layout.cshtml
+++ b/TASVideos/Pages/Shared/_Layout.cshtml
@@ -127,7 +127,13 @@
 			integrity="sha384-ka7Sk0Gln4gmtz2MlQnikT1wXgYsOg+OMhuP+IlRH9sENBO0LRn5q+8nbTov4+1p"
 			crossorigin="anonymous"></script>
 	<script src="~/js/site.js"></script>
-	<script src="~/js/prism.js"></script>
+	<script>
+		if (document.querySelector('code[class*="language-"], [class*="language-"] code, code[class*="lang-"], [class*="lang-"] code') !== null){
+			let prism = document.createElement('script')
+			prism.src = '/js/prism.js'
+			document.body.appendChild(prism)
+		}
+	</script>
 	@await RenderSectionAsync("Scripts", required: false)
 </body>
 </html>


nattthebear:
How does this work on previewing new posts/edits?

Masterjun3:
Uh oh, it doesn't!

adelikat:
Preview pages are all the same partial, maybe it can be loaded there no matter what, instead of an if check.  Or an additional if check for a div with a class or some other indicator that expresses the intent to need this stuff

nattthebear:
Easy now, there are over 200 issues on this tracker.  We're doing what we can.

Masterjun3:
Nah I just meant, having a "code doesn't color on previews" is a much better issue to have than "100ms of almost every page load is wasted".
For me this is a much more critical issue than everyone thinks.

nattthebear:
So this PR will both break preview screens and not even fully fix the linked issue:  as the library is still very slow to load, and on pages that use it we'll still be loading it slowly.

Masterjun3:
Well, I wasn't the one who renamed the issue title. This PR actually fixes the original title and issue I had.

I'm fine with this PR being closed if it's deemed not necessary/sufficient/too many side effects/etc. Just know that my opinion is that solving critical issues quickly with lesser issues as side effects is overall better than taking a longer time to resolve a critical issue fully and perfectly.

nattthebear:
I renamed the issue because the slowness is the real problem.  We have lots of boilerplate that loads on every page even when not needed, but the rest of it isn't an issue because it doesn't take 100ms+ to execute.

nattthebear:
Leaving all of that aside, whether this PR is acceptable as-is mostly boils down to a product question:  Do we need syntax highlighting on code blocks in forum and wiki previews?  I don't have a strong opinion there.

nattthebear:
> How big a difference does this PR make now that #1183 is merged?

We can make some measurements against prod once that PR is deployed.

Masterjun3:
Yeah I was also waiting for deployment to test it. This PR is probably not necessary anymore, but actually seeing the other merged PR in effect would be nicer than a "probably".

adelikat:
Ok, I deployed

Masterjun3:
Okay, on every page prism.js now executes in like 3ms, and on pages with language code blocks, it takes like another 3ms per language loaded. Old durations were always 100ms. (Not counting the time actually highlighting, which is like 0.5-3ms per field, but that's not on us.)
Compared to bootstrap's 9ms, this PR is unnecessary now. The speedup is negligible, and accounting for all possible dynamic site changes is too much effort for the reward here.

In fact, the fact that this PR doesn't work on previews is now a bigger issue than the minimal slowdown, so I'm closing this PR.

PR #1183, opened 2022-03-22
Use prism.js autoloader

### LINKED ISSUE #1130: Script prism.js hurts page load times significantly

nattthebear:
As documented on their site.  Only a few core recommended languages are built into the build now, and the rest are loaded dynamically as needed.  Seems to work fine on both static content and dynamic preview panes.  I did not update to the latest Prism CSS, as ours is heavily customized so replacing would be non-trivial, but there seem to be no style issues.

Fixes #1130 to some extent.
Independent of #1180; in theory, these two can cooperate.


nattthebear:
Here's a full page load on a small simple page from my local dev env; the box highlighted in red is `prism.js`.

image

Masterjun3 (approved):

adelikat:
Thems a lot of files!  I like this approach though, thanks for doing this

adelikat (approved):

PR #1195, opened 2022-03-24
make navbar search a text box instead of a button, fixes #721

### LINKED ISSUE #721: Search bar on the top right instead of a Search page
It would be better if there were a search bar on the top right instead of a link to a search page. There can still be a link to a separate search page but only for advanced searching.

adelikat:


TASVideos/Pages/Shared/_LoginPartial.cshtml
```diff
@@ -29,7 +29,12 @@
 				<button id="Logout" type="submit" class="btn btn-link navbar-btn nav-link text-nowrap fs-6 border-0"><i class="fa fa-sign-out"></i> Log out</button>
 			</nav-item>
 			<nav-item activate="Search">
-				<a id="Search" class="nav-link text-nowrap" asp-page="/Search/Index"><i class="fa fa-search"></i> Search</a>
+				<form method="GET" action="/Search/Index">
+					<div class="input-group mt-1">
+						<button type="submit"><i class="fa fa-search"></i></button>
```
Masterjun3, line 34: Maybe this and the other button could use a better style, like:
`<button type="submit" class="btn btn-sm border border-secondary nav-link">`
It would fit better with the other design. It would turn this:
image
into this
image

Masterjun3 (changes requested):

Masterjun3 (approved):

PR #1216, opened 2022-04-04
Client side validation of rating text input values

### LINKED ISSUE #1182: Inline Rating input value
(edited description)
Adding more digits after the tenths part of selectable values will be temporarily shown to the page, should be identical to saved vote.

Repro:
1. Click Rate
2. Press up once to get 0.1 or 0,1 according to your locale settings
3. Insert extra numbers after 1 for example 987654
4. Shows Rated 0.1987654 / 10 but in reality it's saved as 0.2 or 0,2

Leaving the page will show the correct saved rating value.

TiKevin83:
fixes #1182

### DIFF
--- a/TASVideos/Pages/Shared/_MovieModule.cshtml
+++ b/TASVideos/Pages/Shared/_MovieModule.cshtml
@@ -172,11 +172,17 @@
 					}
 					function ratingConnect(textbox, slider) {
 						textbox.oninput = function () {
-							slider.value = textbox.value;
+							let oldValue = parseFloat(this.value);
+							let newValue = Math.min(Math.max(oldValue.toFixed(1), 0), 10);
+							if (oldValue !== parseFloat(newValue)) {
+								textbox.value = newValue;
+							}
+							slider.value = newValue;
 						};
 						slider.oninput = function () {
-							slider.value = Math.round(Number(this.value) * 2) / 2;
-							textbox.value = slider.value;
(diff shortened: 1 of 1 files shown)


Masterjun3:
This shouldn't be done in Javascript.
We have a whole class for dealing with validation, and this is doubling code.

TiKevin83:
it should be done in both places, otherwise as noted in the original ticket the granularity of ratings isn't communicated to the user. This is because browsers do not consistently enforce the "step" attribute which requires this handling in Javascript

Masterjun3:
Nah, the issue is just that I returned the wrong value to be updated from the server, that's a quick fix.
But this PR here is the wrong approach.

adelikat:
IT isn't necessarily the "wrong" approach to have both client and server side validation.  Client side validation avoid the request to the server and saves precious resources.  It often means duplicated effort, however.  Then it is a judgement call if the effort is worth the savings. I typically like anything that can save server resources, but I would guess that rating validation errors are rare.  Behaving users are mostly sliding a slider. Not a lot that can go wrong there

Masterjun3:
This doesn't even save resources. The server already handles conversion from say 4.58 -> 4.6, because it has to due to non-javascript users on the rating page.
This reduces user control (possibly replacing their input on every single keystroke), and makes code maintenance harder as now rating limits are in several places.

adelikat:
Oh, I misunderstoodk I thought it was doing some validation to prevent it from going to the server just to get kicked back.

PR #1217, opened 2022-04-04
Avoid need for redis locally by disabling the twitter distributor

TiKevin83:


adelikat (changes requested):
Instead of checking debug mode, check if twitter settings are passed in, and only register if they are

TiKevin83:
Updated per review

adelikat (approved):

PR #1236, opened 2022-04-13
Set email confirmed to true if user changes email

adelikat:
This is better than allowing a user to lock themselves out.  And to be fair, isn't any worse than the old forum code


adelikat:
Did this a different way

PR #1241, opened 2022-04-17
Change Email Workflow

adelikat:
Allows user to change email in profile, sends a verification email, email is not changed until successful email confirmation.

It persists the code in cache, which limits the lifetime of it to 1 hour, but I think the token itself only lasts that long.  Won't survive a reboot on memory cache, but it's unlikely that the user will be doing this during a reboot and can always try again.

### DIFF
--- a/TASVideos.Core/Services/UserManager.cs
+++ b/TASVideos.Core/Services/UserManager.cs
@@ -476,4 +476,19 @@ public class UserManager : UserManager<User>
 			})
 			.ToListAsync();
 	}
+
+	/// <summary>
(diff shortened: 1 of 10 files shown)


Masterjun3:
Isn't it kind of bad there is no verification the token belongs to the user who opened the link? From what I can see of the code, it only checks whether *a user* opened it, not whether *the user* opened the link.
(The verification and change could be done with the `ChangeEmailAsync(user, newEmail, token)` function we never use.)

adelikat:
Yeah, it could require the user be authorized and we could do that

Masterjun3 (approved):

PR #1248, opened 2022-04-23
Rough in for viewing deleted pages, fixes #735

### LINKED ISSUE #735: Deleted pages should be viewable
https://tasvideos.org/Wiki/DeletedPages

I can undelete them but I can't view their contents. Kinda makes sense if there's an option to undelete: one should be able to know what they're undeleting.

Also I feel they should be sorted alphabetically instead of reverse that.

adelikat:
Allows one to see the details of any deleted page or specific revision.

Currently no link to specific revision, just on /Wiki/DeletedPages, any page that is "fully" deleted will have a link to the latest deleted revision.  Note that it show the markup not the rendering.  I'm thinking safety here, the rendered page may likely be inappropriate, but also possibly with dangerous links or malicious markup.  Plus, you only need a rough idea of what was on the page.

### DIFF
--- a/TASVideos/Pages/Wiki/DeletedPages.cshtml
+++ b/TASVideos/Pages/Wiki/DeletedPages.cshtml
@@ -15,7 +15,10 @@
 		@foreach (var item in Model.DeletedPages.OrderByDescending(m => m.PageName))
 		{
 			<tr>
-				<td>@item.PageName</td>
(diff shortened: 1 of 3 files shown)


adelikat:
image

vadosnaprimer (approved):

PR #1256, opened 2022-04-26
Implement advanced search with regex allowing for exact phrase searches

### LINKED ISSUE #1197: Search: Support for exact phrase searches
Our current search system works with tsvector's and tsquery's. They take the English language into account, dropping frequently used words and changing plural into singular, and dropping most punctuation.
However, occasionally users want to search for exact phrases and this feature is a limitation.
An "exact search" feature would be useful here.

Masterjun3:
Might resolve a bunch of issues, but I don't exactly know which ones.
Resolves #1197.

### DIFF
--- a/TASVideos/Pages/Search/Advanced.cshtml.cs
+++ b/TASVideos/Pages/Search/Advanced.cshtml.cs
@@ -1,0 +1,106 @@
+﻿using System.ComponentModel.DataAnnotations;
+using System.Text.RegularExpressions;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.EntityFrameworkCore;
+using TASVideos.Data;
+using TASVideos.Data.Entity;
+using TASVideos.Data.Entity.Forum;
+
+namespace TASVideos.Pages.Search;
+
+[AllowAnonymous]
+public class AdvancedModel : PageModel
+{
+	public const int PageSize = 10;
+	private readonly ApplicationDbContext _db;
+
+	public AdvancedModel(ApplicationDbContext db)
+	{
+		_db = db;
+	}
+
+	[FromQuery]
+	[StringLength(100, MinimumLength = 2)]
+	[Display(Name = "Search Terms")]
+	public string SearchTerms { get; set; } = "";
+
+	[FromQuery]
+	public int PageNumber { get; set; } = 1;
+
+	public List<PageSearchModel> PageResults { get; set; } = new();
+	public List<PostSearchModel> PostResults { get; set; } = new();
+	public List<GameSearchModel> GameResults { get; set; } = new();
+
+	public async Task<IActionResult> OnGet()
+	{
+		if (!_db.Database.IsNpgsql())
+		{
+			ModelState.AddModelError("", "This feature is not currently available.");
+			return BadRequest(ModelState);
+		}
+
+		if (!ModelState.IsValid)
+		{
+			return Page();
+		}
+
+		// there is no dedicated method to check whether a regex string is valid, so we use try catch
+		try
+		{
+			Regex.Match("", SearchTerms);
+		}
+		catch (ArgumentException)
+		{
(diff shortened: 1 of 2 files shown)


vadosnaprimer (approved):

Masterjun3:
Ah, and someone else needs to look at the database calls and confirm it's safe like this.
There is user input directly in contact with a database query, so we need to be sure nothing dangerous can happen.

TASVideos/Pages/Search/Advanced.cshtml.cs
```diff
@@ -0,0 +1,163 @@
...
+					t.Id,
+					t.Forum!.Name))
+				.ToListAsync();
+			}
+
+			if (PostSearch)
+			{
+				PostResults = await _db.ForumPosts
+				.ExcludeRestricted(seeRestricted)
+				.Where(p => Regex.IsMatch(p.Text, "(^|[^A-Za-z])" + SearchTerms))
```
adelikat, line 127: Does this generate sql?? If not, this brings back a half a million posts into memory and processes from there.  We MUST know this becomes a sql query, or else we can't do this

TASVideos/Pages/Search/Advanced.cshtml.cs
```diff
@@ -0,0 +1,163 @@
...
+	[Display(Name = "Search Games")]
+	public bool GameSearch { get; set; } = true;
+
+	public int DisplayPageSize { get; set; } = PageSize;
+	public bool EnablePrev { get; set; } = false;
+	public bool EnableNext { get; set; } = false;
+
+	public async Task<IActionResult> OnGet()
+	{
+		if (!_db.Database.IsNpgsql())
```
adelikat, line 60: What postgres specific functions are you utilizing that need this check?

TASVideos/Pages/Search/Advanced.cshtml.cs
Masterjun3: From my tests it does. Here is an example from the log:
```
[20:05:58 INF] Executed DbCommand (6ms) [Parameters=[@__p_0='?', @__p_2='?' (DbType = Int32), @__p_1='?' (DbType = Int32)], CommandType='Text', CommandTimeout='30']
SELECT t.id, g0.code, t.display_name
FROM (
    SELECT g.id, g.display_name, g.system_id, (
        SELECT COUNT(*)::INT
        FROM publications AS p
        WHERE g.id = p.game_id) AS c, (
        SELECT COUNT(*)::INT
        FROM submissions AS s
        WHERE g.id = s.game_id) AS c0
    FROM games AS g
    WHERE g.display_name ~ ('(?p)' || @__p_0)
    ORDER BY (
        SELECT COUNT(*)::INT
        FROM publications AS p
        WHERE g.id = p.game_id) DESC, (
        SELECT COUNT(*)::INT
        FROM submissions AS s
        WHERE g.id = s.game_id) DESC, g.display_name
    LIMIT @__p_2 OFFSET @__p_1
) AS t
INNER JOIN game_systems AS g0 ON t.system_id = g0.id
ORDER BY t.c DESC, t.c0 DESC, t.display_name
```

The important part is the `WHERE g.display_name ~ ('(?p)' || @__p_0)` since the `~` is the regex operator in postgres.

TASVideos/Pages/Search/Advanced.cshtml.cs
Masterjun3: It was copied from the other Search. Since there was no explanation why it was put there I left it there.
I don't know if this is needed.

TASVideos/Pages/Search/Advanced.cshtml.cs
adelikat: WOW. I'm impressed with EF.

TASVideos/Pages/Search/Advanced.cshtml.cs
Masterjun3: Okay, I removed the check.

adelikat (approved):

vadosnaprimer:
I already approved the thing it's doing, I can't assess the tech part tho.

vadosnaprimer (approved):
**Approved harder!!!**

Issue #1188, opened 2022-03-23
Game Entities Revamp Suggestion


Masterjun3:
There are some ( #855, #944, #981, #1030, #1152 ) issues which can be difficult to solve in our current system of storing information about games, rom versions, and publications.
A big problem comes from the isolated design of our Game system, where e.g. Windows VVVVVV and Linux VVVVVV don't have a connection.

<hr>

<details>
<summary>Original Post 2022-03-23</summary>

After some discussions, here is my suggested solution: Redesign the Game Entities.
### Current System
Our current Game Entity is the abstract idea of a game on one platform, and this game is then split into different types of ROM versions. These can have different regions and revision numbers, and all the other information we need to make a movie sync.

And this system might have been fine in the past, but the days are gone where a game can exist only on one platform. Nowadays game versions on different platforms are so similar, it can make sense for one to obsolete the other. And if that's the case, an obsoletion-based design makes more sense. 

### New System
The new Entity will be the abstract idea of a game identified purely by a name. It's then split up into Game Versions, which can be on different platforms, have different types, and have all the information needed for movie sync.
This design of the Entity is based around the fact that different game versions can obsolete each other. And this is what we group together.
We align ourselves to what speedrun.com is doing and create this new grouping of games so that games of various types can be understood as the same game. We create a one-to-one correspondence between game versions in a collection and those who obsolete each other. In short:
*If game versions can obsolete each other, they are grouped together.*

### Suggested Changes
This gif shows my idea. The steps are further described below.
<img src="https://user-images.githubusercontent.com/22375320/159775436-af068126-546b-4f1f-ac8b-c623251e71da.gif" width="600">

<hr>

#### Part 1 - The current system
![TASGameRevamp-Page-1 drawio](https://user-images.githubusercontent.com/22375320/159775578-87b16538-9541-4723-a201-6a683f3d8516.png)

#### Part 2 - Make the new game grouping
![TASGameRevamp-Page-2 drawio](https://user-images.githubusercontent.com/22375320/159775760-37ea7d22-1a61-4afb-bd02-28148e4c800f.png)
Rename Game to GameVersion, and make a new Game Entity that groups together GameVersions. Technically this can be done by making a new Game for *every single* GameVersion (and later combine them), but because we already have chains, we can instead use our current obsoletion chains.
E.g. Pokemon Red, Green, Blue, Yellow GameVersions all get a single Pokemon RGBY Game.
[...]

PR #1276, opened 2022-05-10
Replace every instance of Game.System and rework the logic to use GameRom.System

Masterjun3:
**Important**: Before this is merged we need to assure in the database that:
- Every Game has at least one GameRom
- Every GameRom has a System.

<hr>

Part of the Game Entity Revamp #1188.
This is the biggest step and removes every single instance of the column we plan to remove. It reworks all instances to either remove its usage or replace it with the GameRom.System. It also adds UI to make it possible to add Systems to GameRoms.

Every single commit is self-contained and results in a buildable project and working site.


Masterjun3:
This is ready. It can be merged. Then once we can assure every GameRom in the database is non-null we can deploy it.

adelikat (approved):

PR #1300, opened 2022-05-26
allow Game version hashes (md5, sha) to be nullable

adelikat:


adelikat:
Note that after this deploys, I'll write a script to set all the empty values to null

Masterjun3 (approved):

Masterjun3:
What about the EditModel? https://github.com/TASVideos/tasvideos/blob/a46dcf7bd55d5c3b9cae6f3e28276c6c106e9f62/TASVideos/Pages/Games/Roms/Models/RomEditModel.cs#L18-L28

PR #1311, opened 2022-05-31
Remove Wiki/Diff and support functionality in Wiki/PageHistory

adelikat:
This PR refactors Wiki/PageHistory to support an initial state that is url based.  Then it is linked to instead of /Diff and that page is removed.

Having specific revision histories being bookmarkable is a nice feature as well.  The slight downside is now as you click prev or from/to, it does a page load.


vadosnaprimer:
Does it still have the ability to open the actual diff in a new page, so you don't have to go back and forth while searching when something was changed?

adelikat:
I'm not sure what you mean, but this does not change it.  It just replaces the "latest diff" page with the page history page

Masterjun3:
With this it might be worth it to adjust robots.txt?
Maybe replace this line
https://github.com/TASVideos/tasvideos/blob/f38ca40a3891781f2036d991827d165face1d2b0/TASVideos/Middleware/RobotHandlingMiddleware.cs#L34
with `Disallow: /Wiki/PageHistory`

PR #1315, opened 2022-05-31
Show id, date, and class in publication activity

vadosnaprimer:
no idea if I used `!` corrctly

изображение


adelikat:
Yes, that's a place to use the ! Symbol. It's weird because it can never be null but it is really hard for a compiler to know that

vadosnaprimer:
If it can never be null then why is it nullable?

PR #1322, opened 2022-06-06
Remove Automapper

adelikat:
Auto-mapper is useful in that it abstract away mundane code.  But there is a tradeoff, code complexity.  It also hides things like nullability mismatches, properties that aren't matched, and other complexities when properties are not simple name mappings.  More importantly, contributors are struggling to understand how to use it properly.  Just as easy is to just put the mundane code in or hide it with an extension method.  Then there is no question what is happening, nullability mismatches can be compile warnings, and it is safer for refactors.  Case in point, the API was broken with the RomId -> GameVersionId refactor.  Automapper was the cause of this and hid the fact that this was happening.

All in all, this might be a worthwhile library for enterprise grade situations, but I think for a hobby project it does more harm than good.

Note that a number of subtle bugs were apparent and fixed as part of removing auto-mapper


nattthebear:
I agree with this change, although I don't have time to look at all of the individual files to hit every last detail.

PR #1345, opened 2022-06-18
forum: show tooltip for our custom URL tags if there's replacement text

vadosnaprimer:
example:

`[movie=4689]pub[/movie]` is rendered as `pub` linking to that movie and showing `[4689] GB Gunman Clive by ViGadeomes in 00:47.17` on hover

also added space between [id] and pub title in wiki link

### DIFF
--- a/TASVideos.ForumEngine/Node.cs
+++ b/TASVideos.ForumEngine/Node.cs
@@ -136,6 +136,7 @@ public class Element : INode
 		w.Attribute("href", href);
 		if (Options != "")
 		{
+			w.Attribute("title", await transformUrlText(Options));
 			await WriteChildren(w, h);
 		}
 		else
--- a/TASVideos/ViewComponents/WikiLink.cs
+++ b/TASVideos/ViewComponents/WikiLink.cs
@@ -46,7 +46,7 @@ public class WikiLink : ViewComponent
 			var title = await GetPublicationTitle(id.Value);
 			if (!string.IsNullOrWhiteSpace(title))
 			{
-				model.DisplayText = $"[{id.Value}]" + title;
+				model.DisplayText = $"[{id.Value}] " + title;
 			}
 		}
 


nattthebear:
I don't understand the point of this.  UAs already tell you where links go to, so why is an extra tooltip needed?

image

vadosnaprimer:
It's mostly for internal links like submissions, publications, games, etc. You can't know what movie or game it links to from the ID which is the only thing shown in the URL.

nattthebear:
> It's mostly for internal links like submissions, publications, games, etc.

As implemented, it shows up on all link types as I indicated in my screenshot.  Was that intended?

vadosnaprimer:
All I'm seeing in https://tasvideos.org/Forum/Posts/509042 is submission ID

изображение

nattthebear:
Yes, yes, I understand how it works on submission links.  But what about all of the other link types, why did they all get changed?

vadosnaprimer:
Ah. Didn't consider it harmful to show the tooltip on _all_ links, so didn't even think to explicitly limiting this to only internal ones.

TASVideos.ForumEngine/Node.cs
```diff
@@ -136,6 +136,7 @@ private async Task WriteHref(HtmlWriter w, IWriterHelper h, Func<string, string>
 		w.Attribute("href", href);
 		if (Options != "")
```
vadosnaprimer, line 137: @adelikat should I do

```suggestion
		if (Options != "" && Name != "url")
```

?

nattthebear: @vadosnaprimer What about something like https://github.com/TASVideos/tasvideos/tree/transformurltext?

This captures the idea that we need the tooltip whenever transformUrlText is non-trivial.

vadosnaprimer: Yep that's better, adelikat approved too. Please PR.

PR #1347, opened 2022-06-23
Forum:  Show tooltip on urls when appropriate

nattthebear:
Example:

`[movie=4689]pub[/movie]` is rendered as `pub` linking to that movie and showing `[4689] GB Gunman Clive by ViGadeomes in 00:47.17` on hover.

In the implementation, we observe that `transformUrlText` should be shown somewhere so long as it's a non-trivial value.

### DIFF
--- a/TASVideos.ForumEngine/Node.cs
+++ b/TASVideos.ForumEngine/Node.cs
@@ -129,20 +129,24 @@ public class Element : INode
 		}
 	}
 
-	private async Task WriteHref(HtmlWriter w, IWriterHelper h, Func<string, string> transformUrl, Func<string, Task<string>> transformUrlText)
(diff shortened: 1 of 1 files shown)


vadosnaprimer (approved):

vadosnaprimer:
I also tried to do the same for internal links on the wiki site, but it ended up being way too advanced for me (the old site had it on both forum and wiki).

nattthebear:
>I also tried to do the same for internal links on the wiki site, but it ended up being way too advanced for me (the old site had it on both forum and wiki).

@vadosnaprimer Could you write that up as a separate issue?  I agree that it's more difficult to do in the code, so I can't just bang it out right now, but I'd like to get to it eventually.

Issue #326, opened 2021-08-20
Only shrink move screenshots if they are above 320x240


vadosnaprimer:
For systems that use bigger framebuffer, the publisher would most likely shrink them manually, but for smaller resolution it doesn't seem to be saving any valuable space compared to 1x.

Spikestuff:
Four issue with this.
PAL resolutions are 384x288.
16:9 ratio titles (and the other arbitrary ones that exceed this; see arcade screenshots)
DS and its less used 256x384 for screenshots.
And Multi-Game TASes.

This should be relied on the Publisher to do it right, not some automative process.

vadosnaprimer:
I'd be fine with disable soft-shrinking entirely.

adelikat:
We don't shrink anything.  We just put img tags in html and stuff happens.
What I need to know is if this is specific to the new site vs the old site, so I can prioritize it appropriately.  Note that I don't really understand the problem or care that much, so less is more when describing it

vadosnaprimer:
Somehow it does the opposite now. Pub screenshots are bigger on demo.

![image](https://i.imgur.com/bV4EtwW.png)


Old size displays them pixel-perfect.

![image](https://i.imgur.com/AMZcTAD.png)

![image](https://i.imgur.com/nBKynno.png)

Masterjun3:
Since we want to support a range of pixel densities, screenshots can't simply be put to 100% pixel width and that's it. They need to be scaled up if they appear on a high density display. And sometimes they even need to be scaled down on small low density displays.
Currently, the column they're in resizes to at least the width needed for a 1x scale of the image (unless the surrounding container itself isn't large enough, then it's downscaled). If anything else in the column resizes it even further the image is scaled up to fill the empty space.
Missing from the screenshots you've shown is the long text `Rating: (Too few votes to display rating) (2 votes)` that enlarged the column thus enlarging the images.

vadosnaprimer:
I agree that it makes sense to resize it depending on pixel density in some cases, but first of all, non-integer scaling should use some pixel art friendly method, because my first screenshot looks really bad.

Also, I don't think scaling screenshots of identical resolution differently, depending on things around them, makes sense. If it's viewed on a high density monitor, it doesn't matter what happens around it, it's just hard to see, so all such screenshots need to be upscaled.

Finally, low pixel density doesn't necessarily mean it's a small display. For example, I can afford seeing pixel-perfect screenshots on my 1440x900.

[...]

PR #1352, opened 2022-07-03
Movie module - show "Watch on Youtube" or "Watch on Nicovideo" instead of just "Watch"

adelikat:
Before:
image

After:
image

If display name already set:
image

Nicovideo example:
image

### DIFF
--- a/TASVideos/Pages/Shared/_MovieModule.cshtml
+++ b/TASVideos/Pages/Shared/_MovieModule.cshtml
@@ -124,7 +124,7 @@
 					@foreach (var url in Model.OnlineWatchingUrls)
 					{
 						<a href="@url.Url" class="btn btn-primary btn-sm mt-1" target="_blank">
-							<i class="fa fa-external-link"></i> Watch @(!string.IsNullOrWhiteSpace(url.DisplayName) ? $"{url.DisplayName}" : "") @StreamingSiteDisplay(url.Url)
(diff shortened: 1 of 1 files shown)


Masterjun3:
If this change is wanted, the logic of the frontend needs to be adjusted (or it needs to be a high priority issue).
Currently our layout heavily relies on the button section to be not too wide, because otherwise the screenshot will be scaled up too much. #326.

TASVideos/Pages/Shared/_MovieModule.cshtml
```diff
@@ -55,6 +55,27 @@
...
+	{
+		if (string.IsNullOrWhiteSpace(onlineWatchingUrl))
+		{
+			return "";
+		}
+
+		var lowered = onlineWatchingUrl.ToLower();
+		if (lowered.Contains("youtube"))
+		{
+			return " on Youtube";
```
vadosnaprimer, line 69: `YouTube`

TASVideos/Pages/Shared/_MovieModule.cshtml
```diff
@@ -55,6 +55,27 @@
...
+		}
+
+		if (lowered.Contains("nicovideo"))
+		{
+			return "On Nicovideo";
```
vadosnaprimer, line 74: To be consistent with youtube, it should read `on` instead of `On`. Also not sure about the difference in leading whitespace.

Masterjun3 (approved):

vadosnaprimer (approved):

PR #1354, opened 2022-07-04
Remove Game fields: SearchKey and GoodName

adelikat:
It was deemed we do not need either of these fields.


adelikat:
Oh, another thing to note is moving YoutubeTags into search, so we should be able to get results for Super Mario Bros from smb, and things like that now.

vadosnaprimer:
Yeah since good names only theoretically make sense for rom versions, it's useless to assign a good tools name to entire games, since it would just be the same as the real name.

vadosnaprimer (approved):

PR #1356, opened 2022-07-04
support /Games/{abbr} as a url instead of just game id

adelikat:
/Games/2 will still work
but now so will /Games/SMB2u
https://media.discordapp.net/attachments/738421558783246467/993571668016517310/unknown.png


Masterjun3:
Not only do abbreviations eventually need to be made unique, we also need to prevent them being numbers.

vadosnaprimer (approved):

PR #1358, opened 2022-07-04
Add "letter" dropdown to Games/List

### LINKED ISSUE #1249: Add "starts with letter" dropdown to https://tasvideos.org/Games/List
Currently showing the whole platform (paginated) is the narrowest it can go unless you just use the search. If we want the search to not be the only real option, we need to allow narrowing down the displayed list further.

I think if you can pick which letter the games should start with, most of the time you won't need to paginate the results, and paginating them makes navigation harder because you don't known where each letter is.

So I'm requesting a second dropdown with letters. Or maybe turn page numbers into letters? Probably harder to do.

With that change (and a couple more refinements) maybe that page will obsolete https://tasvideos.org/Games (and be moved there why not).

vadosnaprimer:
fixes #1249

изображение

Of course ideally we would have ABC paginator instead of numeric one. Until that exists, let's discuss this draft.

### DIFF
--- a/TASVideos.Data/Entity/Game/Game.cs
+++ b/TASVideos.Data/Entity/Game/Game.cs
@@ -60,13 +60,6 @@ public static class GameExtensions
 			: query;
 	}
 
-	public static IQueryable<Game> ForLetter(this IQueryable<Game> query, string? code)
-	{
-		return !string.IsNullOrWhiteSpace(code)
-			? query.Where(g => g.DisplayName.Substring(0, 1) == code)
-			: query;
-	}
-
 	public static void SetGenres(this ICollection<GameGenre> genres, IEnumerable<int> genreIds)
 	{
 		genres.Clear();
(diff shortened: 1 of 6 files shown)


TASVideos.Data/Entity/Game/Game.cs
```diff
@@ -60,6 +60,13 @@ public static IQueryable<Game> ForSystemCodes(this IQueryable<Game> query, IEnum
 			: query;
 	}
 
+	public static IQueryable<Game> ForLetter(this IQueryable<Game> query, string? code)
```
adelikat, line 63: If you are only going to support a single character, the param should be a char type.  However, maybe simpler is just to use .StartsWith(), and then this extension method isn't needed

Masterjun3:
Purely from the images, I'm not sure if "Letter" as a field label is good.
I personally would have made it say "Starts with" (coincidentally the same as the function name), and in the selection instead of "All" I'd put "Any".

adelikat (approved):

PR #1363, opened 2022-07-06
Disable password truncation when registering

### LINKED ISSUE #1361: Registration wrongfully claims password missmatch
When trying to register the site claims the two entered passwords do not match, even though they are both the same copy and pasted password.
The password is randomly generated 128 characters.
The error message is `The password and confirmation password do not match.`
My guess is the actual error is either the length or the used characters.
Happens with:
- `SFcp&vSzHUH&Cw^aQWH@wSQV&ACKBdi6#*PxkmAy8LubnCcTZUg6CZ1nYi80s8*Nxm54M*S*cRZgh8fxIRP@Qnp2$d%6AZrCoD^IpN4T^7D05caCZ1AbxXZq5$IcbOVM`
- `vS@GSlkDcRFffp!Ok#5CI9uXgiIjH&T%hNa^GvkM@ITAq7lu0dDU%&st7Pb!r7IdgyNCFu6w1BF0pzwJs^ou!i3xPugov2K00H3yrl9UXiUOBVdXr!Sf!wCDQ1kzChFc`
- `4W#O3FcY!#NRVz5Td^b$Ck$p6451X!m&XnbencYAg76g1$^2wQZy!bmXfRLZIntQ^bH4QQ#7E0YhneEOZXx9YIM*g@k#zI3MA@Opc!HUD*&OPPn!HCNlFn8WGnsYR4sE`

Registration worked when I reduced the character count to 64.

Masterjun3:
Resolves #1361.
The password limit is still there, but now the user will be notified of it instead of the browser truncating the password.

### DIFF
--- a/TASVideos/Pages/Account/Register.cshtml
+++ b/TASVideos/Pages/Account/Register.cshtml
@@ -25,7 +25,7 @@
 			</form-group>
 			<form-group>
 				<label asp-for="Password"></label>
-				<input asp-for="Password" class="form-control" autocomplete="new-password" />
+				<input asp-for="Password" class="form-control" autocomplete="new-password" maxlength="" />
 				<span asp-validation-for="Password" class="text-danger"></span>
 			</form-group>
 			<form-group>


Masterjun3:
Closed because an improved version is in the works.

PR #1370, opened 2022-07-12
Userfiles

vadosnaprimer:
before

after 

I don't know how to stop showing comments if the userfile is viewed among other userfiles (per game or per user). But when it's viewed on its own, it doesn't make sense to collapse/uncollapse them.

### DIFF
--- a/TASVideos/TagHelpers/CollapsableTagHelper.cs
+++ b/TASVideos/TagHelpers/CollapsableTagHelper.cs
@@ -23,14 +23,6 @@ public class CollapsableHeaderTagHelper : TagHelper
 				<a class='collapsed' data-bs-toggle='collapse' {Attr("href", "#" + BodyId)} aria-expanded='false' aria-controls='collapse1' role='button'>
 					{content}
 				</a>
-				<a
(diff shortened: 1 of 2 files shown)


TASVideos/TagHelpers/CollapsableTagHelper.cs
```diff
@@ -23,14 +23,6 @@ public override async Task ProcessAsync(TagHelperContext context, TagHelperOutpu
...
 					{content}
 				</a>
-				<a
-					data-bs-toggle='collapse'
-					class='collapsed btn btn-default btn-xs text-end'
-					{Attr("href", "#" + BodyId)}
-					{Attr("aria-label", $"Expand/Collapse {BodyId}")}
-					aria-expanded='false'
-					role='button'
-				>
```
vadosnaprimer, line 33: The tag helper had a dark circle icon that's not visible in dark mode, and it was obsoleted by Masterjun icons so I removed the dark icon from the tag helper. And then removed the collapse option from userfile comments lol.

adelikat (approved):

PR #1373, opened 2022-07-14
More userfiles

vadosnaprimer:
hide comments in list view

removed profile link from per user view since user is linked in every userfile metadata where it's present

### DIFF
--- a/TASVideos/Extensions/EntityExtensions.cs
+++ b/TASVideos/Extensions/EntityExtensions.cs
@@ -269,7 +269,7 @@ public static class EntityExtensions
 		});
 	}
 
-	public static IQueryable<UserFileModel> ToUserFileModel(this IQueryable<UserFile> userFiles)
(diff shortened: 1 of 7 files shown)


TASVideos/Extensions/EntityExtensions.cs
```diff
@@ -269,7 +269,7 @@ public static IQueryable<RoleDisplayModel> ToRoleDisplayModel(this IQueryable<Ro
 		});
 	}
 
-	public static IQueryable<UserFileModel> ToUserFileModel(this IQueryable<UserFile> userFiles)
+	public static IQueryable<UserFileModel> ToUserFileModel(this IQueryable<UserFile> userFiles, bool hideComments = true)
```
adelikat, line 272: There's a better way to do this, but I'll approve and merge.

For future reference you can declare this boolean in the page model and the view will have access to it

adelikat (approved):

PR #1378, opened 2022-07-20
Wiki: Turn TOC into a card telling it's a TOC

vadosnaprimer:
If TOC is used non-trivially, it can get incredibly confusing.

https://tasvideos.org/Reviews

изображение

After my change it's now

изображение

I don't know how to add an empty line after the card, should I just append a linebreak? Also I dunno if the empty line after the last marker can be removed. Finally, I had no idea what I was doing in terms of html baking, but at least the intent should be obvious.

### DIFF
--- a/TASVideos.WikiEngine/MakeToc.cs
+++ b/TASVideos.WikiEngine/MakeToc.cs
@@ -18,11 +18,11 @@ public static partial class Builtins
 			.Cast<Element>()
 			.ToList();
 
-		var ret = new Element(charStart, "div") { Attributes = { ["class"] = "card" } };
-		var header = new Element(charStart, "div") { Attributes = { ["class"] = "card-header" } };
 		var strong = new Element(charStart, "strong");
 		strong.Children.Add(new Text(charStart, "Table of contents"));
+		var header = new Element(charStart, "div") { Attributes = { ["class"] = "card-header" } };
 		header.Children.Add(strong);
+		var ret = new Element(charStart, "div") { Attributes = { ["class"] = "card" } };
 		ret.Children.Add(header);
 		
 		var stack = new Stack<Element>();
--- a/TASVideos.WikiEngine/MakeToc.cs
+++ b/TASVideos.WikiEngine/MakeToc.cs
@@ -27,8 +27,6 @@ public static partial class Builtins
 		var body = new Element(charStart, "div") { Attributes = { ["class"] = "card-body" } };
 		ret.Children.Add(body);
 		var stack = new Stack<Element>();
-		stack.Push(ret);
-		stack.Push(header);
 		stack.Push(body);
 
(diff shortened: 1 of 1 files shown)


TASVideos.WikiEngine/MakeToc.cs
```diff
@@ -18,9 +18,18 @@ public static Element MakeToc(List<INode> document, int charStart)
...
+		var strong = new Element(charStart, "strong");
+		strong.Children.Add(new Text(charStart, "Table of contents"));
+		header.Children.Add(strong);
+		ret.Children.Add(header);
+		var body = new Element(charStart, "div") { Attributes = { ["class"] = "card-body" } };
+		ret.Children.Add(body);
 		var stack = new Stack<Element>();
 		stack.Push(ret);
+		stack.Push(header);
+		stack.Push(body);
```
nattthebear, line 32: Does it make more sense to only add `body` to `stack`, not `header` or `ret`?

nattthebear (commented):
The implementation seems fine here, excepting the one small issue noted.  This is an awful way to write HTML, and part of what I wanted to do in wiki "v2" was to have distinct AST and rendering steps, so something like this could be part of rendering only and not the AST creation.  I'm unlikely to ever get to that project, though.  So this is fine.


I'm a bit less positive about the actual problem being solved here.  Why are tocs being placed in locations other than the beginning of the document?  Is this really something we want to do?

TASVideos.WikiEngine/MakeToc.cs
vadosnaprimer: I don't know how it's meant to work really. It's just what I managed to guess from the existing code: it pushes children to stack as new items after having added them to parents. Can you give a brief explanation? I couldn't even find similar stack usage anywhere else in wiki creation.

nattthebear: It uses the stack to track what list element it's in as it traverses the TOC, so that when the TOC pops up a layer, it can pop up from one `<ul>` to the previous `<ul>` and start adding to that again.  The initial contents of the stack are relevant here as whatever is on the top of the stack when entering the loop will be the element that gets the top `<ul>` added to it.  But it doesn't pop up any layers beyond that start, so anything else in the stack isn't used.

If you're really interested, the best way to understand the algorithm might be to step through it with a debugger, writing down the contents of the stack at each step.  It's one of those cute algorithms that isn't really that cute, but is useful.  `private bool StartLineLists()` does some similar things; I think an algorithm like this is inevitable if you're trying to turn a one-dimensional list of "depth numbers" into a hierarchical element tree.

nattthebear (approved):

vadosnaprimer:
Does anyone know how to turn

изображение

into

изображение

?

adelikat:
put a mb-2 on the top thing or a mt-2 on the bottom thing

PR #1384, opened 2022-07-27
Display publisher and judge stats as rows (+tiny tweaks)

vadosnaprimer:
- move profile cards 2px away from each other vertically
- move profile buttons to the right
- add space to replacement text of pub links

изображение

Outdated pic, those buttons are in the reversed order on it, fixed since then:

изображение

### DIFF
--- a/TASVideos/Pages/Activity/Publishers.cshtml
+++ b/TASVideos/Pages/Activity/Publishers.cshtml
@@ -2,8 +2,30 @@
 @{ ViewData["Title"] = $"Publisher Activity for {Model.UserName}"; }
 @model PublishersModel
 
+@{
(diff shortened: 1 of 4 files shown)


TASVideos/Pages/Activity/Publishers.cshtml
```diff
@@ -2,8 +2,30 @@
 @{ ViewData["Title"] = $"Publisher Activity for {Model.UserName}"; }
 @model PublishersModel
 
+@{
+	var counts = new Dictionary<string, int>();
```
adelikat, line 6: the combination of 
.GroupBy()
.ToDictionary()

Will simplify this code a good bit

adelikat (approved):

PR #1392, opened 2022-08-05
Use only \n lineendings for <pre> blocks

Masterjun3:
This circumvents the firefox bug that doubles \r\n in &lt;pre&gt; blocks when copying.

This is the bug from 2009 https://bugzilla.mozilla.org/show_bug.cgi?id=514006


adelikat (approved):
Meh

PR #1400, opened 2022-08-13
Create user setting to always/never watch topic for replies

### LINKED ISSUE #929: Feature Request: Default Options for Watching a Topic when Posting
The way of watching topic has changed, and I'm unsure if this is an issue and/or what is the general consensus on this. I personally would like it to be automatically on for every user, so everyone who participates able to get updates defaultly.

Situation:
- Old site: if you posted in a forum thread, you were "watching the topic" automatically (getting notifications in email with the ability to unsubscribe)
- New site: Opt-in from multiple places

The old site logic lets participants of threads getting update if there's any. 

Some scenarios and their probably preferred topic watching value to be set after posting in a thread:
- General user: watching topic after posting
- Any user in Awards: not watching topic after posting
- Moderator editing someone's post: not watching after editing (Sometimes a moderator might post in the thread for example when the previously moderated post had to be deleted and public address was required. Another thing to decide the nature of the thread, where there might be a new derail is about to happen or just simply an off-topic banter in the off-topic threads)

I'm not sure how UserFile comment notifications works/handled, but similar situations apply there.

adelikat:
Resolves #929 

Adds a new user setting with Always, Never, or Auto.  Auto will fallback to default behavior (in topics, it is defaulting to true, in post create, it is only true if already watching the topic).

UI on the Profile/Settings screen
image

### DIFF
--- a/TASVideos.Data/Entity/User.cs
+++ b/TASVideos.Data/Entity/User.cs
@@ -6,6 +6,11 @@ using TASVideos.Data.Entity.Forum;
 
 namespace TASVideos.Data.Entity;
 
+public enum UserPreference
+{
+	Auto = 0, Always, Never
+}
+
 [ExcludeFromHistory]
 public class User : IdentityUser<int>, ITrackable
 {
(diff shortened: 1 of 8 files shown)


vadosnaprimer (approved):

Masterjun3:
What does the setting "Auto" do?

PR #1416, opened 2022-09-09
Use arrows (→) for breadcrumbs when in forum

### LINKED ISSUE #1086: Do not use misleading breadcrumb visual style in forums
image
This page visually tells people the page link is https://tasvideos.org/EncodingGuide/Logo

But on the forum
image
it now looks like this can be reached by https://tasvideos.org/Forum/BizHawk/BugReporting
But this isn't the case.

When adding breadcrumbs I intentionally didn't add it on the forums for this reason.
The forum navigation used to be next to the title box, and used → for separation, so
```
Forum → BizHawk
[Bug Reporting]
```

Masterjun3:
Resolves #1086 .

This accentuates that
image
can be accessed at Movies/GruefoodDelight, but
image
*can't* be accessed at Forum/Site/GruefoodDelight (because it's Forum/Topics/14102).

### DIFF
--- a/TASVideos/Pages/Shared/_Layout.cshtml
+++ b/TASVideos/Pages/Shared/_Layout.cshtml
@@ -16,6 +16,10 @@
 		  integrity="sha512-SfTiTlX6kk+qitfevl/7LibUOeJWlt9rbyDn92a1DqWOw9vWG2MFoays0sgObmWazO5BQPiFucnnEAjpAB+/Sw=="
 		  crossorigin="anonymous"
 		  referrerpolicy="no-referrer" />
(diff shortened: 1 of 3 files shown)


Masterjun3:
This could technically be done by putting a direct <style></style> into the _Layout instead of creating a whole new file forum.scss and bundling it and all that just for a single line.
But this felt more "proper", I don't know, maybe @TiKevin83 has an opinion on if this is a good way to implement this.

vadosnaprimer (approved):

TiKevin83 (approved):

TiKevin83:
Yeah I like this way of implementing it having a dedicated bundle, thanks!

adelikat (approved):

PR #1421, opened 2022-09-10
Back up textarea in localstorage on Post, Topic, and Submission creation

### LINKED ISSUE #212: Backup typed text

Masterjun3:
Attempts to resolve #212 .
Previous unmerged PR with relevant discussion #977 .

Implements a backup for:
- Post creation
- Topic creation
- Submission creation

Not for *editing*, only for *creating*.

<hr>

It saves the written text into the browser localStorage ~~every 5 seconds~~ whenever the visibily changes to hidden (navigated away, closed tab, closed window, changes tab, changing app, and others).
Upon page load, if it finds a backup, it will show a button the user can click to restore the saved text.
image

~~Once you submit the form, this backup is deleted.~~
The backup is deleted once you open the form again and the **determinator** is different.
This determinator is a value that only changes once the form submission was successful. So here I use: the count of your posts / count of your topics / count of your submissions.

### DIFF
--- a/TASVideos/Pages/Forum/Posts/Create.cshtml
+++ b/TASVideos/Pages/Forum/Posts/Create.cshtml
@@ -45,7 +45,7 @@
 		<span asp-validation-for="Post.Text" class="text-danger"></span>
 	</form-group>
 	<fullrow id="backup-restore" class="d-none">
-		<button id="backup-restore-button" type="button" class="btn btn-secondary" onclick="restoreContent()">Restore Text</button>
+		<button id="backup-restore-button" type="button" class="btn btn-secondary">Restore Text</button>
 		<label class="text-muted">from <span id="backup-time"></span></label>
 	</fullrow>
 	<fullrow class="mt-3">
--- a/TASVideos/Pages/Forum/Posts/Create.cshtml
+++ b/TASVideos/Pages/Forum/Posts/Create.cshtml
@@ -16,6 +16,7 @@
 	}
 </show-more>
 <br />
+<span id="backup-submission-determinator" class="d-none">@Model.BackupSubmissionDeterminator</span>
 <form method="POST" class="backup-form">
 	<input type="hidden" asp-for="Post.TopicTitle" />
 	<row>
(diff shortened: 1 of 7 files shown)


Masterjun3:
The reason for why this is only a draft for now:
**This backup is deleted if you click the submit button. Even if it the server fails. This makes the feature a lot less useful.**

Sadly, from the javascript side of things, there is no way of telling whether submitting the form succeeded or not. And if we never delete the backup, it will always be there, offering you the post you just made.

This is a big problem I couldn't solve yet. I'm open for suggestions on how to implement the deletion of the backup if not on form submission.

adelikat:
S0o my thinking is that we 
1) do not delete on submit
2) implement a reasonable expiry: https://www.sohamkamani.com/javascript/localstorage-with-ttl-expiry/
3) create a key such that successful submits would invalidate it, for instance, wiki edits would include the latest revision, and posts would include the latest post id.  If those aren't the same then it does not restore. The down side of course is that you still lose your work if you attempt to submit, and fail, AND someone posts during that time, but it is still very useful and more than just deleting on submit.

The only alternative is much less reasonable in effort and code complexity. We would use ajax form submits, but then that's a can of worms, plus we have to be able to fallback to a scriptless submit for no scripters, which adds even more complexity.

Masterjun3:
Oh, that gives me an idea for detection whether a post went through:
We add the info how many edits a user has / how many posts they have in that topic.
If this value differs from the saved value, we assume a previous submit went through and can delete the data.

It would be easy to implement. I wonder if I missed anything.

adelikat:
I like this idea

Masterjun3:
I removed the draft state, this PR works now. I also updated the opening post to reflect the recent changes.

adelikat (approved):

PR #1427, opened 2022-09-28
Use server time to save topic visit dates, fixing the unread icon behavior when the difference is only a few seconds

### LINKED ISSUE #1341: New posts icon is shown even if the newest post is mine
изображение

изображение

It would make sense to clear the flag after you've posted in a thread, because it kinda counts as having visited it.

Masterjun3:
Resolves #1341 .

### DIFF
--- a/TASVideos/Pages/Forum/Index.cshtml
+++ b/TASVideos/Pages/Forum/Index.cshtml
@@ -31,9 +31,10 @@
 	}
 </div>
 
+<span id="serverUnixTime" class="d-none">@DateTime.UtcNow.UnixTimestamp()</span>
 @section Scripts {
 	<script>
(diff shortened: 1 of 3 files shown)


Masterjun3:
Now you might ask, Masterjun, you are generating the full html, why aren't you just directly putting the time value into the javascript source code. Instead you're putting the time inside a \<span\> and read the value of that <span> in the script.

Why the roundabout way?

Now here's my answer to this. Doing it this way, the javascript source code is *always the same*. I don't know if there's even any slight compiling or execution performance increase due to caching possibilites. But I just felt weird about having to dynamically generate new script code all the time.

vadosnaprimer (approved):

adelikat (approved):

PR #1430, opened 2022-10-02
Support link to an image

### LINKED ISSUE #218: Wiki Markup:  How to make a text link to an image
The goal is simple: make a link whose `href` points to an image (and so ends with `.png` or another well known image extension), and whose body is plain text:

```
<a href=http://example.com/foo.jpg>Link To Foo</a>
```

Neither the old nor the new wiki directly support this.  The logic looks something like this:

Split the contents of the `"[...]"` by `"|"` into `pp[0]`, `pp[1]`, ...

* `pp[0]` looks like a link and `pp[1]` looks like an image => Make a link to `pp[0]` with `pp[1]` as the link body
* `pp[0]` looks like an image => make an image
* `pp[0]` looks like a link => make a link
* (Other lower priority cases that won't help follow...)
 

None of those give the desired result.  The essential problem is that if the first thing in the [] looks like an image, it will be an image unless the second thing also looks like an image, but then you can't have a text link.

In both cases, old and new, you can trick "looks like an image" by putting a space at the end of the link.  Example page: /SiteHistory/EncoderLogos

Can we do something better about this?

nattthebear:
`[https://foo.bar/io.png|link to io]` now renders as a text link.  Previously it rendered as an image, with `link to io` being ignored, just like on the old site.  Since there's no reason to give an image an unused parameter like this, this disambiguates well between them.  The image alternate would be `[https://foo.bar/io.png|alt=link to io]`

Closes #218.

### DIFF
--- a/TASVideos.WikiEngine/MakeBracketed.cs
+++ b/TASVideos.WikiEngine/MakeBracketed.cs
@@ -201,12 +201,16 @@ public static partial class Builtins
 		var pp = text.Split('|');
 		if (pp.Length >= 2 && IsLink(pp[0]) && IsImage(pp[1]))
 		{
-			return new[] { MakeLink(charStart, charEnd, pp[0], MakeImage(charStart, charEnd, pp, 1)) };
+			return new[] { MakeLink(charStart, charEnd, pp[0], MakeImage(charStart, charEnd, pp, 1, out _)) };
 		}
 
 		if (IsImage(pp[0]))
 		{
-			return new[] { MakeImage(charStart, charEnd, pp, 0) };
+			var node = MakeImage(charStart, charEnd, pp, 0, out var unusedParams);
+			if (!unusedParams)
+			{
+				return new[] { node };
+			}
 		}
 
 		if (IsLink(pp[0]))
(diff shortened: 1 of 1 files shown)


adelikat (approved):

TASVideos.WikiEngine/MakeBracketed.cs
```diff
@@ -201,12 +201,16 @@ private static IEnumerable<INode> MakeLinkOrImage(int charStart, int charEnd, st
 		var pp = text.Split('|');
 		if (pp.Length >= 2 && IsLink(pp[0]) && IsImage(pp[1]))
 		{
-			return new[] { MakeLink(charStart, charEnd, pp[0], MakeImage(charStart, charEnd, pp, 1)) };
+			return new[] { MakeLink(charStart, charEnd, pp[0], MakeImage(charStart, charEnd, pp, 1, out _)) };
```
nattthebear, line 204: Don't examine `unusedParams` here since in this case, there's nothing reasonable to disambiguate to (for now, anyway).

nattthebear:
My localdev isn't working so I can't test this right now.

adelikat:
I did some testing and it works as expected

PR #1438, opened 2022-10-14
Make forum activity icons clickable

### LINKED ISSUE #1429: Forums - Unread post icons should navigate to the unread post
This was a behavior in standard phpbb2 forums that is not present in our current forums.  And this was a feature that has been requested but no issue was made.

(Currently it isn't very feasible but we need to document the request)

Masterjun3:
Resolves #1429 by upgrading the way activity icons work. Instead of locally tracking only the forum visit, we now locally track all "activity posts" the user visits.

A post is an activity post if it was created or edited in the "past few" days. The actual range is determined by the constant `ForumConstants.DaysPostsCountAsActive` and is currently 14 days. Though due to caching it *can* include posts longer ago than that.

If you click on an icon you will directly go to the activity post.
If multiple activity posts belong to the icon, you will be brought to the first one you haven't visited yet.
If you have visited all activity posts, no icon will be shown.

There are both post creations icons (orange) and post edit icons (blue).
The edit icon will only be visible if you already visited all post creations, even if an edit came before a different post.

Additionally, I added the same colors of icons on the posts themselves. Meaning no matter how you navigate there, you will see the activity post icons on the posts themselves. Of course, you will see them only once. Afterwards they're just regular white icons.

image

image

image

### DIFF
--- a/TASVideos.Core/Services/ForumService.cs
+++ b/TASVideos.Core/Services/ForumService.cs
@@ -11,8 +11,8 @@ public interface IForumService
 	Task<PostPositionDto?> GetPostPosition(int postId, bool seeRestricted);
 	Task<IReadOnlyCollection<ForumCategoryDisplayDto>> GetAllCategories();
 	void CacheLatestPost(int forumId, int topicId, LatestPost post);
-	void CacheNewPostActivity2(int forumId, int topicId, int postId, DateTime createTimestamp);
-	void CacheEditedPostActivity2(int forumId, int topicId, int postId, DateTime createTimestamp);
+	void CacheNewPostActivity(int forumId, int topicId, int postId, DateTime createTimestamp);
+	void CacheEditedPostActivity(int forumId, int topicId, int postId, DateTime createTimestamp);
 	void ClearLatestPostCache();
 	void ClearTopicActivityCache();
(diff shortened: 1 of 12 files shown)


vadosnaprimer (approved):
Sounds incredible!

TASVideos.Core/Services/ForumService.cs
```diff
@@ -257,41 +284,72 @@ public async Task<int> CreatePost(PostCreateDto post)
...
+								.Where(fffr => fffr.CreateTimestamp > minimumDate)
+								.ToDictionary(
+									tttkey => tttkey.Id,
+									tttvalue => tttvalue.CreateTimestamp.UnixTimestamp())),
+							JsonSerializer.Serialize(ttvalue
+								.Where(fffr => fffr.PostEditedTimestamp != null && fffr.PostEditedTimestamp > minimumDate)
+								.ToDictionary(
+									tttkey => tttkey.Id,
+									tttvalue => ((DateTime)tttvalue.PostEditedTimestamp).UnixTimestamp()))
+						)));
```
YoshiRulz, line 300: Turning the `tvalue =>` and `ttvalue =>` lambdas into local methods would make this more readable.

PR #1441, opened 2022-10-18
Fix info button contrast

### LINKED ISSUE #1299: Forum tag buttons color in light mode
At some point a cyan color was introduced for some elements on the site, and it looks cool in the dark mode

изображение

But in the light mode the text on cyan buttons is black, which makes it VERY hard to read for me, because button BG is much darker than the rest of the site, and the text is also black. Maybe if the entire BG was cyan, it would have been easier to read the text. But it's even more confusing for the eye right nearby the white text buttons where everything is ideally readable.

изображение

Same issue is with these cyan buttons

изображение

IMO text on those buttons should be made white.

mellonite:
- Fixes #1299 
- Cyan buttons meet WCAG contrast requirements

Dark mode:
image

Light mode:
image


TASVideos/wwwroot/css/darkmode.scss
```diff
@@ -1,4 +1,4 @@
 @import "partials/colors";
 @import "partials/variables";
 @import "partials/darkmode";
-@import "partials/_prism-dark"
+@import "partials/_prism-dark";
```
TiKevin83, line 4: While you're in here can you remove the leading underscore for consistency?

TASVideos/wwwroot/css/partials/_darkmode.scss
```diff
@@ -110,4 +110,12 @@ body {
 	&-white {
 		filter: none;
 	}
-}
\ No newline at end of file
+}
+
+.btn-info {
```
TiKevin83, line 115: The dark mode file should minimize custom code and focus on variable overrides, if you want the info text to always be light in both themes I would define another variable in the bootstrap_overrides to use like info-text instead of bs-white so it doesn't get changed by the theme. Alternatively since you're already changing the colors you could make the info blue light enough for black text to be wcag AA compliant on light theme and use the same color as text-dark did before

mellonite: I think having the text be white across all of the buttons there would look nice and potentially be more readable, I can go ahead and use a new variable as you suggested.

Masterjun3:
Btw, this change now makes the the contrast check on the topics grid fail.
Maybe these global color changes should be more thoroughly checked before approving changes, otherwise we'll just go back and forth between colors fixing one and then the other.
image

mellonite:
Good catch, you're right that it's the fault of making a global color change. I unfortunately was only checking btn-info and bg-info elements. Perhaps new variables would be better suited for changes like this.

For now I'll go ahead and fix this instance and check where else uses the info colors.

mellonite:
While technically it does meet WCAG AA contrast for large text, I'm inclined to agree with you to change the color. This particular color is tricky as getting the info color dark enough to work with white text AND to work as text on a dark background would take some fiddling.

Thoughts on changing the color of the sticky wording to something else, perhaps a yellow/text-warning? I believe that would have it stand out better against the blue as well.

I can also revert the changes to the info colors and create new variables specifically for what was info buttons, leaving the sticky to the original info text.

We could also make the "Sticky" itself have its own background, similar to the "Featured" tag on the front page.

Masterjun3:
> Thoughts on changing the color of the sticky wording to something else, perhaps a yellow/text-warning?

We already have that.
image


> I'm inclined to agree with you to change the color

No I didn't even say I wanted to change it. I just said we shouldn't blindly resolve an issue ticket without considering all effects it will have. (Not all tickets are even a *good* change in the first place.)
And too many color changes after another are also just a plain bad user experience for people who visit a lot. So I'd even want to say, just leave it at this point.

PR #1442, opened 2022-10-19
Make awards on profile page original size

### LINKED ISSUE #1233: Awards on Profile page should be original size
Currently https://tasvideos.org/Profile page shows the -2x.png ending Award images ~because~ and the max-height is set to 48px instead of 92px.

mellonite:
Fixes #1233

Works while viewing your own profile page, or other user's

### DIFF
--- a/TASVideos/Pages/Shared/_Award.cshtml
+++ b/TASVideos/Pages/Shared/_Award.cshtml
@@ -1,9 +1,24 @@
 ﻿@model TASVideos.Core.Services.IAwardAssignmentSummary
 @{
 	string awardName = $"{Model.ShortName}_{Model.Year.ToString()}";
+	string pageGroup = ViewContext.PageGroup();
 }
-<a href="/Awards/@Model.Year.ToString()#@(Model.ShortName)_@Model.Year" title="Award - @Model.Description"><img style="max-height: 48px;" 
+@if (pageGroup == "Users" || pageGroup == "Profile")
+{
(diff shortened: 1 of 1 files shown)


Masterjun3 (changes requested):
Is it necessary to repeat the whole code block again? What was the problem with just making a conditional on the max-height? Do the x descriptors really need to be changed, too?

mellonite:
If you only change the max height, the awards look like this:
image

The solution in the PR results in this:

mellonite:
Sorry for the double comment, accidentally submitted a comment before I finished writing it up.

Awards look like this if you only change the max height:
image

The solution in the PR results in the following:
image

I felt that the second one looked like a crisper solution. Otherwise it appears to show the awards at double size instead of original.

Masterjun3:
That's because the original srcset makes it so it takes the bigger size and then scale it down to fit your max height. See how in this PR the max height of 92px *doesn't even do anything*.
In the "blurry" image you posted it did what the code was designed to do, fill out the height you set by choosing the bigger one and scaling down.

It makes sense to have crispness on this page where we have enough space for it, but then you want to remove the max-height, since there is no good reason for it.
I'd put the conditional only around the  tag, not around the whole <a>, since that's just repeated code. I'd personally also calculate the boolean in the code above and only use the variable for the conditional.

adelikat:
reworked this approach with the code suggestion by @Masterjun3 in cd7f21a6775c0f2f7d3bb9ed87c24664fa756c89
@mellonite, thanks for the contribution 👍

PR #1443, opened 2022-10-19
Make titles bigger on certain pages

### LINKED ISSUE #1087: Make titles bigger on certain pages
On busy pages like submission threads or submission pages the title can get buried in the design of surrounding elements.
I made them bigger to counteract this, but the size increase was removed in the title redesign.

Example:
image
In this image the status is bigger than the title, making the status, judge, and publisher look more important than the actual title and author.

mellonite:
Fixes #1087

Title is now larger than Status on Submission Page
image

This fix addresses other pages that use header cards as well.

### DIFF
--- a/TASVideos/wwwroot/css/partials/_bootstrap-overrides.scss
+++ b/TASVideos/wwwroot/css/partials/_bootstrap-overrides.scss
@@ -38,6 +38,10 @@ h1 {
 	.card & {
 		font-size: 1em;
 	}
+
+	&.card {
+		font-size: 1.75em;
+	}
 }
 
 h2 {


Masterjun3 (changes requested):
This changes all card sizes. This is not what I wanted with the issue.
Even worse, with this change the titles size are inconsistent. The Games List page is now bigger than the Movies page for example.
The submission threads I mentioned in the issue aren't even changed, so this doesn't even fix the issue at all.

mellonite:
My apologies, I'm still new to the site, and it seems I didn't understand what you meant by submission threads.

You're right, the use of h1 cards is inconsistent between pages. This is not the ideal implementation for this site.

How do I get to a submission thread from the local development version of the site?

adelikat:
You can simply create a submission, and the thread will get created.  You just need a parsable movie file

adelikat:
Rejecting this PR as it takes the wrong approach to the problem as mentioned above

TiKevin83:
> You can simply create a submission, and the thread will get created. You just need a parsable movie file

for reference I tend to just download a recent submission's movie file to test this locally, it's pretty straightforward and you don't need to learn how to spin up bizhawk to create one

PR #1448, opened 2022-10-21
Don't split usernames on homepages

### LINKED ISSUE #1118: Homepages page fill usernames with spaces before letters/numbers
https://tasvideos.org/Homepages

It looks like this is the only root/directory page, where usernames are get the same treat as normal wiki titles should get. But these usernames should stay the same as the user wrote it.

https://tasvideos.org/HomePages/ViGadeomes has Vi Gadeomes
https://tasvideos.org/HomePages/ais523 has ais 523
https://tasvideos.org/HomePages/AngerFist has Anger Fist
https://tasvideos.org/HomePages/TiKevin83 has Ti Kevin 83

mellonite:
Fixes #1118

Cases where names aren't split anymore:

List of home pages
image

Breadcrumb on home page & its sub pages
image

Breadcrumb for subpage list on home page
image

### DIFF
--- a/TASVideos.Common/Extensions/StringExtensions.cs
+++ b/TASVideos.Common/Extensions/StringExtensions.cs
@@ -51,6 +51,18 @@ public static class StringExtensions
 			: str[..limit];
 	}
 
+	public static string SplitCamelCase(this string? str, bool isHomePage)
+	{
+		if (!isHomePage || string.IsNullOrEmpty(str))
+		{
+			return SplitCamelCase(str);
+		}
+
+		// Don't split usernames on homepages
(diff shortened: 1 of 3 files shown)


Masterjun3 (changes requested):
I personally don't think carrying an isHomePage flag throughout the code is the correct way to implement this. This whole thing would have to be completely reworked if we ever have another set of pages where usernames show up.
If this is to be implemented, the best way would be to make the front-end ignorant about the types of paths it splits by. And in fact, that can be implemented in less code changes than this.

Additionally, we have SplitCamelCase tests that could be added to if we implement new behaviors.

PR #1449, opened 2022-10-21
Don't split Usernames in Path Displays

### LINKED ISSUE #1118: Homepages page fill usernames with spaces before letters/numbers

Masterjun3:
Resolves #1118 .
This PR solves the same issue as PR #1448 , but instead of keeping a isHomePage flag throughout the code, this code changes the frontend to be independent of the path it has to display.
Instead of splitting the path and CamelCasing every element, it first CamelCases the whole path and afterward splits and displays every element. The advantage is that you have the whole path information available for the CamelCase operation, so you can conditionally skip converting certain elements.

If we want to be super convenient, we'd make an entire helper class to do all this for us. The code of this PR would be easy to change into that, if we ever decide that's what we want.

### Before | After
image

image

image

### DIFF
--- a/TASVideos.Common/Extensions/StringExtensions.cs
+++ b/TASVideos.Common/Extensions/StringExtensions.cs
@@ -91,21 +91,21 @@ public static class StringExtensions
 			return "";
 		}
 
-		if (str.StartsWith("HomePages/"))
+		if (!str.StartsWith("HomePages/"))
 		{
-			var pathFragments = str.SplitWithEmpty("/");
-			for (int i = 0; i < pathFragments.Length; i++)
-			{
-				if (i != 1)
-				{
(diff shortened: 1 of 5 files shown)


vadosnaprimer:
Does it work with users that have a space as a part of their username? Can't find any with homepages tho...

adelikat (approved):

PR #1456, opened 2022-10-31
Make Topic Reply Notification Email template a bit more user friendly

Masterjun3:
Includes better spacing and more intuitive format (most important link is the biggest one).
It also may fix gmail accidentally grouping different topic reply notifications together due to having the same email text. Now the link text is different, so that hopefully prevents ~~the grouping~~ Edit: see comment below.

I'm fine with suggestions for spelling or expressions. The main purpose of this PR is the change in link formatting and positioning.

<hr>

Before:
image

After:
image


vadosnaprimer (approved):
Looks great!

To clarify: the problem is not that gmail groups them (it should) but that it hides text if the new message is the same as the previous in the group.

PR #1457, opened 2022-11-01
add note icons for alternative class

vadosnaprimer:


Masterjun3:
Since these icons are used for the replacement of Moons, do we still need the moon.png icons?

vadosnaprimer:
Right now maybe not, but I wouldn't want to recreate them manually if we decide to use them in the future.

Masterjun3:
I mean they're version controlled, you can't really lose them anymore.

Masterjun3 (approved):

adelikat (approved):

PR #1465, opened 2022-11-22
Fix bk2/fm2/dsm parsing when the input log file is >1GB large

CasualPokePlayer:
This is currently blocking me from submitting max score Desert Bus :)


Masterjun3:
Wouldn't this be one of those cases where a comment in the code would be very useful for explaining why a roundabout method was used to count? (Else it might be accidentally reverted by a refactor.)

adelikat (approved):

PR #1481, opened 2022-12-02
Sort systems by display name, redirect to system page after creation

### LINKED ISSUE #1479: After system creation, go it its page
You create a system at https://tasvideos.org/Systems/Create and then it redirects you back to https://tasvideos.org/Systems and you have to scroll all the way down to find it again and to edit it.

### LINKED ISSUE #1480: Sort systems by display name in the Systems list
https://tasvideos.org/Systems

Sorting by ID makes them harder to find because it's only dependent on creation order, which is arbitrary. Sorting them by Code would also be less handy because you have to remember which code it's meant to have (manufacturer is not a part of the code, and some codes are also shortened). But if they are sorted by full name, manufacturer goes first and then it's a matter of alphabetic order that's easy to navigate, and you get to know the official full names too.

GMPranav:
Resolves #1479 and also Resolves #1480

### DIFF
--- a/TASVideos/Pages/Systems/Index.cshtml.cs
+++ b/TASVideos/Pages/Systems/Index.cshtml.cs
@@ -17,7 +17,7 @@ public class IndexModel : BasePageModel
 	public async Task OnGet()
 	{
 		Systems = (await _systemService.GetAll())
-			.OrderBy(s => s.Id)
+			.OrderBy(s => s.DisplayName)
 			.ToList();
 	}
 }
--- a/TASVideos/Pages/Systems/Create.cshtml.cs
+++ b/TASVideos/Pages/Systems/Create.cshtml.cs
@@ -38,7 +38,7 @@ public class CreateModel : BasePageModel
 			default:
 			case SystemEditResult.Success:
 				SuccessStatusMessage($"System {System.Code} successfully created.");
-				return BasePageRedirect("Index");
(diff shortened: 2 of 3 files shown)


TASVideos/Pages/Systems/Index.cshtml.cs
```diff
@@ -17,7 +17,7 @@ public IndexModel(IGameSystemService systemService)
 	public async Task OnGet()
 	{
 		Systems = (await _systemService.GetAll())
-			.OrderBy(s => s.Id)
+			.OrderBy(s => s.DisplayName)
```
vadosnaprimer, line 20: Will it also make it the first column? May not be important, but probably nicer for the user (sorry I forgot to mention it originally).

GMPranav: No it will not. But we can change that as well.

vadosnaprimer: Pretty please :) It makes sense when column order reflects sorting priority.

GMPranav: Looks good?
image

vadosnaprimer: Perfect!

vadosnaprimer (approved):

adelikat (approved):

PR #1483, opened 2022-12-03
Show official games first in search

### LINKED ISSUE #1260: Small tweak in sort algorithm for search results
Currently, the search results show the first 10 relevant results and the sort seems to be alphabetical. This is a small problem in games with a lot of ROM hacks like Super Mario Bros. for example. the original game itself gets submerged.

One solution I can propose is to sort the official games in alphabetical order first and only then list the unofficial games. If a better solution is thought of, that is fine as well.

GMPranav:
Resolves #1260

Basically for example when we search "super mario bros" the actual game is not even visible in the first page as there a lot of ROM hacks that take priority alphabetically. So I am checking whether "Unofficial Game" is one of the genres of the game entry and if so they will be displayed at the end of the list.

### DIFF
--- a/TASVideos/Pages/Search/Index.cshtml.cs
+++ b/TASVideos/Pages/Search/Index.cshtml.cs
@@ -73,7 +73,7 @@ public class IndexModel : BasePageModel
 
 			GameResults = await _db.Games
 				.Where(g => EF.Functions.ToTsVector(g.DisplayName + " || " + g.Aliases + " || " + g.Abbreviation).Matches(EF.Functions.WebSearchToTsQuery(SearchTerms)))
-				.OrderBy(g => g.DisplayName)
+				.OrderBy(g => g.GameGenres.Select(gg => gg.GenreId).Contains(14)).ThenBy(g => g.DisplayName)
 				.Skip(skip)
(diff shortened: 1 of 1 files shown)


TASVideos/Pages/Search/Index.cshtml.cs
```diff
@@ -73,7 +73,7 @@ public async Task<IActionResult> OnGet()
 
 			GameResults = await _db.Games
 				.Where(g => EF.Functions.ToTsVector(g.DisplayName + " || " + g.Aliases + " || " + g.Abbreviation).Matches(EF.Functions.WebSearchToTsQuery(SearchTerms)))
-				.OrderBy(g => g.DisplayName)
+				.OrderBy(g => g.GameGenres.Select(gg => gg.GenreId).Contains(14)).ThenBy(g => g.DisplayName)
```
vadosnaprimer, line 76: Should be better to use a named constant, or even pull human readable meaning of the value from genre strings, if it's there.

GMPranav: Better now? I used the Display Name property instead of Id

adelikat (approved):

PR #1485, opened 2022-12-05
add movie classes link to navbar

vadosnaprimer:


TASVideos/Pages/Shared/_NavBarPartial.cshtml
```diff
@@ -7,6 +7,7 @@
 		<a id="Games" class="nav-link" asp-page="/Games/List">Games</a>
 	</nav-item>
 	<nav-dropdown id="MoviesDropdown" activate="Movies">
+		<a id="Publications" class="dropdown-item" href="/Class">Movie Classes</a>
```
Masterjun3, line 10: We probably want to avoid two elements with the same id in an html document.

adelikat (approved):

PR #1513, opened 2023-01-29
Always sort Two Column Select alphabetically by the display text

### LINKED ISSUE #1496: Group list sorting gets lost after you've picked a group
https://tasvideos.org/Games/Edit/2584?returnUrl=%2F2584G

изображение

Masterjun3:
Resolves #1496 .
This change always sorts two column select, and also sorts them alphabetically instead of by the value integer.

For the most control, ideally the javascript would keep the initial order, so we provide an arbitrary ordering. But I had like 10 minutes to implement a fix someone wanted. so I did this. Alphabetical ordering should always be fine anyway.


vadosnaprimer (approved):

adelikat:
I don't like this, the core level code assumes the calling code would always want alphabetical, but it seems to work out to something we want in each case.  I'll merge, somewhat begrudgingly, especially knowing that another component should eventually replace this one

PR #1514, opened 2023-02-04
Frontpage Tables Redesign

Masterjun3:
Changes the frontpage tables into more mobile-friendly and colorful designs.

Before | After
image

image

image


adelikat (approved):

vadosnaprimer (approved):
This feels much better than the [initial version](https://user-images.githubusercontent.com/7092625/216759874-deb92b79-8af9-4963-9bc4-59b6ed2a9555.png), even if it's mostly similar.

vadosnaprimer:
I'd only ask if it's feasible to make screenshots bigger.

Masterjun3:
Bigger publication images is a bit of a problem due to the height the whole thing is going to end up taking. You'll end up having a lot of empty space and padding.
It's still supposed to be a compact list, after all.

PR #1519, opened 2023-02-13
Making only Experienced Forum Users be able to set signatures

### LINKED ISSUE #1517: Restrict signatures completely until an amount of posts.
Making it easier to go after bot accounts as there's some that commonly have their signatures display the spam material without ever posting.
Possibly only needs at least 4 (Newbie) posts...

GMPranav:
Resolves #1517

### DIFF
--- a/TASVideos/Pages/Profile/Settings.cshtml
+++ b/TASVideos/Pages/Profile/Settings.cshtml
@@ -120,10 +120,13 @@
 			</small>
 		</column>
 	</row>
-	<fullrow class="mt-3">
-		<label asp-for="Settings.Signature"></label>
-		<textarea asp-for="Settings.Signature" class="form-control" rows="5"></textarea>
-	</fullrow>
+	@if (Model.Settings.Roles.Where(role => role.Name == "Experienced Forum User").Count() != 0)
+	{
+		<fullrow class="mt-3">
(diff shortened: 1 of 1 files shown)


Masterjun3 (changes requested):
Wasn't the point of our [Permission](https://github.com/TASVideos/tasvideos/blob/main/TASVideos.Data/Entity/PermissionTo.cs) system to add specific permissions instead of hardcoding roles?

GMPranav:
I was only trying to implement @CasualPokePlayer's idea. I was not aware of that permission system and also find it hard to decipher it.

adelikat:
@Masterjun3 is correct, we must add a new permission, and then check that permission when choosing to display things.

Checking roles means that a simple name change on a role can break functionality

Masterjun3:
I know this is closed already, but I missed a very important part when writing my feedback:
This PR only changes the visibility of the according element in the HTML. A "malicious" user could just reinsert that specific part and be able to change their signature without problems.

In general our idea is to *never trust the user*, so permission things should **always** be done server-side.

GMPranav:
Yeah, that was quite the dirty quickfix

adelikat:
To be clear, having the code in the view is good, but it just needs the check server side on the POST call in addition

PR #1528, opened 2023-03-08
Automatically Trim all Model Binding Strings

Masterjun3:
This adds a string [Custom Model Binding](https://learn.microsoft.com/en-us/aspnet/core/mvc/advanced/custom-model-binding) which calls Trim() on all strings before binding to the property, deleting all whitespace from the beginning and end.
Does not trim any properties with the new `DoNotTrimAttribute`, and also does not trim any properties with the `DataType(DataType.Password)` attribute.

The PR adds the DoNotTrim attribute to every Wiki Markup we have, because a space at the beginning is used to start a non-formatted block in our Wiki Markup. (Debatable if this is a necessary feature.)

<hr>

Why only a draft:
- Because this affects *everything* I have only tested comparably few things here. I want to do a bit more testing.
- `TrimStringModelBinder` and `TrimStringModelBinderProvider` are in the same file, probably bad practice.
- ~~I put the whole thing in TASVideos.Core because that's where the `services.AddControllers` was. So every `DoNotTrim` needs to add a new reference to TASVideos.Core now, not sure if that's fine or if there can be a better solution.~~
Fixed by moving it into the main TASVideos project and using `AddMvc` instead of `AddControllers`.

### DIFF
--- a/TASVideos/Models/TrimStringModelBinder.cs
+++ b/TASVideos/Models/TrimStringModelBinder.cs
@@ -1,8 +1,8 @@
-﻿using Microsoft.AspNetCore.Mvc.ModelBinding;
+﻿using System.ComponentModel.DataAnnotations;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
 using Microsoft.AspNetCore.Mvc.ModelBinding.Metadata;
-using System.ComponentModel.DataAnnotations;
 
-namespace TASVideos.Core.Services;
+namespace TASVideos.Models;
 
 public class TrimStringModelBinder : IModelBinder
 {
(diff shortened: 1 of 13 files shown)


adelikat (approved):

TASVideos/Models/TrimStringModelBinder.cs
```diff
@@ -0,0 +1,36 @@
...
+using Microsoft.AspNetCore.Mvc.ModelBinding.Metadata;
+
+namespace TASVideos.Models;
+
+public class TrimStringModelBinder : IModelBinder
+{
+	public Task BindModelAsync(ModelBindingContext bindingContext)
+	{
+		string? value = bindingContext.ValueProvider.GetValue(bindingContext.ModelName).FirstValue;
+		if (!string.IsNullOrWhiteSpace(value))
```
nattthebear, line 12: This will allow strings with `Length > 0` that are entirely whitespace through.  Is that intended?

Masterjun3: Yes, whitespace strings are already handled in default Asp Net Core stuff. They have weird behavior so I changed the code to only Trim actual content. See how it was before f2edc39b658df491d5d2f5996a316c8a506bbe28 .

PR #1540, opened 2023-04-04
Properly bust caches for darkmode stylesheet

Masterjun3:
Our WebOptimizer uses version string hashes on our included stylesheets, so that when they change, browsers *have* to reload them because the filename changed.
However, this seemed to not apply to our darkmode stylesheet:

image

The reason for this is because we never included the stylesheet in our `_Layout.cshtml`, and only added it dynamically based on user local storage and whether they choose darkmode. They were missing a version hash, and we would have to add more crazy code to compute it ourselves. And without the version hash, the browser would not reload styles when they changed.

This PR changes this behavior by including the stylesheet in the file, but *disabled*! That way, WebOptimizer is able to assign the hash, and we can dynamically enable and disable the stylesheets, instead of creating and removing html elements.


TiKevin83 (approved):
Approving, the benefit in cache busting matters more than the tiny size of always getting the darkmode stylesheet over the network

PR #1547, opened 2023-04-14
Make sure doubles are converted to strings invariantly when necessary

### LINKED ISSUE #1515: Media publisher strings should be invariant, instead of depending on the users locale
Messages go out to people everywhere, so the invariant strings should be used.

image
vs
image

Masterjun3:
- Resolves #1515 .
- Also resolves a newly found bug where people with comma as decimal separators did not see forum post text size changes. `[size=7]Hello.[/size]` was shown as regular sized text.

<hr>

And now I want to ramble. So here we go:
I would like to point out that always using locale by default for conversions was a very wrong choice. The fact that `$"font-size: {fontsize};"` can go wrong is absurd. I've already had to deal with so many bugs that appeared just because all the devs are only on the US locale which is also the one where everything works. If I had the choice, I'd make a middleware at the very top, add `CultureInfo.CurrentCulture = CultureInfo.InvariantCulture` to remove any custom cultures, and have fixed every past and future bug. I'm not saying this is the fault of our code, I just dislike the choice of .NET to localize everything automatically and by default.

<hr>

Oh maybe someone is wondering how I made sure to catch every instance of this. Well I wrote my own full Roslyn Analyzer to catch every double to string conversion, both explicit and implicit, and then analyzed the whole code base.

How do we make sure these kinds of bugs don't appear in the future? We can't! Hurray localization.

### DIFF
--- a/TASVideos.ForumEngine/Node.cs
+++ b/TASVideos.ForumEngine/Node.cs
@@ -1,4 +1,5 @@
-﻿using System.Text;
+﻿using System.Globalization;
+using System.Text;
 using System.Web;
 using TASVideos.Common;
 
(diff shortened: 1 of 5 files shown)


nattthebear (approved):
> And now I want to ramble. So here we go:

Yes, no argument here.  Or from zeromus, for sure.  Bizhawk was hit by this one many times.  I suspect the default is this way so as to not feel like you're disrespecting other cultures, but in practical real code it usually has nothing to do with that.

adelikat (approved):

PR #1555, opened 2023-05-01
Avatar preview

vadosnaprimer:
turn avatar URLs and requirements into fullrows:
avatar URLs are wide fields, and requirement text is editable wiki text of uncapped height, so they look weird if put side by side

hide signature for users without the permission instead of disabling it

use consistent label for normal avatar url

before

изображение

after

изображение

### DIFF
--- a/TASVideos/Pages/Profile/Settings.cshtml
+++ b/TASVideos/Pages/Profile/Settings.cshtml
@@ -96,12 +96,14 @@
 		<label asp-for="Settings.Avatar"></label>
 		<input type="text" asp-for="Settings.Avatar" class="form-control" />
 	</fullrow>
-	<fullrow class="mt-2" id="avatar-preview" show="!string.IsNullOrWhiteSpace(Model.Settings.Avatar)">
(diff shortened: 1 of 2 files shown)


TASVideos/Pages/Profile/Settings.cshtml
```diff
@@ -92,37 +92,33 @@
...
-			<small>
-				@await Component.RenderWiki("System/MoodAvatarRequirements")
-			</small>
-		</column>
-	</row>
 	<fullrow class="mt-3">
+		<label asp-for="Settings.Avatar"></label>
+		<input type="text" asp-for="Settings.Avatar" class="form-control" />
+	</fullrow>
+	<fullrow class="mt-2" id="avatar-preview" show="!string.IsNullOrWhiteSpace(Model.Settings.Avatar)">
```
vadosnaprimer, line 99: `show` doesn't work because tag helpers don't seem to print it out

adelikat (approved):

PR #1567, opened 2023-06-04
Fix: Users with email not confirmed can't request a resend

VelpaChallenger:
If a new user is registering and due to site errors is not able to receive the confirmation email but was still registered in the database, then they are not able to login, even if an admin gives login permissions, since only users with email confirmed can login. Additionally, if the user is registering and then closes the browser, they wouldn't be able to access the EmailConfirmationSent page either. For this reason, an update was made so that if users do not have their email confirmed but they exist in the database, then they will be redirected to the EmailConfirmationSent page, so that they can request a resend, only if the username and email entered match.

### DIFF
--- a/TASVideos/Pages/Account/EmailConfirmationSent.cshtml
+++ b/TASVideos/Pages/Account/EmailConfirmationSent.cshtml
@@ -1,6 +1,35 @@
 ﻿@page
+@addTagHelper *, AspNetCore.ReCaptcha
+@model EmailConfirmationSentModel
 @{
 	ViewData.SetTitle("Email Sent");
 }
 
 <p>To complete the registration process look for an email in your inbox that will provide further instructions.</p>
+
+<p>In case you didn't receive any email, please enter your username and email below and we will resend it to you.</p>
+
+<form method="post">
+	<row>
+		<column md="6">
+			<form-group>
+				<label asp-for="UserName"></label>
+				<input asp-for="UserName" class="form-control" autocomplete="username" />
+				<span asp-validation-for="UserName" class="text-danger"></span>
+			</form-group>
+			<form-group>
+				<label asp-for="Email"></label>
+				<input asp-for="Email" class="form-control" autocomplete="off" />
+				<span asp-validation-for="Email" class="text-danger"></span>
+			</form-group>
+			<environment exclude="Development">
+				<form-group>
+					<recaptcha />
+				</form-group>
+			</environment>
+		</column>
+	</row>
+	<div class="text-center">
+		<button type="submit" class="btn btn-secondary">Resend Confirmation Email</button>
+	</div>
+</form>
--- a/TASVideos/Pages/Account/EmailConfirmationSent.cshtml
+++ b/TASVideos/Pages/Account/EmailConfirmationSent.cshtml
@@ -5,9 +5,7 @@
 	ViewData.SetTitle("Email Sent");
 }
 
-<p>To complete the registration process look for an email in your inbox that will provide further instructions.</p>
-
-<p>In case you didn't receive any email, please enter your username and email below and we will resend it to you.</p>
+@await Component.RenderWiki("System/EmailConfirmationSentMessage")
 
 <form method="post">
 	<row>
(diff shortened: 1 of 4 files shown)


TASVideos/Pages/Account/EmailConfirmationSent.cshtml
```diff
@@ -1,6 +1,35 @@
...
+				<label asp-for="UserName"></label>
+				<input asp-for="UserName" class="form-control" autocomplete="username" />
+				<span asp-validation-for="UserName" class="text-danger"></span>
+			</form-group>
+			<form-group>
+				<label asp-for="Email"></label>
+				<input asp-for="Email" class="form-control" autocomplete="off" />
+				<span asp-validation-for="Email" class="text-danger"></span>
+			</form-group>
+			<environment exclude="Development">
```
adelikat, line 25: nice idea to put a captcha on this page

TASVideos/Pages/Account/EmailConfirmationSent.cshtml.cs
```diff
@@ -0,0 +1,69 @@
...
+	[BindProperty]
+	[Required]
+	[EmailAddress]
+	[Display(Name = "Email")]
+	public string Email { get; set; } = "";
+	public async Task<IActionResult> OnPost()
+	{
+		if (!await _signInManager.EmailAndUserNameMatch(UserName, Email))
+		{
+			ModelState.AddModelError(nameof(Email), "Username and email entered do not match. Please try again.");
```
adelikat, line 43: Unfortunately we can't return a message like this, especially on an unauthenticated page.  It could be used to sniff for emails (though it would be a lot harder with the username matching).  

I think we have no choice but to just return as if the email was sent, and adjust the wording on the page to say something like "if the username and email match, and email will be sent"

TASVideos/Pages/Account/EmailConfirmationSent.cshtml.cs
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		if (!ModelState.IsValid)
+		{
+			return Page();
+		}
+
+		if (ModelState.IsValid)
+		{
+			var user = _db.Users.Where(u => u.Email == Email && u.UserName == UserName).FirstOrDefault();
```
adelikat, line 53: this can be simplified to this:
var user = _db.Users.FirstOrDefault(u => u.Email == Email && u.UserName == UserName);

Also, I'd prefer SingleOrDefaultAsync, to be consistent.  But admittedly it's a nitpick difference in situations like this

TASVideos/Pages/Account/EmailConfirmationSent.cshtml.cs
```diff
@@ -0,0 +1,69 @@
...
+			if (user != null)
```
adelikat, line 54: nitpick: for consistency with the rest of the code
"user is not null" is the preferred way to check for null

adelikat (changes requested):
Requesting changes for the model validation comment.  The rest are suggestions and nitpicks :)

Masterjun3:
What happens (or should happen) if someone with an already confirmed email tries this?

adelikat:
Good point, the code shoudl check if the user has a confirmed email and not send it, if so

TASVideos/Pages/Account/EmailConfirmationSent.cshtml.cs
VelpaChallenger: Done, updated the code so that users can't see whether the validation failed, but they can still know that the button click was received.

TASVideos/Pages/Account/EmailConfirmationSent.cshtml.cs
VelpaChallenger: Makes sense, I updated it to use the SingleOrDefaultAsync method.

TASVideos/Pages/Account/EmailConfirmationSent.cshtml.cs
VelpaChallenger: Agree, I copied part of the code from the Login.cshtml.cs file, and I saw a user != null there, but yes, I too prefer "user is not null", so I just updated it!

adelikat (approved):

VelpaChallenger:
@Masterjun3 Thank you for bringing it up! We discussed this with @adelikat and we decided to add an OnGet() method, so that if the user is logged in, they are redirected to the main page directly, given there are no other ways to get to this page other than the Register and Login page, so they can only come here on purpose.

PR #1576, opened 2023-06-15
Forum editor: syntax highlighting POC

### LINKED ISSUE #1574: [F.Rq.] Syntax highlighting for forum markup
I'm thinking JS could replace the textarea with a richer UI, eventually a MediaWiki-style editor, but initially it could be as simple as underlining BBCode tags in red or grey depending on whether they're in the list of valid tags. I've mocked that up here:

mockup

nattthebear:
As far as I know, this is the main "trick" used to do things like this.  If you're not willing to go all the way to contenteditable hell, you just hide the `<textarea>` and then put something that exactly mirrors its content.  I'm not convinced we should do this; it's hard to make it work out just right in every browser, and there may be accessibility concerns.

If we do decide to implement this, a more detailed syntax highlighting model is easily added.

Closes #1574.

### DIFF
--- a/TASVideos/Pages/Forum/Posts/Create.cshtml
+++ b/TASVideos/Pages/Forum/Posts/Create.cshtml
@@ -41,7 +41,7 @@
 		<partial name="_CreatePostHelper" model="@("Post_Text")" />
 	</fullrow>
 	<form-group>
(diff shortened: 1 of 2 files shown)


nattthebear:
I don't think this implementation idea is good enough to continue through with, but if anyone else feels like taking a shot at it, go ahead.

PR #1587, opened 2023-06-29
Display "unknown" when submission rerecords are 0

### LINKED ISSUE #1572: Display "unknown" when submission rerecords are 0
Old site used to have this logic, and we used it as a reminder for encoders to not put 0 into encode subtitles (because that wouldn't make sense). Currently the parser simply says 0.

vadosnaprimer:
also reset warnings upon submission replacement

fixes #1572

### DIFF
--- a/TASVideos.Core/Services/QueueService.cs
+++ b/TASVideos.Core/Services/QueueService.cs
@@ -400,6 +400,7 @@ internal class QueueService : IQueueService
 		submission.System = system;
 		submission.CycleCount = parseResult.CycleCount;
 		var warnings = parseResult.Warnings.ToList();
+		submission.Warnings = null;
 		if (warnings.Any())
 		{
 			submission.Warnings = string.Join(",", warnings);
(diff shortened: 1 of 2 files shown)


TASVideos.Core/Services/QueueService.cs
```diff
@@ -400,6 +400,7 @@ public async Task<string> MapParsedResult(IParseResult parseResult, Submission s
 		submission.System = system;
 		submission.CycleCount = parseResult.CycleCount;
 		var warnings = parseResult.Warnings.ToList();
+		submission.Warnings = null;
```
adelikat, line 403: what's going on here?

TASVideos/Pages/Submissions/View.cshtml
```diff
@@ -158,7 +158,10 @@
 				<th>@Html.DisplayNameFor(m => m.Submission.Branch)</th>
 				<td>@Model.Submission.Branch</td>
 				<th>@Html.DisplayNameFor(m => m.Submission.RerecordCount)</th>
-				<td>@Model.Submission.RerecordCount</td>
+				<td>@(Model.Submission.RerecordCount > 0
```
adelikat, line 161: are we sure this is the only place we show a rerecord count on a submission?

TASVideos.Core/Services/QueueService.cs
vadosnaprimer: Warnings string was created upon submission creation, and then persisted on updates. I'm clearing it now so the newly parsed movie is the only one involved in warning creation.

TASVideos/Pages/Submissions/View.cshtml
vadosnaprimer: I think so

adelikat (approved):

Issue #1495, opened 2023-01-01
Code sample download button has black text on dark BG in light mode


vadosnaprimer:
https://tasvideos.org/Forum/Topics/23896

![изображение](https://user-images.githubusercontent.com/7092625/210164399-1427da03-9782-4c4d-9161-c0d00a3af2a7.png)

The text is super hard to see. No such problem in dark mode since the text is white there,

vadosnaprimer:
Similar issue with the Referrers button on game edit page

https://tasvideos.org/Games/Edit/2584?returnUrl=%2F2584G

![изображение](https://user-images.githubusercontent.com/7092625/210164462-51ea450e-73b5-4f8e-bb5d-00d2f5cbccd4.png)

PR #1589, opened 2023-06-29
fix color of code download button

vadosnaprimer:
see #1495

no idea how to fix the Referrers button text color


adelikat:
before/after pics?

vadosnaprimer:
The only thing different is text color became white.

vadosnaprimer:
before

изображение 

after

изображение 

dark mode hasn't changed

изображение

adelikat (approved):

PR #1591, opened 2023-07-01
clearer indication of what the submit button will do

vadosnaprimer:
- "create" for creating a new thing
- "save" for saving your edits to an existing thing ("update" could be confused with "refresh", we want a term that couldn't)
- "send" when you're sending a message ("save" is an entirely different function for messages, it puts them to the savebox)

### DIFF
--- a/TASVideos/Pages/Flags/Create.cshtml
+++ b/TASVideos/Pages/Flags/Create.cshtml
@@ -37,7 +37,7 @@
 		</column>
 	</row>
 	<div class="text-center mt-3">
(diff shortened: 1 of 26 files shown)


Masterjun3:
I wasn't quite sure at first, but after taking a look I actually like the changes and their reasoning.

The one thing that still irks me is the [+Post] button in the topic view, that actually only brings you to the post screen, where then there's a [+Post] button again. (This is already the case, not actually affected by this PR.)
image

But honestly maybe it's not that bad as I make it out to be.

vadosnaprimer:
Oh that's confusing indeed.

Masterjun3 (approved):
Yes!

Issue #268, opened 2019-12-19
Auto-update referrers when the page gets moved


vadosnaprimer:
Now it has to be done manually each time (thankfully we even have referrer lists), but it makes a lot of sense to update them automatically to point to the new page location.

adelikat:
To be clear, you want it to update those wiki page's markup to have the new page name? That could be an expensive operation.

vadosnaprimer:
I don't think it matters too much how the link is formatted, as long as it's updated (for example, TVA knows how to turn full wiki links into short ones).

YoshiRulz:
IIRC MediaWiki does this by using an internal ID for pages for which the path acts as a pointer

Randomno:
Doesn't MediaWiki just use page redirects?

nattthebear:
This is sort of blocked on https://github.com/TASVideos/tasvideos/issues/219.  Doing this in a sane way would require a system that can both parse to ast and emit wiki syntax.

vadosnaprimer:
It would be incredibly helpful to have this before we start heavily reorganizing the wiki pages. If something is linked to a lot and you have to move it, the routine of updating every referrer will get overwhelming very quickly.

MemoryTAS:
Leaving Redirects like in mediawiki behind when moving would probably be optimal, lots of forum posts will no doubt be broken if I move say Judge Guidelines.

https://www.mediawiki.org/wiki/Help:Redirects
https://www.mediawiki.org/wiki/Manual:Redirects

nattthebear:
We could auto update forum posts as well 🤔

Randomno:
Any update on this?

Bare minimum needed is a redirect module that, well, redirects.

Other desired but unnecessary features:

* Notice of redirection
![image](https://github.com/TASVideos/tasvideos/assets/17330834/e5baef66-b2ed-4247-807d-a0f21cfb03d6)

* Way to view the page without redirection like https://en.wikipedia.org/w/index.php?title=UK&redirect=no

* Automatic creation of a redirect when moving a page. Without it, the former page will need to be re-created and have the redirect module added, but that's still far less effort than changing every existing link.

PR #1593, opened 2023-07-02
Redirect module

Randomno:
#268 

Technically it works but it's garbage. I couldn't even figure out how to get the base URL and redirect to the specified path rather than a whole website. That's the only change that would be necessary to be merged (because it'll change how the module is used), but the whole thing should be redone.

Current usage: `[module:redirect|page=https://tasvideos.org/SandBox]`
Desired usage: `[module:redirect|page=SandBox]`

### DIFF
--- a/TASVideos.WikiEngine/WikiModules.cs
+++ b/TASVideos.WikiEngine/WikiModules.cs
@@ -37,6 +37,7 @@ public static class WikiModules
 	public const string PlayerPointsTable = "playerpointstable";
 	public const string PublicationHistory = "publicationhistory";
 	public const string PublicationPoints = "publicationpoints";
+	public const string Redirect = "redirect";
 	public const string Screenshots = "screenshots";
 	public const string SupportedMovieTypes = "supportedmovietypes";
 	public const string TabularMovieList = "tabularmovielist";
--- a/TASVideos/ViewComponents/Redirect.cs
+++ b/TASVideos/ViewComponents/Redirect.cs
@@ -10,7 +10,16 @@ public class Redirect : ViewComponent
 	{
 		HttpContext context = ViewContext.HttpContext;
 
-		context.Response.Redirect("/" + page);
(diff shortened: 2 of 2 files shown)


adelikat:
To get teh baseurl, add AppSettings to the constructor of your class, then it will be in the BaseUrl property of that class

Randomno:
Works as intended now. Still very minimal implementation and could do with the other features mentioned in https://github.com/TASVideos/tasvideos/issues/268#issuecomment-1616599787 in the future. 2 I can do, others I have no idea.

Masterjun3:
How are infinite redirects prevented?

Randomno:
By the browser from what I tested

image

Masterjun3:
I know we would avoid putting in loops, so it'd be rare, but I don't know if I'm comfortable trusting all browser behaviors to properly catch stuff like that. The problem is that it'd be a massive resource drain if any one doesn't stop it.

Btw, the Preview button gives an error alert if you redirect to a page that doesn't exist. Which is probably fine, it's just interesting that the 404 that comes from the PageNotFound goes directly into the JS fetch Preview request, and so it thinks something didn't work.

Randomno:
Valid concern and I don't know how to address it. Check the contents of the target page and refuse to redirect if it also contains a redirect? Add something to the target URL like `?redirected=yes` and ignore the next redirect if the URL contains it?

Randomno:
Fixed repeated redirects.

As for the implementation, I'm not a big fan of it for the same reason you gave. I'm basing it on the MediaWiki syntax of

`#REDIRECT [[Page]]`

This issue has been around for a long time now and it's [holding back wiki reorganisation](https://tasvideos.org/Forum/Topics/23197?CurrentPage=1&Highlight=513984#513984).

adelikat:
Closing because this has not seen any progress in awhile, and may not be the direction we want to go to handle redirects

PR #1601, opened 2023-07-10
parsers: add GBAHawk .gbmv support

RetroEdit:
(.gbmv is currently just renamed .bk2)


Masterjun3:
I know the gbmv is just a renamed bk2, but I think we should still have [sample files](https://github.com/TASVideos/tasvideos/tree/60e5d649390d718b8a90ec206703d9b7d94eb67d/tests/TASVideos.MovieParsers.Tests) for new file formats.

RetroEdit:
I'll add that shortly. I personally don't think it makes sense to duplicate tests from bk2 currently, but there are one or two tests I can do for future-proofing.

Masterjun3:
I think if an emulator fork is different enough that it can't be merged into the main emulator project, then I'd say it's different enough to get its own independent parser, test code, and test files.
But I'm not blocking this PR.

RetroEdit:
Okay, that testcase is not properly implemented, I don't think.

adelikat:
@RetroEdit : the sample files you added, have to be added as embedded resources to the test project

RetroEdit:
@adelikat I can do that, but to test it I would have to wait to build it. Currently, I'm waiting for a few dependencies to install and still have to configure the build.

PR #1603, opened 2023-07-12
Add title text to publication class image

Randomno:
It's currently not very clear what the note block image on a publication means, this is intended to help with that


vadosnaprimer:
Before/after comparison?

Randomno:
After:

image

vadosnaprimer (approved):

PR #1604, opened 2023-07-13
send discord updates to different channels

### LINKED ISSUE #731: Bot notification types and Discord
Right now all site activity goes into one place (updates), and every time someone replaces encodes for a publication, the channel gets spammed.

It would be very _very_ **very** useful if we could send wiki edits to one channel, publication edits to another channel, forum edits to yet another, etc. That way you don't have to skim through a ton of info you don't need to see, or you can hide specific notifications entirely.

Is there a list of currently available types of what kind of event is being reported by the bot?

### LINKED ISSUE #1267: Separate update feeds to Discord for wiki edits/posts/submissions
Currently all sorts of (public) updates are sent to one `#updates` channel in Discord. This can cause clogging issues with certain types of updates (see #1266)

It would be nicer if these were split into different channels, to make it easier to follow different types of updates. I can imagine many people would be interested in, say, new submissions, but not wiki edits or other such news.

I suggest the following category divisions:

* Forum updates (new posts, new topics, post edits)
* Submission updates (new submissions, submission status updates, new publications)
* Wiki/site updates (wiki edits, publication updates, catalog updates, everything else that's not one of the above)

(contact me on Discord if/when I need to create new channels for this)

vadosnaprimer:
* public
  * updates - forum/wiki
  * tas-updates - pubs/subs/userfiles
  * game-updates - game updates
* hidden
  * hidden-updates - staff forum, hidden userfiles
  * user-updates - user updates

fixes #731
fixes #1267

One thing we need to discuss is whether we want all catalog updates to be considered `PostGroups.Game`. I think it makes sense to do it since cataloging doesn't **change** movies per se, but it does assign game info which is usually created right before that and just for that. And cataloging is something we need to encourage, so there's gonna be a lot of it, and it would spam the movies channel making it harder to find **changes** to movies. Minor edit could be used for cataloging but then it becomes harder to track.

### DIFF
--- a/TASVideos.Core/Services/ExternalMediaPublisher/Distributors/DiscordDistributor.cs
+++ b/TASVideos.Core/Services/ExternalMediaPublisher/Distributors/DiscordDistributor.cs
@@ -34,9 +34,36 @@ public sealed class DiscordDistributor : IPostDistributor
 
 		var messageContent = new DiscordMessage(post).ToStringContent();
 
-		string channel = post.Type == PostType.Administrative
-			? _settings.PrivateChannelId
-			: _settings.PublicChannelId;
+		string channel;
+
+		if (post.Type == PostType.Administrative)
+		{
+			if (post.Group == PostGroups.UserManagement)
+			{
+				channel = _settings.PrivateUserChannelId;
+			}
+			else
+			{
+				channel = _settings.PrivateChannelId;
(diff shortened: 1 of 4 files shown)


adelikat:
This approach looks good 👍

adelikat (approved):

PR #1607, opened 2023-07-14
Add id to moviefilelinks on movie module

### LINKED ISSUE #1605: Verification movie link broken on movie list
https://tasvideos.org/Movies-fds-SortU

This link only works on the publication page.

image

More generally I think this is any additional links of the type MovieFile, but I haven't checked.

Randomno:
Should fix #1605.


adelikat (approved):
Approving but I think it is best to also filter to the Id in the additional download query, but I will add that

PR #1608, opened 2023-07-14
Record page moves in page history

### LINKED ISSUE #1597: Page moves aren't recorded in page history
The old site seemed to record this. It would be helpful to keep track of.

image

Randomno:
Fixes #1597. Tested but not heavily. I don't know what the Markup attribute does.


Randomno:
Making edit minor or else it would send an announcement twice

Masterjun3:
Looking at this, it's somewhat important to note that because this is not done inside the WikiPage Move method, it won't create a new history entry for everything using MoveAll, which is currently only used for renaming people (for moving their Homepage).

adelikat:
This approach doesn't quite work and we have seen no progress in awhile, closing

Issue #1207, opened 2022-03-29
Allow referencing posts in quotes


scrimpeh:
Very minor thing but it'd be neat.

I would like it if quotes can optionally embed a link to the post they're from. If you hit the "Quote Post" button, it automatically inserts a reference to the post in the BBCode.

This could involve allowing quotes to refer to post IDs like `[quote=12345]` (will cause trouble since there exist users with purely numeric usernames), or creating a new BBCode tag for the purpose.

As for presentation of the quote, you can already get something pretty close to what I have in mind by doing this beautiful hack:
`[quote="[post=12345]user[/post]"]` which highlights the author name in the "user wrote" text and makes it link to the post. Maybe it'd be confusing if clicking that doesn't send you to their profile but the post though. Perhaps a small icon next to it could be used instead.

nattthebear:
I'm not sure how the details would all work out, but I like this idea.

Randomno:
The hacky way is very straightforward. Not sure about the proper way. I think the syntax `[quote="author" post="12345"] text [/quote]` would make the most sense, but it seems to be unlike any other forum syntax, so might not be very easy to implement (as least for me).

vadosnaprimer:
I don't see a problem with making the Quote button simply do
```
[quote="[post=12345]user[/post]"]
text
[/quote]
```
instead of just
```
[quote="user"]
text
[/quote]
```

Randomno:
For the same reason scrimpeh mentioned

> Maybe it'd be confusing if clicking that doesn't send you to their profile but the post though. Perhaps a small icon next to it could be used instead.

![image](https://github.com/TASVideos/tasvideos/assets/17330834/ced2d7d4-3cd3-4bac-b8a9-e062a45b24c8)

However, this would probably be fine:

![image](https://github.com/TASVideos/tasvideos/assets/17330834/62a4e963-bd33-409a-aa06-3bcef5b04d73)

PR #1613, opened 2023-07-14
Add quote link to posts

Randomno:
#1207 

image

Very simple BBCode implementation. Should probably be sure about how we want to do this before going ahead with it.

### DIFF
--- a/TASVideos/Pages/Forum/Posts/Create.cshtml.cs
+++ b/TASVideos/Pages/Forum/Posts/Create.cshtml.cs
@@ -87,7 +87,7 @@ public class CreateModel : BaseForumModel
 
 			if (qPost != null)
 			{
-				Post.Text = $"[quote=\"{qPost.Poster!.UserName}\"]{qPost.Text}[/quote]";
+				Post.Text = $"[quote=\"[post={QuoteId}]⇈[/post] {qPost.Poster!.UserName}\"]{qPost.Text}[/quote]";
 			}
 		}
 


nattthebear:
This seems reasonable.  Ideally, this sort of markup wouldn't be baked into the post, but that ship sailed long before the current site, as the old site had the " wrote" text burned into posts as well.

Could we take another pass at the visuals though?  I feel like that single arrows are more commonly used for this sort of thing than double.

vadosnaprimer:
Single arrow might look nicer but it should also be something easy to click, so I suggested wide arrows.

nattthebear:
Is there a fat arrow we can use as opposed to double arrow?

Masterjun3:
For me an up arrow actually implies that the page will be moved (like our "move up" button on the bottom right if you scroll down), where instead it's a link to a completely new page.
Maybe a [speech bubble icon](https://fontawesome.com/icons/comment?f=classic&s=solid) would work here, similar to how Discord does it:
image

vadosnaprimer:
The bubble was very first suggestion, but maybe the post icon would work since it'd be in a different place.

adelikat:
How about this one? https://fontawesome.com/icons/comment?f=classic&s=regular

vadosnaprimer:
Not bad.

nattthebear:
As proposed in this PR, it has to be a plain unicode character to be rendered in posts.  There's nowhere to put a font awesome.  There are other ways it could be implemented that would allow for that, but someone would have to do that first and it would be more complex and maybe not worth it.

adelikat:
Note that this method will only apply to future posts created after this is deployed

adelikat:
Thanks for the PR, but rather than merge I took the idea and did the actual icon in css instead of a baked in unicode character, and also changed it to a comment bubble as mentioned above

PR #1618, opened 2023-07-19
Disable ligatures in wiki page editor

YoshiRulz:


Masterjun3 (approved):

YoshiRulz:
Before and after:
screencap

Masterjun3:
Maybe you'd want to disable the same thing in the wiki diff view?
https://github.com/TASVideos/tasvideos/blob/c5fd546e6984441c30bc99411f8584e04f4ed9d5/TASVideos/wwwroot/css/diff.scss#L11

nattthebear (approved):
This makes sense to me.  That being said, I don't understand why anyone would ever actively want ligatures on a monospace coding font, so I might not be the right target audience.

PR #1619, opened 2023-07-20
subs/pubs: clarify Authors labels

RetroEdit:
Just text-changes.

(There is a separate issue of similar text and schemas being in three separate places).


adelikat:
Rather than address the s vs (s) stuff, I made a commit that simply changes additional to external on the display name to help clarify

PR #1622, opened 2023-07-24
Twitter -> X, this is exactly the type of pull requests we should be using our resources on

Masterjun3:
Resolves #nothing


nattthebear:
For the best clarity for our developers, we should keep all internal and code names as Twitter because that's the only one anyone will know anything about.  We should only rename to X anywhere it's mentioned in actual UI text.

adelikat:
This PR violates common best practices with variable naming, please use a more descriptive name
https://workat.tech/machine-coding/tutorial/writing-meaningful-variable-names-clean-code-za4m83tiesy0

In all seriousness, I intend to remove this integration, so renaming it is no longer needed

PR #1625, opened 2023-07-28
Change Published Authors obsolete link to obsolete only

Randomno:
https://tasvideos.org/Publications/Authors

Makes the obsolete movie list links here use obsonly.


adelikat:
Changes the link but doesn't update the count that is the count of all their pubs

Masterjun3:
Nah this PR is correct. Both links now link to the proper page where the number of movies match what is shown in the table.

Masterjun3 (approved):

PR #1628, opened 2023-07-30
Update to Bootstrap v5.3 with native dark mode support

### LINKED ISSUE #1395: Upgrade to Bootstrap 5.2
Bootstrap 5.2 converted more things to using CSS variables so upgrading should make our codebase cleaner with fewer overrides. It shouldn't be too complicated of a migration otherwise as there don't seem to be any removals of classes. There's a new text-bg-* shorthand that should make it easier to ensure text contrasts well when setting a background which will probably be useful in a lot of places.

Masterjun3:
Resolves #1395 .
This upgrades bootstrap from v5.1.3 to v5.3.1 .

We now include the scss part of bootstrap in our source code and build it ourselves. The js part is still taken from the cdn.
The reason for this is that bootstrap now has native support for dark mode (or custom color modes in general).
To make best use of this, we bundle bootstrap ourselves and inject a few custom colors we need. Bootstrap will then generate a whole palette of colors for us to use.

Some colors are changed, most stayed the same, and some additional color bugs were easy to fix now.

image

image

image

image

image


adelikat (approved):

vadosnaprimer (approved):
For the record, I think the now-previous colors of this plaque were better in terms of site feel

изображение
изображение
compared to this PR's
изображение
изображение
which feels dull.

But otherwise...

PR #1637, opened 2023-08-08
Add link to publication to submission threads where applicable

YoshiRulz:
Untested.


vadosnaprimer:
How does it look now?

YoshiRulz:
No idea, I don't have a dev end set up.

adelikat:
The approach almost works but it is not wise to pull the entire submission which includes the file data, but we can just project to an anonymous object for the two values we need.  Since I want to see this tested, I'll just adapt your changes, test  and push, thanks for the PR

PR #1646, opened 2023-08-16
Surface TOS + privacy policy better

YoshiRulz:
Untested.

I'd also like to suggest these changes to `/SiteRules`:
```diff
+As the stewards of this platform, staff and contributors also have rules to follow:
+the [Staff/Conduct|staff code of conduct], the [System/PrivacyPolicy|Privacy Policy], and the [https://github.com/TASVideos/tasvideos/blob/main/LICENSE|server software's license].
 
 !! Legal issues
 
 Users are __strictly prohibited from posting references to copyrighted material (such as ROM or BIOS images) or asking for such__. Any such posting will be deleted.
-
 Screenshots and AVI recordings are considered fair use and are fully allowed.
+
+To upload content such as movies and scripts, users will need to release it under [SiteLicense|CC BY 2.0].
```

And this change to `/System/PrivacyPolicy`:
```diff
-TASVideos does not share data with any third parties.
+The server does not transfer user data to any third-parties.
+
+Your user agent (= browser) should automatically request some necessary resources from Cloudflare's cdnjs ([https://www.cloudflare.com/privacypolicy|privacy policy]) and jsDelivr ([https://www.jsdelivr.com/terms/privacy-policy-jsdelivr-net|privacy policy]).
+On the registration and login forms, it should also request and run code from Google as part of their ''reCAPTCHA'' spam prevention service ([https://policies.google.com/privacy|privacy policy], FWIW). For the time being, circumventing this will prevent login.
+
+Additionally, some pages may contain embedded images and videos from various third-party servers, which your user agent should request automatically.
```

### DIFF
--- a/TASVideos/Pages/Account/Register.cshtml
+++ b/TASVideos/Pages/Account/Register.cshtml
@@ -49,6 +49,10 @@
 				</div>
 				<span asp-validation-for="Coppa" class="text-danger"></span>
 			</form-group>
+			<form-group>
+				Also make sure to read the <a href="/SiteRules">Site Rules</a> if you haven't already.<br/>
+				Your data will be stored according to our <a href="/System/PrivacyPolicy">Privacy Policy</a>. Remember posts can be deleted, but not un-sent.
+			</form-group>
 			<environment exclude="Development">
 				<form-group>
 					<recaptcha />
--- a/TASVideos/Pages/Account/Register.cshtml
+++ b/TASVideos/Pages/Account/Register.cshtml
@@ -51,7 +51,7 @@
 			</form-group>
 			<form-group>
 				Also make sure to read the <a href="/SiteRules">Site Rules</a> if you haven't already.<br/>
-				Your data will be stored according to our <a href="/System/PrivacyPolicy">Privacy Policy</a>. Remember posts can be deleted, but not un-sent.
+				Your data will be stored according to our <a href="/System/PrivacyPolicy">Privacy Policy</a>. Always be mindful of what you share.
 			</form-group>
 			<environment exclude="Development">
 				<form-group>
--- a/TASVideos/Pages/Account/Register.cshtml
+++ b/TASVideos/Pages/Account/Register.cshtml
@@ -51,7 +51,7 @@
 			</form-group>
 			<form-group>
 				Also make sure you've read the <a href="/SiteRules">Site Rules</a>.<br/>
-				Your data will be stored according to our <a href="/System/PrivacyPolicy">Privacy Policy</a>. Always be mindful of what you share.
+				Your data will be stored according to our <a href="/System/PrivacyPolicy">Privacy Policy</a>.
 			</form-group>
 			<environment exclude="Development">
 				<form-group>
--- a/TASVideos/Pages/Account/Register.cshtml
+++ b/TASVideos/Pages/Account/Register.cshtml
@@ -50,7 +50,7 @@
 				<span asp-validation-for="Coppa" class="text-danger"></span>
 			</form-group>
 			<form-group>
-				Also make sure to read the <a href="/SiteRules">Site Rules</a> if you haven't already.<br/>
+				Also make sure you've read the <a href="/SiteRules">Site Rules</a>.<br/>
 				Your data will be stored according to our <a href="/System/PrivacyPolicy">Privacy Policy</a>. Always be mindful of what you share.
 			</form-group>
 			<environment exclude="Development">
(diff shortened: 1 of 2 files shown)


Masterjun3:
I feel like adjustments to terms and privacy policies should be done carefully and deliberately.
This is one of those thing where I would want a precise list of what we want. Stuff like linking to our rules on every page, or making the privacy policy available on a non-system page.

I'm also not really happy with some of the changes requested here. Like using weird terms like "stewards" (I had to look it up), and things like "posts can be deleted, but not un-sent", what does that even mean? I'm also not a fan of using "FWIW" (which I _also_ had to look up) and all other kinds of unexplained abbreviations.
I would also like an explanation of why things were added, like, why do you suggest adding the information about the captcha about "For the time being, circumventing this will prevent login".

Like, the title here says "privacy policy better", but I don't know about that...

YoshiRulz:
> I feel like adjustments to terms and privacy policies should be done carefully and deliberately. [...] I'm also not really happy with some of the changes requested here.

Agreed. These are only suggestions, and I don't doubt the wording can be improved for the benefit of non-native speakers.

> "posts can be deleted, but not un-sent", what does that even mean?

"Be mindful of what you share." But now I see that's a shorter way of putting it, which is what I was going for...

> I'm also not a fan of using "FWIW" (which I also had to look up) and all other kinds of unexplained abbreviations.

Fair, I was just full of anti-Google sentiment at that point (hence #1647). Where else did I abbreviate?

> why do you suggest adding the information about the captcha about "For the time being, circumventing this will prevent login".

It's possible to [pi-hole](https://en.wikipedia.org/wiki/Pi-hole#Difference_from_traditional_advertisement_blockers) Google (and Cloudflare but [that doesn't break anything](https://www.localcdn.org/)), but until #1647 is resolved, those people won't be able to create an account or log in. As it stands, it's not obvious what the cause is or whether that's intentional. (We can't assume every pi-hole user's first thought would be to check for blocked requests.)

> Like, the title here says "privacy policy better", but I don't know about that...

(Surface (TOS + privacy policy)) better. I am of the opinion that the current privacy policy has a lot of problems (for one, *where* is user data being stored or processed?), but that's outside the scope of this PR and not really my call to make.

vadosnaprimer:
Suggestions don't explain much without providing all the relevant content and reasons, ideally in a way that people who don't know anything about server technology could understand. And yes it should be a forum thread.

vadosnaprimer:
No discussion of those topics on the site suggests me we don't need to leave this PR open indefinitely. If there's a discussion, and an agreement, the agreed on changes can be suggested separately.

TASVideos/Pages/Account/Register.cshtml
```diff
@@ -49,6 +49,10 @@
 				</div>
 				<span asp-validation-for="Coppa" class="text-danger"></span>
 			</form-group>
+			<form-group>
+				Also make sure to read the <a href="/SiteRules">Site Rules</a> if you haven't already.<br/>
```
vadosnaprimer, line 53: I think "if you haven't already" is unnecessary. "Make sure" already implies it.

YoshiRulz:
```suggestion
				Also make sure you've read the <a href="/SiteRules">Site Rules</a>.<br/>
```

TASVideos/Pages/Account/Register.cshtml
```diff
@@ -49,6 +49,10 @@
 				</div>
 				<span asp-validation-for="Coppa" class="text-danger"></span>
 			</form-group>
+			<form-group>
+				Also make sure you've read the <a href="/SiteRules">Site Rules</a>.<br/>
+				Your data will be stored according to our <a href="/System/PrivacyPolicy">Privacy Policy</a>. Always be mindful of what you share.
```
vadosnaprimer, line 54: Also unsure about "Always be mindful of what you share." If we link privacy policy in other places it kinda make sense to say it there too, but can't we say it directly on the privacy policy page instead?

vadosnaprimer, line 54:
```suggestion
				Your data will be stored according to our <a href="/System/PrivacyPolicy">Privacy Policy</a>.
```

vadosnaprimer (approved):
Code changes look ok IMO. The rest is a site talk.

TASVideos/Pages/Account/Register.cshtml
```diff
@@ -49,6 +49,10 @@
 				</div>
 				<span asp-validation-for="Coppa" class="text-danger"></span>
 			</form-group>
+			<form-group>
+				Also make sure you've read the <a href="/SiteRules">Site Rules</a>.<br/>
+				Your data will be stored according to our <a href="/System/PrivacyPolicy">Privacy Policy</a>.
```
Masterjun3, line 54: Isn't it weird to link to a System page that clearly says "not a page intended for standalone viewing"?

YoshiRulz: I've just copied the login page:
https://github.com/TASVideos/tasvideos/blob/ba987957384a1087e00159f9a6a8865b6789da75/TASVideos/Pages/Account/Login.cshtml#L34
I agree the warning could confuse new users.

PR #1665, opened 2023-08-27
Make game version optional in submission form

Randomno:
More debatable than the other PR so making it separate


adelikat:
The database allows nulls, so this change will work.  I'll leave it up to staff, if we want to change this to be optional.  If someone approves the idea, I'll merge

vadosnaprimer (approved):

PR #1682, opened 2023-09-02
Support newer OMR format

### LINKED ISSUE #1586: System detection for openMSX movie files (.omr) post-18.0 is broken
The format was changed so the system is now under snapshots>item>config>config>msxconfig>info>type. As far as I could tell, nothing else has changed, though I'm not sure why it throws an error trying to parse 18.0 files instead of defaulting to MSX system.

18.0 example movie (renamed to .gz):

[openmsx0038.gz](https://github.com/TASVideos/tasvideos/files/11909612/openmsx0038.gz)

Relevant page: https://tasvideos.org/EmulatorResources/OpenMsx/OMR

fsvgm777:
Fixes #1586.

Also fixes a rounding issue when calculating the total frame count, as it'd result in the total run time being wrong most of the time.

Included are also additional system detection tests to account for the recent format changes.

I've tested it locally, and it works fine.


fsvgm777:
After a brief discussion, I am not sure whether the rounding should be changed back (Math.Ceiling) or if the rounding should be kept (with the midpoint rounding changed to AwayFromZero to fix more rounding issues). As the format doesn't exactly have a concept of a *frame* per se (in fact, from my understanding of the format, it's cycle-based, which can then be converted to seconds), I wouldn't even be opposed to put in a frame rate override. Opinions?

MBilderbeek:
I'd approve this if I had the power 🙂

adelikat (approved):

PR #1683, opened 2023-09-03
API summary wording

Randomno:


adelikat (commented):
I believe the comma is correct grammar

adelikat (approved):

PR #1689, opened 2023-09-06
Restore PageTitle on Submission Page

Masterjun3:
This restores the PageTitle section that was removed in a9bf6069d402f5e8d4b2888030aca2e8a6d5f913 .
I don't know why it was removed.


Masterjun3:
Ah, this breaks the implementation of that commit. Closing because we need another approach.

PR #1691, opened 2023-09-06
Only let users unassign roles they are allowed to assign

### LINKED ISSUE #1690: Role editing grants unexpected privileges
As a Senior Moderator, I'm able to remove _any_ existing role, including the Site Admin role, by either unchecking it or pressing the "x" next to it. Doesn't matter if I don't have permissions to _set_ it in the first place.

example

Masterjun3:
Resolves #1690 .
It also supports neat disabled attributes for the client view.
image


vadosnaprimer:
I suggest making text for disabled list items gray, like on disabled buttons.

Masterjun3:
image

vadosnaprimer:
Perfect. Now I can see if I'm able to set the cleared ones or not.

PR #1729, opened 2024-01-04
fix module:GameName on general GameResources subpages

### LINKED ISSUE #1688: Game Resource wiki subpages have wrong/empty disclaimer
On e.g. https://tasvideos.org/GameResources/SNES you have the disclaimer: 
> This page documents information about Super Nintendo Entertainment System. Many of the tricks demonstrated here are near impossible in real time and documented for the purposes of creating Tool-assisted Speedruns.

This is correct.

However, on pages like https://tasvideos.org/GameResources/CommonTricks or https://tasvideos.org/GameResources/BossFightingGuide you have:
> This page documents information about . Many of the tricks demonstrated here are near impossible in real time and documented for the purposes of creating Tool-assisted Speedruns.

image

vadosnaprimer:
Somewhat hacky? The idea is to reword it to "[system] **games**" when there's a system, because those pages don't actually have any info about the system itself, only about its games. And when there's no system to display, that means we're on a general subpage, so we say "**various** games" instead.

fix #1688

### DIFF
--- a/TASVideos/Pages/Shared/Components/GameName/Default.cshtml
+++ b/TASVideos/Pages/Shared/Components/GameName/Default.cshtml
@@ -1,7 +1,7 @@
 ﻿@model IList<GameNameModel>
 @if (Model.Any(g => g.IsSystem))
 {
-	<text><span>@Model.First().System</span></text>
+	<text><span>@Model.First().System games</span></text>
 }
 else
 {
--- a/TASVideos/ViewComponents/GameName.cs
+++ b/TASVideos/ViewComponents/GameName.cs
@@ -23,10 +23,9 @@ public class GameName : ViewComponent
 		{
 			var system = await _db.GameSystems
 				.SingleOrDefaultAsync(s => s.Code == path.SystemGameResourcePath());
-			if (system is not null)
-			{
-				gameList.Add(new GameNameModel { System = system.DisplayName });
-			}
+			gameList.Add(new GameNameModel { System = system is not null
+				? system.DisplayName
+				: "various" });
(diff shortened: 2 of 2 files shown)


TASVideos/ViewComponents/GameName.cs
```diff
@@ -23,10 +23,9 @@ public async Task<IViewComponentResult> InvokeAsync()
 		{
 			var system = await _db.GameSystems
 				.SingleOrDefaultAsync(s => s.Code == path.SystemGameResourcePath());
-			if (system is not null)
-			{
-				gameList.Add(new GameNameModel { System = system.DisplayName });
-			}
+			gameList.Add(new GameNameModel { System = system is not null
```
adelikat, line 26: Can I get you to cleanup these stylecop warnings?

TASVideos/Pages/Shared/Components/GameName/Default.cshtml
```diff
@@ -1,7 +1,7 @@
 ﻿@model IList<GameNameModel>
 @if (Model.Any(g => g.IsSystem))
 {
-	<text><span>@Model.First().System</span></text>
+	<text><span>@Model.First().System games</span></text>
```
adelikat, line 4: Is this module also the one on system pages like https://tasvideos.org/GameResources/A7800 ? If so, I think this is a breaking change?

vadosnaprimer: On those it says "[system] games".

adelikat (approved):

PR #1736, opened 2024-01-05
rate-limit submissions

### LINKED ISSUE #1709: Implement system for rate-limiting submissions
Due to the increased submission backlog over the recent years, and the pressure it puts on the judging staff, it would be wise to implement a per-user rate limit on submissions. Meaning, if a submitter has submitted _x_ runs within the past _y_ hours, they are blocked from submitting movies until enough time has passed.

Based on a discussion we've recently had, a generally agreed on rate would be permitting up to 3 submissions within 1 week (168 hours). It might be good to make those numbers configurable by an admin somewhere.

A general idea of the implementation would be: upon going to the submission form page, check if the user has any submitted movies from between now and 168 hours ago. If the number of submitted movies is 3 or greater, then grey out or cut out the submission form, and replace it with a message that the user has to wait until (time of first submission within the range period + 168 hours) until they can submit again.

Also note: it may or may not be worthwhile to add an exemption to this limit for April Fools. It's still possible to get by if the restriction is 3 per week, but if the goal is to keep April Fools' Day uninhibited, that might require a special case.

vadosnaprimer:
When you attempt to submit by going from sub list page to SubmitMovie, I replace the entire page with the notice telling how much you need to wait, if you're beyond the limit. The limit is configured in appsettings.

I doubt it's done properly, but should be fixable.

fix #1709

### DIFF
--- a/TASVideos/Pages/Submissions/Submit.cshtml.cs
+++ b/TASVideos/Pages/Submissions/Submit.cshtml.cs
@@ -202,12 +202,9 @@ public class SubmitModel : BasePageModel
 
 	public bool SubmissionAllowed(int userId)
 	{
-		_earliestTimestamp = _db.Submissions
-			.Where(s => s.Submitter != null && s.SubmitterId == userId)
-			.OrderByDescending(s => s.CreateTimestamp)
-			.ToList()
-			.ElementAt(_settings.SubmissionRate.Submissions - 1)
-			.CreateTimestamp;
-		return _earliestTimestamp < DateTime.UtcNow.AddDays(-_settings.SubmissionRate.Days);
+		var subs = _db.Submissions.Where(s => s.Submitter != null && s.SubmitterId == userId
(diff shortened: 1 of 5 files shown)


Masterjun3:
This seems to be only a soft limit, a client limit, right? So for example someone opening 20 tabs of the submission page could still submit them all one after another?

vadosnaprimer:
Woops...

TASVideos/Pages/Submissions/Submit.cshtml.cs
```diff
@@ -183,4 +193,24 @@ private async Task ValidateModel()
...
+		"We limit submissions to " +
+		_settings.SubmissionRate.Submissions +
+		" in " +
+		_settings.SubmissionRate.Days +
+		" days per user. ",
+		"You will be able to submit again on " +
+		_earliestTimestamp.AddDays(_settings.SubmissionRate.Days)
+	};
+
+	public bool SubmissionAllowed(int userId)
```
adelikat, line 209: This needs to be async

adelikat (approved):

Issue #1431, opened 2022-10-02
Rework the concept of branch into goal


vadosnaprimer:
Right now branches are just string attributes of submissions and publications. It's been working fine, until we faced challenges related to Playground, and any kind of a potential leaderboard UI in general.

I suggest to introduce the concept of goals that exist globally. Sometimes they are represented by existing tags, but sometimes they're just too unique, and we use the Demonstration tag for those (like the ["ring attack"](https://tasvideos.org/2593M) goal).

Goals would consist of IDs in the code, and have string names attached to them, and for any publication, submission, and user file, you'd be able to pick a goal from the dropdown, and it would become its branch. If a goal is implicit and doesn't need to be displayed, we still have tags to categorize what has what, but the goal may then not be picked.

The purpose of this concept would be separation of existing branches and connection between them. That way we'll be able to determine which global goals are available for a given game, and make a list of publications or submissions that share it, making the leaderboard idea much more sustainable than via simple string comparison to every other entity.

adelikat:
I like the term goal rather than branch for sure. It is important to point out, what we need is a goal table linked to a game, instead of a column off of the publication.  This way the goal (branch) can even be created ahead of time.  Similar to SDA, we could "open up" different types of goals for particular games.  These could be known and visible even before a submission.

adelikat:
Another question for leaderboard concepts, would it be enough to compare only against frame count? Otherwise we need the time to be de-normalized into a column

vadosnaprimer:
I don't know if other sort criteria will be heavily used, at least time is commonly known as a good default.

MemoryTAS:
imho the exact nature of leaderboards and how they should be implemented requires a bit more discussion. Pure time is a bad sort criteria without implicitly encouraging faster emulators over accurate ones.

vadosnaprimer:
What about date then? I didn't think of enforcing official preference chains in PG, the main goal is just listing those movies together **somehow**. If time can be confusing, date would at least communicate that there's no official precedence.

MemoryTAS:
I extremely recommend holding on before working on this issue, there's some things we need to discuss first

vadosnaprimer:
Is there anything else we need to discuss? Default sorting by date seems to be the best idea to me.

vadosnaprimer:
I wonder if this will remain useful with the playground display being based not on submissions but on semi-publication pages with only the needed data.

vadosnaprimer:
[...]

PR #1742, opened 2024-01-14
Rework the concept of branch into a goal

adelikat:
https://github.com/TASVideos/tasvideos/issues/1431

We now have a GameGoal table instead of a branch column.  Note that publications still have a branch column that is unused, that will be removed in a later commit.

Submissions also still have a branch that the user fills out.  Queue maintainers will formalize it into the proper goal


vadosnaprimer (approved):
No idea bout the code but...

PR #1748, opened 2024-01-21
submissions table: remove table-container

### LINKED ISSUE #1741: Submission table on narrow screen
Unless my browser window is about 1000 pixels wide, the submission list https://tasvideos.org/Subs-List hides one or more columns on the right, and to see them I need to use the Right arrow, or drag the mouse to the right, or scroll to the very bottom of the page to move the scrollbar and then back to see the row I need.

I think it would be much better if the table went beyond the borders like game version tables do https://tasvideos.org/Games/2584/Versions/List That way there would be much more room for the table since bootstrap borders (don't remember the actual name) snap to a few fixed widths instead of being flexible, and not everyone may be aware of arrow and mouse functionality with scrolling.

vadosnaprimer:
it was causing scrollbar to only appear at the bottom of the table rather than the entire page, fixes #1741)

**versions table**:
- put hashes in the same cell
- add bold and text-break so action buttons are there on small screens
- renamed name to file name to indicate what we need to be putting there

before (buttons are missing because logged out)
изображение

after (name tweak is not on the pic)
изображение

### DIFF
--- a/TASVideos/Pages/Games/Index.cshtml
+++ b/TASVideos/Pages/Games/Index.cshtml
@@ -170,7 +170,7 @@
 <table condition="Model.Game.Versions.Any()" class="table table-responsive table-sm">
 	<tr>
 		<th>Type</th>
-		<th>Name</th>
+		<th>File Name</th>
 		<th>Title Override</th>
 		<th>Region</th>
 		<th>Version</th>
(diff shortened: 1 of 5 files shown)


Masterjun3:
This PR is in direct contest of PR #400 which implemented "table-container" in the first place.

With this PR, the page layout now breaks:
image

vadosnaprimer:
Apparently all those things need different solutions.

PR #1684, opened 2023-09-04
Use markdown for URLs in Discord publications

qixils:
Future work could expand on this with more URLs (say, linking to the responsible users) but figured this might be a good start to reduce visual clutter in the Discord update channels. Not sure if this might be considered too platform-specific to embed as a variable in Post? I figure markdown is relatively platform-agnostic so it's not too bad but 🤷🏻‍♀️ 

edit: closes #1681

PR #1749, opened 2024-01-22
Escape formatting symbols for Discord publisher

### LINKED ISSUE #1698: Sanitize user-provided text before posting to Discord
This has been shown a few times: edit messages will be parsed as Discord Markdown and this has implications.

I think it might be good to prevent the following in edit messages:
- [ ] Links with displayed text
- [ ] Pinging via edit messages
- [ ] Clickable links entirely maybe?

The first two could be done by escaping the following six characters with a `\` prefix: `()[]<>`; clickable links could be done by escaping the `/` character.

My rationale here is that we don't necessarily want people pinging or inserting misleading links into edit messages, but maybe opinions differ?

For reference, this is what seems to be the official documentation for Discord Markdown: [Discord Markdown Text 101](https://support.discord.com/hc/en-us/articles/210298617-Markdown-Text-101-Chat-Formatting-Bold-Italic-Underline-)

Edit (2023-10-20): Since then, I've seen some more situations, like a user with a username that formats text with `__`. It might be good to just sanitize everything.

Masterjun3:
Resolves #1698 .
Escapes the formatting symbols Discord uses, so that we don't accidentally format.
(Before | After)
image

Sadly, I can't add ( ) [ ] to the list of escaped characters, due to the implementation of #1684 . It would break our own formatting, and there is no way to differentiate between our own formatting and accidental formatting at that point in the code.

### DIFF
--- a/TASVideos.Core/Services/ExternalMediaPublisher/Distributors/DiscordDistributor.cs
+++ b/TASVideos.Core/Services/ExternalMediaPublisher/Distributors/DiscordDistributor.cs
@@ -2,6 +2,7 @@
 using Microsoft.Extensions.Logging;
 using TASVideos.Core.Settings;
 using TASVideos.Core.HttpClientExtensions;
+using System.Text.RegularExpressions;
 
 namespace TASVideos.Core.Services.ExternalMediaPublisher.Distributors;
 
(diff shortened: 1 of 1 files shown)


vadosnaprimer:
Does it make sense to check if the link in the TVA post is **not** to tasvideos, and to add an external link icon or some other indication to it?

Masterjun3:
What do you mean? Our links always point to TASVideos, don't they?

Masterjun3:
feos clarified on Discord.
It seems he means links that are made by Discord from the user input.

My answer is that this is waay beyond the scope of this PR. First of all, you'd have to somehow recreate how Discord parses links, and you have to parse the `[ ]( )` grammar. Since I use regex here, [that will be difficult](https://stackoverflow.com/a/1732454).

vadosnaprimer (approved):

PR #1750, opened 2024-01-23
Add title to movie links in wiki

### LINKED ISSUE #1349: Wiki: Show tooltip on internal urls
Same functionality as #1347 just for wiki, which is a fair bit more complicated than forum. Old site had this on both forum and wiki.

nattthebear:
This matches behavior in the old site, where a ##M / ##S / ##G link with link text provided would still get the standard text as a title.

```
[1M]             ->       <a href=/1M>[1] Some Really Awesome Movie</a>
[1M|foo]         ->       <a href=/1M title="[1] Some Really Awesome Movie">foo</a>
```

The implementation logic be reasonably straightforward; it just has to process substitutions and replacements in a specific order to get the fallback behavior we need.

fixes #1349

### DIFF
--- a/TASVideos.WikiEngine/MakeBracketed.cs
+++ b/TASVideos.WikiEngine/MakeBracketed.cs
@@ -227,14 +227,14 @@ public static partial class Builtins
 		{
 			if (pp.Length >= 2)
 			{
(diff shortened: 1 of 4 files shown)


vadosnaprimer:
I checked and it's ready now.

vadosnaprimer (approved):

PR #1751, opened 2024-01-24
refinements of goals UI

vadosnaprimer:
- show the baseline goal without quotes and bold but with parentheses, like before goals were implemented
- put baseline goal as the first tab on game pages, and add there link to goals list
- make goal list back button actually go back
- turn goal list into bullet list
- make goal edit buttons small
- rename goal list catalog button from "new" to "manage"

### DIFF
--- a/TASVideos/Pages/Games/Goals/List.cshtml
+++ b/TASVideos/Pages/Games/Goals/List.cshtml
@@ -6,7 +6,7 @@
 }
 
 <h4>Game: <a asp-page="/Games/Index" asp-route-id="@Model.GameId">@Model.GameDisplayName</a>
-	<a condition="!string.IsNullOrWhiteSpace(returnUrl)" asp-page="/Goals/Index" class="btn btn-secondary mb-2 float-end"><span class="fa fa-arrow-left"></span> Back</a>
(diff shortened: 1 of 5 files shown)


TASVideos/Pages/Games/Versions/Edit.cshtml.cs
```diff
@@ -174,7 +174,7 @@ public async Task<IActionResult> OnPost()
 
 		return string.IsNullOrWhiteSpace(HttpContext.Request.ReturnUrl())
 			? RedirectToPage("List", new { gameId = GameId })
-			: BaseReturnUrlRedirect($"?GameId={GameId}&GameVersionId={version.Id}");
```
vadosnaprimer, line 177: if I leave this here it appends the game twice both times with `?` which breaks the selection syntax.

TASVideos/Pages/Games/Versions/Edit.cshtml.cs
```diff
@@ -174,7 +174,7 @@ public async Task<IActionResult> OnPost()
...
+			: BaseReturnUrlRedirect($"&GameVersionId={version.Id}");
```
vadosnaprimer, line 177: this can't be right, because if the game already existed and we selected it without creating, its id won't appear after `?` on catalog page, we now just have `&`, so creating a version redirects to System Error

adelikat (approved):

PR #1758, opened 2024-01-28
allow line wrap for annotations to make them look like in libtas

### LINKED ISSUE #1744: Implement Word Warp in the Submission Annotation Field
#1139 - "please use the non-code formatting tools and feel free open a new issue if the non-code formatting tools are insufficient for formatting things that aren't code"

The [Submission Annotations](https://tasvideos.org/6954S) should probably have word warping considering its small dimensions:
image

Compared to libTAS (since that's what the movie was made on) which actually has word warping this seems like a bit of an oversight on the site end.

vadosnaprimer:
they are not code (aside from a few samples) so people format them as normal text and we don't want to avoid line wrapping, yet we want to avoid parsing them as our wiki

fixed 4 tabs in front of annotations

added `pre` and proper border for userfile annotations card

fix #1744

### DIFF
--- a/TASVideos/Pages/Shared/_UserFileInfo.cshtml
+++ b/TASVideos/Pages/Shared/_UserFileInfo.cshtml
@@ -75,14 +75,16 @@
 		</div>
 	</cardbody>
 </div>
+
 <div condition="!string.IsNullOrWhiteSpace(Model.Annotations)" class="mb-2">
-	<card>
+	<div class="card border-primary-subtle border mb-2">
 		<cardheader><h4>Annotations</h4></cardheader>
 		<cardbody>
-			@Model.Annotations
+			<pre style="white-space: pre-wrap">@Model.Annotations</pre>
 		</cardbody>
-	</card>
+	</div>
 </div>
+
 <div condition="!string.IsNullOrWhiteSpace(Model.ContentPreview)">
 	<wiki-markup markup="@ToMarkup(Model.ContentPreview, Model.Extension)"></wiki-markup>
 </div>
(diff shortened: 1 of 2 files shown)


TASVideos/Pages/Shared/_UserFileInfo.cshtml
```diff
@@ -75,14 +75,16 @@
...
 	</cardbody>
 </div>
+
 <div condition="!string.IsNullOrWhiteSpace(Model.Annotations)" class="mb-2">
-	<card>
+	<div class="card border-primary-subtle border mb-2">
 		<cardheader><h4>Annotations</h4></cardheader>
 		<cardbody>
-			@Model.Annotations
+			<pre style="white-space: pre-wrap">@Model.Annotations</pre>
```
adelikat, line 83: please do not use inline styles.  If bootstrap doens't have a style for this, add a class and put the style in a style tag

TASVideos/Pages/Shared/_UserFileInfo.cshtml
```diff
@@ -75,14 +75,16 @@
 		</div>
 	</cardbody>
 </div>
+
 <div condition="!string.IsNullOrWhiteSpace(Model.Annotations)" class="mb-2">
-	<card>
+	<div class="card border-primary-subtle border mb-2">
```
adelikat, line 80: card creates div class="card", it is better to keep it as card and simply add the other classes

TASVideos/Pages/Shared/_UserFileInfo.cshtml
vadosnaprimer: > add a class and put the style in a style tag

How?

adelikat (approved):

PR #1759, opened 2024-01-29
protect against assigning award to the wrong type

vadosnaprimer:
pass in type when revoking, so in problematic cases (which we now protect against) when an award is assigned to user and movie at once, it can be separately revoked from each

sort 2 more more pages by description

button colors

### DIFF
--- a/TASVideos/Pages/AwardsEditor/Assign.cshtml.cs
+++ b/TASVideos/Pages/AwardsEditor/Assign.cshtml.cs
@@ -50,6 +50,7 @@ public class AssignModel : BasePageModel
 			ModelState.AddModelError("", "Cannot assign both a user and a publication to an award.");
 		}
 
+		// TODO: rework this logic, CategoryExists() has almost the same query
 		var type = await _awards.AwardCategories()
 			.Where(c => c.ShortName == AwardToAssign.Award)
 			.Select(c => c.Type)
--- a/TASVideos/Pages/AwardsEditor/Assign.cshtml.cs
+++ b/TASVideos/Pages/AwardsEditor/Assign.cshtml.cs
@@ -3,6 +3,7 @@ using Microsoft.AspNetCore.Mvc.Rendering;
 using TASVideos.Core.Services;
 using TASVideos.Data;
 using TASVideos.Data.Entity;
+using TASVideos.Data.Entity.Awards;
 
 namespace TASVideos.Pages.AwardsEditor;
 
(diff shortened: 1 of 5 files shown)


TASVideos/Pages/AwardsEditor/Assign.cshtml.cs
```diff
@@ -49,6 +50,21 @@ public async Task<IActionResult> OnPost()
 			ModelState.AddModelError("", "Cannot assign both a user and a publication to an award.");
 		}
 
+		var type = _awards.AwardCategories()
+			.Where(c => c.ShortName == AwardToAssign.Award)
+			.Single()
```
vadosnaprimer, line 55: This was questioned on discord but I dunno how do do it better. Currently it crashes because there's more than one award with the same name, because it allowed assiging the same award to user and movie if it was done separately. The protection was only against doing it in one go.

adelikat: .Single() wil throw an exception if there is 0 ore more than 1.  That will result in a 500 error.  Maybe you want to do a .ToList() and check how many there are and return a ModelState error?

vadosnaprimer: Sorry I misremembered. This one should be fine, because we only have the award category here because it was selected from the dropdown
https://github.com/TASVideos/tasvideos/blob/78bf0969adbefacccfb06fcd1bfefaf9a4318406/TASVideos/Pages/AwardsEditor/Assign.cshtml#L15
which gets populated from the same source
https://github.com/TASVideos/tasvideos/blob/78bf0969adbefacccfb06fcd1bfefaf9a4318406/TASVideos/Pages/AwardsEditor/Assign.cshtml.cs#L107
and we're simply checking what type that category has.

TASVideos/Pages/AwardsEditor/Assign.cshtml.cs
```diff
@@ -90,7 +106,7 @@ public async Task<IActionResult> OnPostRevoke(string shortName)
 	{
 		var awardToRevoke = (await _awards
 			.ForYear(Year))
-			.SingleOrDefault(a => a.ShortName == shortName);
+			.FirstOrDefault(a => a.ShortName == shortName);
```
vadosnaprimer, line 109: Now here's the part that fixes the crash (in those problematic cases we have the same award twice, because it was mistakenly assigned to a user and to a movie), but it got this comment on discord:

> also I don't think you should ever use First/FirstOrDefault if you don't have an ordering. I think EF Core even warns you about that (if executed and you look into the logs). First is "safer" than Single in the sense that it doesn't crash if there's duplicates, but I think you actually want to crash if there's something wrong

TASVideos/Pages/AwardsEditor/Assign.cshtml.cs
vadosnaprimer: So I don't have anything to add here, because `AwardToAssign.Award` is guaranteed to match at least 1 `ShortName`, since it's taken from that list in `Initialize()`. And it can't have several matches unless someone hacks the DB, which seems to have happened in 2021 before we had this UI, and 2 duplicate awards were created, which were removed yesterday.

adelikat (approved):

PR #1768, opened 2024-02-04
Implement text for wikiLink module

### LINKED ISSUE #1760: Wiki Links are not converted into text (e.g. for YouTube descriptions)
The recent PR #1750 fixing issue #1349 implemented movie titles for wiki links like `[1234S]`.

This was done by not parsing it as a simple link anymore, but as a module. Which in turn has the effect that it's now more complicated to convert into text.
In fact, for most modules we simply skip it when converting into text, which is the case here.

So it needs to be implemented.

It seems our way of converting modules to text is by adding the `[TextModule]` attribute to the module class (in this case `ViewComponents/WikiLink.cs`) and adding a `RenderTextAsync` method, which I suppose would look like
`public async Task<string> RenderTextAsync(IWikiPage? pageData, string href, string? displayText)`.

Masterjun3:
Resolves #1760 . (Text missing from Youtube Descriptions)
Also resolves Text missing from TOC, which is a different method.

This implements InnerText of __wikiLinks being href if displaytext is empty. (For TOCs)
It also implements RenderTextAsync by using the same method as InvokeAsync, but only taking the display name. (For Youtube)

**NOTE**: One problem remaining is that on YouTube, the wiki links won't be a link, only text. That's because I can't format it into the `displaytext ( href )` format, as with links, because I don't have the `writerHelper` that is used to convert the href to a full url. Trying to inject it gives `System.InvalidOperationException: Unable to resolve service for type 'TASVideos.WikiEngine.AST.IWriterHelper' while attempting to activate 'TASVideos.ViewComponents.WikiLink'.`

However, a non-link is better than no text at all.

### DIFF
--- a/TASVideos.WikiEngine/NodeImplementations.cs
+++ b/TASVideos.WikiEngine/NodeImplementations.cs
@@ -464,8 +464,16 @@ public class Module : INode
 		// to the end user.  TODO:  __wikiLink really needs to be its own AST type.
 		if (Name == "__wikiLink")
 		{
-			Parameters.TryGetValue("displaytext", out var ret);
-			return ret ?? "";
+			if (Parameters.TryGetValue("displaytext", out var displaytext))
+			{
+				return displaytext;
+			}
+			else if (Parameters.TryGetValue("href", out var href))
+			{
+				return href[1..];
+			}
+
+			return "";
 		}
 
 		return "";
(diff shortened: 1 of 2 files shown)


adelikat (approved):

nattthebear (approved):
This is a good step forward.  Thanks for taking care of this.

PR #1774, opened 2024-02-13
improve goal editing

vadosnaprimer:
- scroll to goal we're editing
- add cancel button
- move buttons to the next row and make them small (buttons on the same row as text field makes the field text barely readable)
- don't pass goalToEdit when hitting Manage button (having to manage means we need a new goal, and only very rarely we need to edit an existing one, which should not be encouraged by UI as a default action)

before

изображение

after

изображение

### DIFF
--- a/TASVideos/Pages/Games/Goals/List.cshtml
+++ b/TASVideos/Pages/Games/Goals/List.cshtml
@@ -22,12 +22,15 @@
 		.ThenBy(g => g.DisplayName))
 	{
 		var canDelete = !goal.Publications.Any() && !goal.Submissions.Any();
-		<tr>
+		<tr id="@goal.Id">
 			<td condition="@(goal.Id == Model.GoalToEdit && goal.DisplayName != "baseline")">
 				<form method="post" class="d-flex gap-2" asp-page-handler="Edit">
-					<input type="hidden" name="gameGoalId" value="@Model.GoalToEdit" />
-					<input type="text" name="newGoalName" class="form-control" value="@goal.DisplayName" />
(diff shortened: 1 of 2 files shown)


TASVideos/Pages/Games/Goals/List.cshtml
```diff
@@ -22,12 +22,15 @@
...
 	{
 		var canDelete = !goal.Publications.Any() && !goal.Submissions.Any();
-		<tr>
+		<tr id="@goal.Id">
 			<td condition="@(goal.Id == Model.GoalToEdit && goal.DisplayName != "baseline")">
 				<form method="post" class="d-flex gap-2" asp-page-handler="Edit">
-					<input type="hidden" name="gameGoalId" value="@Model.GoalToEdit" />
-					<input type="text" name="newGoalName" class="form-control" value="@goal.DisplayName" />
-					<button type="submit" class="btn btn-secondary">Save</button>
+					<column>
```
adelikat, line 28: you don't want to use column without also using a column for row.  you can use <full-row> which will add both of those for you

adelikat (approved):

TASVideos/Pages/Games/Goals/List.cshtml
```diff
@@ -22,12 +22,15 @@
 		.ThenBy(g => g.DisplayName))
 	{
 		var canDelete = !goal.Publications.Any() && !goal.Submissions.Any();
-		<tr>
+		<tr id="@goal.Id">
```
Masterjun3, line 25: HTML id's should start with a letter, not a number.

PR #1777, opened 2024-02-20
Don't require unique Email for dev environments

Masterjun3:
Fixes a bug when launching with sample data, because we set all the emails to the same one.


vadosnaprimer (approved):

Masterjun3:
Closing this, it was instead resolved by making emails actually unique in the sample data in 1bc0f9376294aac75d9747b3c894fb4fd36f0809 .

PR #1781, opened 2024-02-27
set octoshock fps overrides to mednafen values

vadosnaprimer:
game version should not be a required field for submission


vadosnaprimer:
How do I enable invariant culture for an input field? It wants a comma there for me

изображение

vadosnaprimer:
Undrafted since I changed my mind on the thing I wanted to include.

adelikat (approved):

PR #1782, opened 2024-02-28
Prevent select element from flashing right before it's replaced

Masterjun3:
Look at this code, this feels very hacky.

### DIFF
--- a/TASVideos/Pages/Shared/_Layout.cshtml
+++ b/TASVideos/Pages/Shared/_Layout.cshtml
@@ -27,6 +27,7 @@
 	<link rel="alternate" href="/publications.rss" type="application/rss+xml" title="TASVideos RSS Publication Feed">
 	<link rel="alternate" href="/wiki.rss" type="application/rss+xml" title="TASVideos RSS Wiki Feed">
 	<partial name="_MetaTags" />
+ 	<script>document.head.appendChild(document.createElement("style")).innerHTML = '.d-none-js { display: none !important; }';</script>
 </head>
 <body>
 	<nav id="banner" class="navbar navbar-expand-md navbar-dark mb-4">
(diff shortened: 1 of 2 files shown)


TASVideos/Pages/Shared/_Layout.cshtml
```diff
@@ -27,6 +27,7 @@
 	<link rel="alternate" href="/publications.rss" type="application/rss+xml" title="TASVideos RSS Publication Feed">
 	<link rel="alternate" href="/wiki.rss" type="application/rss+xml" title="TASVideos RSS Wiki Feed">
 	<partial name="_MetaTags" />
+ 	<script>document.head.appendChild(document.createElement("style")).innerHTML = '.d-none-js { display: none !important; }';</script>
```
YoshiRulz, line 30: Alternatively:
```suggestion
	<!-- can move to main stylesheet --><style>:root.supports-js .d-none-js { display: none !important; }</style>
	<script>document.documentElement.classList.add("supports-js");</script>
```

Masterjun3: (I can't test right now) Have you tested whether that prevents the flashing on repeated reloading on the usual browsers on desktop and mobile?
It may look the same, but I noticed adding a <style> blocks cached rendering better than adding a css class.

YoshiRulz: I haven't (I also can't test atm)

adelikat:
I'm not sure i want this PR

Masterjun3:
I'm converting this to a draft, I want to try out the suggested thing and some other stuff. If I can't find anything, I'll just close this.

TASVideos/Pages/Shared/_Layout.cshtml
adelikat: they are all a single taghelper now if that helps with a better strategy.  But can we close this? we dont' want this hacky approach

Masterjun3: Yeah I'm going to close this PR. This seems to need a different approach altogether.

Masterjun3:
Closing because we need a different approach altogether.

PR #1784, opened 2024-02-28
implement custom localization

### LINKED ISSUE #908: Customizable locale settings
There has been frequent feedback of people wanting their date format to be different. This currently relies on the locale setting sent by the browser.
However, requiring people to change their entire web browsing experience to show better dates on tasvideos.org is **not** a good solution.
We need a way to set certain locale related settings. And ideally not as a profile setting but a site setting like dark mode.

Side note: As a (poor) workaround, I currently use a manually created `.AspNetCore.Culture` cookie, with a value of `c=kw-gb|uic=kw-gb`, which is a locale using yyyy-MM-dd, a 24 hour clock, and dot separators for decimals.

(Edit: I added some implementation details in the [comment below](https://github.com/TASVideos/tasvideos/issues/908#issuecomment-1969392775))

Masterjun3:
This PR needs to be checked properly.

Resolves #908 .

It implements 3 user settings to customize the date format, the time format, and the decimal number format (see images below).

## Implementation Details
I build a custom CultureInfo, and then, in **every single request** overwrite CultureInfo.CurrentCulture as soon as possible. By using Middleware.

### Middleware Location
Where to place the Middleware? We actually use the `UseRequestLocalization()` middleware to let .NET select the best culture initially. So I thought immediately afterwards would be a good place to place my overwrite middleware. But it turns out I need the User object, so `UseAuthentication` has to run first. That's why I placed it directly after that.

### Caching
Caching is critical. We don't want to read the DB on every request just for the locale. The first idea was to just cache each user's CultureInfo. This alone would work, but there are two Problems:
1. On every restart the cache is cleared, and each user will do one additional DB call to get their CultureInfo build, again and again.
2. Most users don't use custom cultures, so the cache will be filled with non-modified objects.

That's why I decided to use a second cache. This second cache caches the HashSet of the users that actually use custom locales.

The HashSet cache and the user's own CultureInfo cache is cleared when they change their Profile Settings.

### Hot Path
- **For logged out visitors**: The Middleware checks whether they are logged in. They aren't, so it is skipped.
- **For non-custom-locale users**: After confirming the login, it loads the cached HashSet of custom locale users, and they aren't in there. So it is skipped.
- **For custom-locale users**: After loading the cached HashSet and finding the user, it loads the cached CultureInfo and overwrites the CurrentCulture. Then it is done.

<hr>

image

### DIFF
--- a/TASVideos.Common/Constants.cs
+++ b/TASVideos.Common/Constants.cs
@@ -24,6 +24,8 @@ public static class CacheKeys
 	public const string AwardsCache = "AwardsCache";
 	public const string UnreadMessageCount = "UnreadMessageCountCache-";
 	public const string MovieTokens = "MovieTokenData";
+	public const string UsersWithCustomLocale = "UsersWithCustomLocale";
+	public const string CustomUserLocalePrefix = "CustomUserLocale-";
 }
 
 // These perform site functions, maybe they should be in the database?
--- a/TASVideos.Core/Services/UserManager.cs
+++ b/TASVideos.Core/Services/UserManager.cs
@@ -550,4 +550,10 @@ public class UserManager : UserManager<User>
 
 		return true;
 	}
+
+	public void ClearCustomLocaleCache(int userId)
+	{
+		_cache.Remove(CacheKeys.UsersWithCustomLocale);
+		_cache.Remove(CacheKeys.CustomUserLocalePrefix + userId);
+	}
 }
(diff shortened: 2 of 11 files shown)


vadosnaprimer:
I tested with a few language settings in the browser and it works well with obeying and overriding them. I'd only with the settings to be near Time Zone settings in the GUI and not between avatar and signature.

adelikat (approved):

PR #1788, opened 2024-03-07
Run Migration after Sample Data Generation

Masterjun3:
For Sample Data Generation, if the code base is on a later migration than the sample data, loading the data gives an error.
This is because we first migrate the sample db to the latest schema with no data, and then try to add the sample data.

This PR changes the code, so that the DB is created without schema, then the schema+data is added from the sample data, and then finally we migrate if necessary.

This also means the schema needs to be exported with the sample data, and the `__EFMigrationsHistory` table needs to stay.

### DIFF
--- a/TASVideos.Core/Data/DbInitializer.cs
+++ b/TASVideos.Core/Data/DbInitializer.cs
@@ -3,6 +3,8 @@ using Microsoft.Extensions.Options;
 using Npgsql;
 using TASVideos.Core.Settings;
 using SharpCompress.Compressors;
+using Microsoft.EntityFrameworkCore.Storage;
+using Microsoft.EntityFrameworkCore.Infrastructure;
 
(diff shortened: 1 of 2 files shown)


Masterjun3:
@adelikat I tested everything for this PR: having the sample data on a previous migration, or up-to-date, everything works fine.
Still, I do use the `IRelationalDatabaseCreator`, to which [the docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.storage.irelationaldatabasecreator?view=efcore-8.0) say:
> It is generally not used in application code. 

But this is the only way to properly create the DB without the schema. The official `EnsureCreated()` and `Migrate()` methods both use this same `IRelationalDatabaseCreator`, but afterwards they also create a prefilled schema, making the sample data insertion fail.

If this is unwanted, you can close the PR.

PR #1817, opened 2024-04-27
Use NuGet central package management

YoshiRulz:
squashme


adelikat (approved):

adelikat:
I tested with Visual Studio and it was able to figure out to update the props, and figure out which packages are installed when managing both the solution and the individual projects

PR #1818, opened 2024-04-28
Clean up project files

YoshiRulz:
Result of https://github.com/TASVideos/tasvideos/commit/fe207681f58bd8ae4e071594a9962e85d612aef3 (red=removed):
tasvideosdotorg_module_dag

What the other commits do should be clear (and should be *nothing*, that is, everything should function the same).

---

I believe this stuff:
https://github.com/TASVideos/tasvideos/blob/4ab79b97af29ace1be728d990d7a9cb18c977e66/TASVideos/TASVideos.csproj#L69-L77
can be shrunk too, but I wanted to ask whether the tests would flag if I broke it.


tests/TASVideos.Test/Pages/UserFiles/InfoModelTests.cs
```diff
@@ -8,7 +8,7 @@
 namespace TASVideos.RazorPages.Tests.Pages.UserFiles;
 
 [TestClass]
-internal class InfoModelTests
+public sealed class InfoModelTests
```
adelikat, line 11: Why make this one random class sealed?

tests/TASVideos.MovieParsers.Tests/TASVideos.MovieParsers.Tests.csproj
```diff
@@ -18,209 +19,30 @@
...
-		<EmbeddedResource Include="VbmSampleFiles\gb.vbm" />
-		<EmbeddedResource Include="VbmSampleFiles\gbc.vbm" />
-		<EmbeddedResource Include="VbmSampleFiles\savestate.vbm" />
-		<EmbeddedResource Include="VbmSampleFiles\sgb.vbm" />
-		<EmbeddedResource Include="VbmSampleFiles\sram.vbm" />
-		<EmbeddedResource Include="VbmSampleFiles\wrongheader.vbm" />
-		<EmbeddedResource Include="WtfSampleFiles\2frames.wtf" />
-		<EmbeddedResource Include="WtfSampleFiles\noframerate.wtf" />
-		<EmbeddedResource Include="WtfSampleFiles\wrongheader.wtf" />
+		<EmbeddedResource Include="Bk2SampleFiles\*.bk2" />
```
adelikat, line 22: Wow, didn't know you could do that, that makes life so much easier since everyone forgets to add their files to the project in their PRs *thumbs up*

TASVideos.Core/TASVideos.Core.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
+		<RunAnalyzers>false</RunAnalyzers> <!-- this project did not have StyleCop applied; when it was added, there were 66 extra warnings -->
```
adelikat, line 3: WOW! that's one of the most important projects too, adding that to my todo

tests/TASVideos.Test/Pages/UserFiles/InfoModelTests.cs
YoshiRulz: I do that out of habit for Hawk, since it makes reflection slightly faster.
Key here is I'm making it `public` to resolve a warning which is from an Analyzer that was previously not active. I'll just revert all this and leave the warning since it seems that's what's causing the CI build to fail.

tests/TASVideos.Test/Pages/UserFiles/InfoModelTests.cs
```diff
@@ -7,6 +7,7 @@
 
 namespace TASVideos.RazorPages.Tests.Pages.UserFiles;
 
+#pragma warning disable UTA001 //TODO this class should be public, or these shouldn't be tests
```
YoshiRulz, line 10: Ugh this just makes it flag 3 different warnings instead -_-

adelikat (approved):

PR #1831, opened 2024-05-05
Track Wiki Pages Deletes and Moves

### LINKED ISSUE #866: [Request]: Ability to give reasons for deleting
Preferably these reasons would be viewable at deleted pages. Would be nice to reference why x and y were deleted.

adelikat:
resolves #866 

Provides a reason field on the delete modal.  If provided it will create a dummy revision with the reason as the revision edit, it will also put the reason in the external media messages.

Also page moves will create a dummy edit that shows where the page moved from

### DIFF
--- a/TASVideos/Pages/Wiki/Move.cshtml.cs
+++ b/TASVideos/Pages/Wiki/Move.cshtml.cs
@@ -55,6 +55,17 @@ public class MoveModel(IWikiPages wikiPages, ExternalMediaPublisher publisher) :
 
 		var result = await wikiPages.Move(OriginalPageName, DestinationPageName);
 
+		// At a dummy commit to track the move
+		var page = (await wikiPages.Page(DestinationPageName))!;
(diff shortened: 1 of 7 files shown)


TASVideos/Pages/Wiki/Move.cshtml.cs
```diff
@@ -55,6 +55,17 @@ public async Task<IActionResult> OnPost()
 
 		var result = await wikiPages.Move(OriginalPageName, DestinationPageName);
 
+		// At a dummy commit to track the move
+		var page = (await wikiPages.Page(DestinationPageName))!;
+		await wikiPages.Add(new WikiCreateRequest
+		{
+			PageName = DestinationPageName,
+			Markup = page.Markup,
+			RevisionMessage = $"Page Moved from {OriginalPageName}",
```
vadosnaprimer, line 64: For cases when a page is moved several times, maybe also tell where the move is to?

PR #1839, opened 2024-05-07
Set more security-related HTTP headers

YoshiRulz:
With my recent version of FF, I can see that `window.self.crossOriginIsolated` is now `true`, indicating that the COEP is in effect. But what is COEP? To oversimplify: what CORS should have been.

re: `Permissions-Policy`, they all default to `self` (iframes can't use those APIs), but better safe than sorry. If first-party JS actually needs one, its line can be commented out.

### DIFF
--- a/TASVideos/Extensions/ApplicationBuilderExtensions.cs
+++ b/TASVideos/Extensions/ApplicationBuilderExtensions.cs
@@ -42,13 +42,29 @@ public static class ApplicationBuilderExtensions
 
 	public static IApplicationBuilder UseMvcWithOptions(this IApplicationBuilder app, IHostEnvironment env)
 	{
+		var permissionsPolicyValue = string.Join(", ", [
+			"camera=()", // defaults to `self`
+			"display-capture=()", // defaults to `self`
+			"fullscreen=()", // defaults to `self`
+			"geolocation=()", // defaults to `self`
+			"microphone=()", // defaults to `self`
+			"publickey-credentials-get=()", // defaults to `self`
+			"screen-wake-lock=()", // defaults to `self`
+			"web-share=()", // defaults to `self`
+
+			// ...and that's all of the non-experimental options listed on MDN as of 2024-04
+		]);
 		app.Use(async (context, next) =>
 		{
+			context.Response.Headers["Cross-Origin-Embedder-Policy"] = "credentialless";
+			context.Response.Headers["Cross-Origin-Opener-Policy"] = "same-origin";
+			context.Response.Headers["Cross-Origin-Resource-Policy"] = "cross-origin"; // TODO this is as unsecure as before; should be `same-site` or `same-origin` when serving auth-gated responses
+			context.Response.Headers["Permissions-Policy"] = permissionsPolicyValue;
 			context.Response.Headers.XXSSProtection = "1; mode=block";
 			context.Response.Headers.XFrameOptions = "DENY";
 			context.Response.Headers.XContentTypeOptions = "nosniff";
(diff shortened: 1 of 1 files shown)


TASVideos/Extensions/ApplicationBuilderExtensions.cs
```diff
@@ -42,13 +42,29 @@ public static IApplicationBuilder UseStaticFilesWithExtensionMapping(this IAppli
...
+			context.Response.Headers["Cross-Origin-Embedder-Policy"] = "credentialless";
+			context.Response.Headers["Cross-Origin-Opener-Policy"] = "same-origin";
+			context.Response.Headers["Cross-Origin-Resource-Policy"] = "cross-origin"; // TODO this is as unsecure as before; should be `same-site` or `same-origin` when serving auth-gated responses
+			context.Response.Headers["Permissions-Policy"] = permissionsPolicyValue;
 			context.Response.Headers.XXSSProtection = "1; mode=block";
 			context.Response.Headers.XFrameOptions = "DENY";
 			context.Response.Headers.XContentTypeOptions = "nosniff";
 			context.Response.Headers["Referrer-Policy"] = "strict-origin-when-cross-origin";
-			context.Response.Headers.XPoweredBy = "";
+			context.Response.Headers.XPoweredBy = string.Empty; // why tho --yoshi
```
adelikat, line 67: Please don't introduce string.Empty.  We use "" in this repo and I want to keep it consistent

TASVideos/Extensions/ApplicationBuilderExtensions.cs
```diff
@@ -42,13 +42,29 @@ public static IApplicationBuilder UseStaticFilesWithExtensionMapping(this IAppli
...
+			"screen-wake-lock=()", // defaults to `self`
+			"web-share=()", // defaults to `self`
+
+			// ...and that's all of the non-experimental options listed on MDN as of 2024-04
+		]);
 		app.Use(async (context, next) =>
 		{
+			context.Response.Headers["Cross-Origin-Embedder-Policy"] = "credentialless";
+			context.Response.Headers["Cross-Origin-Opener-Policy"] = "same-origin";
+			context.Response.Headers["Cross-Origin-Resource-Policy"] = "cross-origin"; // TODO this is as unsecure as before; should be `same-site` or `same-origin` when serving auth-gated responses
```
adelikat, line 61: so why are we setting it to cross origin if it is "unsecure" is ther ea reason we can't do same-site or same-origin?

TASVideos/Extensions/ApplicationBuilderExtensions.cs
YoshiRulz: I'll force-push to revert this once the other thread is resolved.

Why is `X-Powered-By:` being added though?

TASVideos/Extensions/ApplicationBuilderExtensions.cs
YoshiRulz: Sending `CORP: same-origin` with every response will prevent JS on other sites from fetching that page. For a site whose content is mostly public, that doesn't seem fitting, but of course you may want to push third-parties towards the API rather than scraping the site. I say "unsecure", but I'm pretty sure this header only applies to the JS Fetch API, not `<iframe/>`s, and browsers won't include cookies with those requests.

TASVideos/Extensions/ApplicationBuilderExtensions.cs
adelikat: If we do not set it to nothing, the default is to broadcast what technology we are using including which version, which is flagged as not ideal

adelikat (approved):

PR #1842, opened 2024-05-09
Improve handling of "extlinks"

YoshiRulz:


YoshiRulz:
@nattthebear Is this the right approach? It's working well now, but I foresee problems with trying to combine `rel` values—specifically, I'm thinking of adding `rel="ugc"` to only *some* wiki-format excerpts, but as I've written it that would override the other values. So I would need to pass `"noopener external nofollow ugc"` and then that has to be kept in sync. Is there an existing mechanism for combining `class` that I've overlooked?

nattthebear:
@YoshiRulz I don't think you're missing any capabilities.  The system has about enough abstraction layers to handle what it does, but no more.

YoshiRulz:
Then should I contribute first-class support for those attributes before finalising this?

nattthebear:
If you think it's useful, I could see some sort of `rel` and `class` handling as a useful part of the `Node` capabilities.

adelikat (approved):

adelikat:
@YoshiRulz : can you fix the merge conflicts in this PR? and then I'll merge

PR #1849, opened 2024-05-13
Enhance handling of `class` attr in `HtmlWriter`

YoshiRulz:
per https://github.com/TASVideos/tasvideos/pull/1842#issuecomment-2104604073

### DIFF
--- a/TASVideos.Common/HtmlWriter.cs
+++ b/TASVideos.Common/HtmlWriter.cs
@@ -36,12 +36,7 @@ public partial class HtmlWriter(TextWriter w)
 			throw new InvalidOperationException("Can't open a void tag");
 		}
 
-		if (_inTagOpen)
-		{
-			w.Write('>');
-		}
-
-		tagName = tagName.ToLowerInvariant();
+		FinalizeOpenTag();
 		w.Write('<');
 		w.Write(tagName);
 		_openTags.Push(tagName);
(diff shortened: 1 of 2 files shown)


TASVideos.Common/HtmlWriter.cs
```diff
@@ -4,6 +4,142 @@ namespace TASVideos.Common;
...
+			throw new NotImplementedException();
+		}
+
+		public bool Remove(string item)
+		{
+			throw new NotImplementedException();
+		}
+*/
+
+		public readonly string Serialise()
```
YoshiRulz, line 106: Use AmE spelling (and below)

TASVideos.Common/HtmlWriter.cs
```diff
@@ -4,6 +4,142 @@ namespace TASVideos.Common;
...
+			else
+			{
+				return string.Empty;
+			}
+
+			return classes.Count is 0 ? string.Empty : $" class=\"{string.Join(' ', classes)}\"";
+		}
+
+/*
+		public readonly bool SetEquals(IEnumerable<string> other)
```
YoshiRulz, line 126: Didn't end up using the `ISet` impl., and I don't see that being necessary in the future. Will remove.

YoshiRulz (commented):
> Pull request authors can’t request changes on their own pull request

TASVideos.Common/HtmlWriter.cs
```diff
@@ -4,6 +4,49 @@ namespace TASVideos.Common;
...
+			else if (_list is not null)
+			{
+				classes = _list.SelectMany(static s => s.Split(' ', StringSplitOptions.RemoveEmptyEntries)).ToHashSet();
+			}
+			else
+			{
+				return "";
+			}
+
+			return classes.Count is 0 ? "" : $" class=\"{string.Join(' ', classes)}\"";
```
nattthebear, line 46: Two problems here:

* We're unnecessarily allocating an additional string and making an extra copy; why not just write to the output?
* Does this handle quote and ampersand escaping?  It doesn't seem to.  Ideally, it would use the same logic as the rest of the system uses to handle that.

YoshiRulz: Whoops! You're totally right.

nattthebear (approved):

PR #1854, opened 2024-05-19
Re-add `Cross-Origin-Embedder-Policy` header with a more lax value

### LINKED ISSUE #1852: All Youtube Embeds are blocked
All YouTube embeds show "Blocked Page".

In firefox, the request is cancelled with a "NS_ERROR_DOM_COEP_FAILED" error, which means
> Used to indicate that a resource load was blocked because of the Cross-Origin-Embedder-Policy response header

Example:
image

YoshiRulz:
see #1852, https://github.com/TASVideos/tasvideos/commit/59dac25a981d90fad1b29a03874b98e31aeb532c

### DIFF
--- a/TASVideos/Extensions/ApplicationBuilderExtensions.cs
+++ b/TASVideos/Extensions/ApplicationBuilderExtensions.cs
@@ -56,6 +56,7 @@ public static class ApplicationBuilderExtensions
 		]);
 		app.Use(async (context, next) =>
 		{
+			context.Response.Headers["Cross-Origin-Embedder-Policy"] = "unsafe-none"; // this is as unsecure as before, but can't use `credentialless`, see https://github.com/TASVideos/tasvideos/issues/1852
 			context.Response.Headers["Cross-Origin-Opener-Policy"] = "same-origin";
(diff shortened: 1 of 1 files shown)


TASVideos/Extensions/ApplicationBuilderExtensions.cs
```diff
@@ -56,6 +56,7 @@ public static IApplicationBuilder UseMvcWithOptions(this IApplicationBuilder app
 		]);
 		app.Use(async (context, next) =>
 		{
+			context.Response.Headers["Cross-Origin-Embedder-Policy"] = "unsafe-none"; // this is as unsecure as before, but can't use `credentialless`, see https://github.com/TASVideos/tasvideos/issues/1852
```
adelikat, line 59: If it is as unsecure as before, what is the value of adding this?

YoshiRulz: Being explicit, and resolving warnings on securityheaders.com and the like.

adelikat (approved):

Issue #1858, opened 2024-05-24
Wiki links with spaces are rendered without spaces


Randomno:
For example on https://tasvideos.org/RecentChanges, the page has a link to `[new movies]`, but this is rendered as `Newmovies`. It was previously rendered as `new movies` until 1c8034d.

nattthebear:
`[new movies]` gets normalized to a url of `/Newmovies`, which is correct and has not changed.  Formerly, we passed the original unnormalized text to use as the `displaytext` argument to `__wikiLink`.  Since 1c8034d679136dec196126ffcfa1fbac5f0c07f6, we don't do that because we need to distinguish cases where the user didn't explicitly provide display text and replace those with canonical object titles.  So `[123M]` gets the display text "NES Guerrilla War by lithven in 14:39.18", but `[123M|123M]` gets the display text "123M" and the tooltip text "NES Guerrilla War by lithven in 14:39.18".

I think to resolve this case we'd need to distinguish between "strong" display text, where the user explicitly provided something after the `|`, and "weak" display text, where the display text was inferred from the link, but due to the normalization process, still might have some textual difference from the link.

The effective behavior is the result of a combination of logic interacting in both https://github.com/TASVideos/tasvideos/blob/b09dee4810d727256e5b214f8aaa7f03b145430c/TASVideos.WikiEngine/MakeBracketed.cs and https://github.com/TASVideos/tasvideos/blob/ac84694cde36974a5e3e3834b8f6c69716d39e6a/TASVideos/WikiModules/WikiLink.cshtml.cs.  So what we really need are some automated tests that test the entire wiki process end to end, starting at wiki markup and ending with the final HTML output, including any Razor module results.  I can easily test the wiki by itself, but I don't know how to make meaningful offline tests that boot up the entire MVC+Razor system.

adelikat:
[recent changes] is effectively syntactical sugar for [RecentChanges|recent changes, maybe this isn't robust enough for syntactical sugar and the effort isn't worth the value?  We could just drop that support in a way that causes a broken link such that we can go through and fix those instances

nattthebear:
> `[recent changes]` is effectively syntactical sugar for `[RecentChanges|recent changes]`

[...]

PR #1859, opened 2024-05-26
Fix rendering of certain wiki links

nattthebear:
When the link has no display text, but the normalized form differs from the unnormalized form, we were missing formatting since 1c8034d679136dec196126ffcfa1fbac5f0c07f6.

Fixes https://github.com/TASVideos/tasvideos/issues/1858.


adelikat:
the approach looks good to me

nattthebear:
I don't have a localdev set up at the moment so I can't validate the correctness of this, but otherwise, I'm fine with merging it.

adelikat:
Seems to work in all the scenarios I could think of

adelikat (approved):

PR #1869, opened 2024-06-02
Allow case edits in Game Goals

### LINKED ISSUE #1868: Can't change casing of goal name
https://tasvideos.org/Games/180/Goals/List

I tried changing `Maximum Levels` to `maximum levels` and got an error `Cannot change goal to maximum levels because it already exists.`

Masterjun3:
Resolves #1868 .

@adelikat ~~I assume my edit is what that particular code block was meant to do in the first place. If my assumption is wrong you should definitely look at this.
(Because we use case-insensitive columns, right? Requiring us to preempt the DB and manually allow changes if the only difference is their case.)~~
I removed that whole if-block, but I wasn't quite sure what its purpose was. I assumed it had something to do with our DB using case-insensitive columns, but this shouldn't matter here.

### DIFF
--- a/TASVideos/Pages/Games/Goals/List.cshtml.cs
+++ b/TASVideos/Pages/Games/Goals/List.cshtml.cs
@@ -86,7 +86,7 @@ public class ListModel(ApplicationDbContext db) : BasePageModel
 
 		var oldGoalName = gameGoal.DisplayName;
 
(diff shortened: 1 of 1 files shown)


Masterjun3:
Actually, this short-circuits the function, so Pubs and Subs aren't updated.
And it seems this code block isn't needed anyway. So let me add a commit.

adelikat (approved):

Issue #1701, opened 2023-09-18
Rework the interface for user role management to allow temporary roles


moozooh:
Right now all roles can only be removed manually, and the current interface is not very conducive to implementation of temporary roles in general.

The immediate issue that stems from this is the inability to make temporary bans. Every such thing needs to be managed manually, by putting dates on the calendar and all that. Not ideal, and prone to human factors.

The change I propose is the interface to optionally set a time limit (in days, or by entering a specific date/time) to a role upon which it is removed automatically. This time limit, if set, would count down and be visible to moderators, senior staff, and the user in question on their profile.

The underlying logic could be useful as a stepping stone for implementing potential future features like probation, various notifications (e.g. informing the staff that such-and-such has been unbanned), possible safety-related features, etc.

adelikat:
I don't think you need any of this at all.

I think the root of the problem is bans are done as a removal of a role, which led to another workaround of having a ban role with no perms.

Simpler would be to detach banning from roles and have some kind of ban flag. It can be a date to make it temporary, that would prevent login regardless of role

moozooh:
I don't know if turning it into a flag will make it conceptually simpler than a role, but... sure, I suppose? Either way there needs to be an interface for this with some sort of a countdown timer.

adelikat:
This is related to #1543 

I think for both, we need a flag and/or expiry date.  We need to know that someone was banned, that they simply lack roles

adelikat:
The crux of this issue was the ability to do temporary bans which can now be done, closing this issue

PR #1870, opened 2024-06-02
Banned until

adelikat:
Resolves the immediate problem in #1701 
The full rollout will require more iterations, this PR adds a "Banned Until" datetime on user.  If set, the user cannot log in and effectively has no permissions, so long as the current time is before the ban.  

If a user is banned this way, their role on the forum will display as "Banned User"

Important note that this ties banning to the EditUser permisisons where before it was the AssignUser (due to needing to remove a Login permission).

In future commits and migrations, we will deprecate the Login permission, and remove the Banned User role

Examples of how it would display on the User profile (no Spikes were harmed in the making of these images, was purely from test data):
image

image

### DIFF
--- a/TASVideos.Core/Services/SignInManager.cs
+++ b/TASVideos.Core/Services/SignInManager.cs
@@ -36,6 +36,11 @@ public class SignInManager(
 			return (SignInResult.Failed, null);
 		}
 
+		if (user.BannedUntil < DateTime.UtcNow)
+		{
+			return (SignInResult.Failed, null);
+		}
+
 		var claims = await userManager.AddUserPermissionsToClaims(user);
 		var canLogIn = claims.Permissions().Contains(PermissionTo.Login);
 
--- a/TASVideos.Core/Services/SignInManager.cs
+++ b/TASVideos.Core/Services/SignInManager.cs
@@ -36,7 +36,7 @@ public class SignInManager(
 			return (SignInResult.Failed, null);
 		}
 
-		if (user.BannedUntil < DateTime.UtcNow)
+		if (user.IsBanned())
 		{
 			return (SignInResult.Failed, null);
 		}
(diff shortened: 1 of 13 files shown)


TASVideos/Pages/Forum/Topics/_TopicPost.cshtml
```diff
@@ -68,7 +68,7 @@
 									<small condition="@(Model.PosterPronouns != PreferredPronounTypes.Unspecified)" class="text-body-tertiary col-auto col-md-12" style="line-height: 22px;">@Html.DisplayFor(m => m.PosterPronouns)</small>
 								</row>
 								<div class="ms-2 ms-md-0" style="line-height: 1">
-									<small>@string.Join(", ", Model.PosterRoles.OrderBy(s => s).Append(Model.PosterPlayerRank).Where(s => !string.IsNullOrEmpty(s)).Select(s => s!.Replace(' ', '\u00A0')))
+									<small>@string.Join(", ", Model.CalculatedRoles)
```
Masterjun3, line 71: This seems to change things quite a bit. 
1. The roles now don't have a space between them anymore `Editor,Judge,Expert player` instead of `Editor, Judge, Expert player`.
2. The second string.Join is not necessary anymore.
3. Banned users will still show their points, but not their rank, `Banned User (534)` which seems weird.

TASVideos.Core/Services/SignInManager.cs
```diff
@@ -36,6 +36,11 @@ public async Task<(SignInResult Result, User? User)> SignIn(string userName, str
 			return (SignInResult.Failed, null);
 		}
 
+		if (user.BannedUntil < DateTime.UtcNow)
```
Masterjun3, line 39: Shouldn't this be the other way around?

TASVideos/Pages/Forum/Topics/_TopicPost.cshtml
adelikat: fixed those 3 things

TASVideos.Core/Services/SignInManager.cs
Masterjun3: >

Masterjun3 (changes requested):

Masterjun3 (approved):

PR #1871, opened 2024-06-02
Add Pub and Sub counts to Platform Framerates module

Masterjun3:
For the page https://tasvideos.org/PlatformFramerates .
We have many single-use framerates. Having the counts make the table more useful as a resource to see which are the "correct" framerates.

Example image:

image


adelikat (approved):

adelikat:
This is a nice addition to the page :)

vadosnaprimer (approved):

PR #1874, opened 2024-06-04
More image adjustments.

Spikestuff:
This includes finally fixing the newbie icon to actually be 2x/4x (and actually be transparent), as well as some sort of artifacting error left behind (or something) on verified.png for instance where it hid the star icon.png within it.


Masterjun3 (approved):
The hidden star icon in the verified png is finally gone :(
It was my favorite easter egg.

PR #1881, opened 2024-06-12
Add script-src to Content-Security-Policy

Masterjun3:
By adding the script-src to our Content-Security-Policy (CSP), we limit the places scripts can be executed from. This commit 29a84245f862a112bc71bd86c7173ceacf3f8606 prepared us for this, because we don't want to allow unsafe-inline and unsafe-eval.

However, now come some problems. I think some of our embeds throw weird errors with a CSP like this.
Luckily, browsers' javascript console tells us exactly if something failed due to our CSP, so maybe we just need to go through a bunch of our pages to validate everything works.

~~Anyway, the embed problem is why it's a draft for now.~~
Actually, it was my adblocker using CSP to block things. It's an iframe, so it has its own stuff. If we want to block what iframes can do, we need to use a different part of CSP, `frame-src` or `child-src`. But allowing iframes from everywhere is less of a security risk for now.

So this PR is ready now.

### DIFF
--- a/TASVideos/Extensions/ApplicationBuilderExtensions.cs
+++ b/TASVideos/Extensions/ApplicationBuilderExtensions.cs
@@ -65,7 +65,7 @@ public static class ApplicationBuilderExtensions
 			context.Response.Headers.XContentTypeOptions = "nosniff";
 			context.Response.Headers["Referrer-Policy"] = "strict-origin-when-cross-origin";
 			context.Response.Headers.XPoweredBy = "";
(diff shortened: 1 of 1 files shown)


adelikat:
Embeds were already throwing errors because of our policy but the playback is fine, does this cause new errors?

Masterjun3:
I edited the opening post! Errors from my adblocker seemed related, but weren't. Playback works fine. This PR is ready now.

adelikat (approved):

PR #1882, opened 2024-06-12
list languages partial - convert to a dropdown in the wiki menu bar

adelikat:
Converts the translations list into a dropdown and puts it in the button bar along with the edit button

image


Masterjun3 (approved):

Masterjun3 (changes requested):
Now only people who can edit the wiki can see the language select, no?

Masterjun3 (approved):

vadosnaprimer (approved):

PR #1884, opened 2024-06-14
this seems to fix #1883

### LINKED ISSUE #1883: Permissions are sometimes broken, Require Attribute not allowing anything
Anyone who logs out and back in can't make e.g. Posts or Topics.

~~This seems to be the cause eac610fe9734671064897641fd7e2b192552a614 .~~ Sometimes I can post, sometimes I can't, I can't figure out when this happens.

Via code debugging it seems sometimes the user claims simply don't include the permissions.

adelikat:


Masterjun3 (commented):
It turns out this actually adds our permissions to the claims *twice*.
Once in the `AddUserPermissionsToClaims` with the `AddClaimsAsync`, and once in the `SignInWithClaimsAsync`.

So instead of SignInWithClaimsAsync we could just do SignInAsync.
(Or we could not add them in the userManager.AddUserPermissionsToClaims, but that method is used elsewhere, so I'm not sure.)

adelikat:
Updated, I chose to use SignInAsync.  However, I think we are doing too much work, everytime we do a requirePermission check we are adding those claims.  I'm wondering if this would be unnecessary if we did the SignInWithClaims when the user logged in.  But I'm not sure

Masterjun3 (approved):

PR #1890, opened 2024-06-27
Add fm2 binary parser

Randomno:
FM2 format officially supports binary format, even though the only way you would probably end up with such a movie is by renaming an FM3.

Does this need a test?


Masterjun3:
> Does this need a test?

Yes.

adelikat (changes requested):
Add unit tests please

Randomno:
Added

adelikat (approved):

Issue #1644, opened 2023-08-13
Idea: Include page content in our embeds (Open Graph Meta Tags)


Masterjun3:
_This is an idea, and rather than immediately implementing it, I wanted this ticket to disuss options._
Embeds of links to our site are controlled by og tags, and currently our embeds look like this:

<details><summary>Current embeds</summary>

![image](https://github.com/TASVideos/tasvideos/assets/22375320/e67993b1-3e27-49ce-a02d-9680f384851a)
![image](https://github.com/TASVideos/tasvideos/assets/22375320/ff6dd097-6f26-43ac-a895-39843d0e5168)
![image](https://github.com/TASVideos/tasvideos/assets/22375320/2912d660-3fb1-46f6-95d2-81b598f84667)

</details>

My idea was to _actually include the first interesting parts_ of the Post/Topic/Wiki Page into the description field of the embed. Additionally one idea is to take the first image of the content as the display image (it seems many pages do this).

A result might look like this (before | after):
![image](https://github.com/TASVideos/tasvideos/assets/22375320/1e4c06b2-d037-4612-ab81-c950bd8efb73)

Ignoring Publications and Submissions for now, because their embeds have a different layout, and are easily recognizable as that.

## Discussion

It is difficult to properly include every information that might be important. For example the author of Wiki Pages and Topics is not that important, but for Posts including the author is essential.

I've thought about what I want to include as necessary into embeds, and this is what I came up with:
### Forum or Wiki Page
There needs to be a way to tell whether you're going to click on something of the forum, or on something on the wiki (the rare System pages should have automatic or easy recognition). The reason why this is necessary can be seen with these two pages: https://tasvideos.org/Forum/Topics/21178 and https://tasvideos.org/Bizhawk/GB .
The first one is `Forum → Bizhawk → Gameboy BIOS`, a question about BG bios, the second is `Wiki → BizHawk/GB`, the official wiki page about the bizhawk GB core.
Taking away the information about Forum or Wiki would result in confusion about the page you're about to land on.

### Location
"Location" is what I call the `LuaScripting` part of the `LuaScripting/Display` page, or the `Site` of a `Site → Question.` topic or a post in there. The location seems less important, and yet I don't think embeds are useful without them.

### Topic Title
This one is obvious, forum topics have titles, and those have to make it into the embeds, ideally into the title.
Currently topic titles are in the description, because Posts don't have titles, so "Topic by user" works next to "Post by user".

### Post Author
[...]

PR #1896, opened 2024-06-27
Put Wiki Content into Open Graph Descriptions

Masterjun3:
Relevant to #1644 .

Basically, our meta tag descriptions on wiki pages currently show when the page was last edited. Which isn't really useful.
What would be useful instead is to take the first few lines of the wiki and put the text of that into the description!

So this was the idea (before | after):
image

Now, our wiki system is quite complicated. And we don't know if we want to add even more complexity to it. I basically tried to minimize the changes, and I made 3 commits that build upon each other, but are self-contained and can be run on their own.
- 1st commit: Use the existing Wiki-To-Youtube text conversion.

_Problem_: The Youtube renderer renders links. But that isn't really useful for meta tag descriptions. This is somewhat solved by the 2nd commit.

- 2nd commit: Write my own renderer, but still use the existing WriteTextAsync methods on the INode implementations.

_Problem_: All the big things like TOCs and Tabs are rendered at the start, which makes the meta tag description useless again.

- 3rd commit: Write my own renderer, and also make my own WriteMetaDescriptionAsync in the INode implementations.

Finally it looks good to me.

### However
However, this is quite the complexity added. If this is too much or we want to implement it differently, maybe after a rewrite or refactor of wiki stuff, then this PR can be closed. I just wanted to try how much it really is, because I do enjoy the results:

More examples:
image
image
image


TASVideos.WikiEngine/Node.cs
```diff
@@ -22,6 +22,8 @@ public interface INode
 	/// </summary>
 	Task WriteTextAsync(TextWriter writer, WriterContext ctx);
 
+	Task WriteMetaDescriptionAsync(StringBuilder sb, WriterContext ctx);
```
nattthebear, line 25: Could these take a `TextWriter` instead of a `StringBuilder` for consistency with other node implementations?  This shouldn't require significant changes to the caller as we can use `StringWriter` there as a concrete `TextWriter` implementation.

TASVideos.WikiEngine/NodeImplementations/Module.cs
```diff
@@ -65,6 +65,33 @@ public async Task WriteTextAsync(TextWriter writer, WriterContext ctx)
...
+	public async Task WriteMetaDescriptionAsync(StringBuilder sb, WriterContext ctx)
+	{
+		if (sb.Length == SiteGlobalConstants.MetaDescriptionLength)
+		{
+			return;
+		}
+
+		if (ModuleNames.IsModule(Name))
+		{
+			StringWriter sw = new();
```
nattthebear, line 77: Could this pass through the parent `TextWriter` instead of making a new one?

nattthebear (commented):
This is a solid implementation.  There's a complexity increase, but it's not significant and we gain useful value.  Many commercial wiki products have significantly worse open graph description renderers than this 😅.

I've left a few suggestions.

TASVideos.WikiEngine/NodeImplementations/Module.cs
Masterjun3: What parent `TextWriter`?

TASVideos.WikiEngine/Node.cs
Masterjun3: No, because I need the length of the currently written text. TextWriters don't have that ability.

nattthebear: Could the truncation be the responsibility of a custom `TextWriter` implementation?

TASVideos.WikiEngine/NodeImplementations/Module.cs
nattthebear: This is assuming the other comment is acted on; then there would be a parent `TextWriter` passed in here.

TASVideos.WikiEngine/Node.cs
Masterjun3: Truncating would work, yes, but I'm using the length for deciding whether to go through more nodes, to avoid having to iterate the entire node tree once we already have enough.

nattthebear: > to avoid having to iterate the entire node tree once we already have enough.

I think that's a good idea, but I don't see it happening in most places; I'll open separate comments there.

TASVideos.WikiEngine/Util.cs
```diff
@@ -72,6 +72,25 @@ public static async Task RenderTextAsync(string content, TextWriter w, IWriterHe
...
+		{
+			results = NewParser.Parse(content);
+		}
+		catch (NewParser.SyntaxException e)
+		{
+			results = Builtins.MakeErrorPage(content, e);
+		}
+
+		var ctx = new WriterContext(h);
+		foreach (var r in results)
```
nattthebear, line 88: We should break this loop early once the `sb` is full.  Most pages will have a lot of their content directly under the top level list, so stopping early here could save time.

TASVideos.WikiEngine/NodeImplementations/Element.cs
```diff
@@ -164,6 +165,18 @@ public async Task WriteTextAsync(TextWriter writer, WriterContext ctx)
 		}
 	}
 
+	public async Task WriteMetaDescriptionAsync(StringBuilder sb, WriterContext ctx)
+	{
+		// write all except for divs that aren't p (which aims to exclude TOC, tabs, etc.)
+		if (Tag != "div" || Attributes.TryGetValue("class", out string? classes) && classes.Split(' ').Contains("p"))
+		{
+			foreach (var c in Children)
```
nattthebear, line 173: We should break this loop early if the `sb` is full.  If a page has tabs, large portions of its content may be nested inside individual tags and we could save time by stopping the iteration here.

TASVideos.WikiEngine/Node.cs
nattthebear: Resolving this as I think it makes sense to stick with the stringbuilder.

TASVideos.WikiEngine/NodeImplementations/Module.cs
nattthebear: Resolving as we'll keep the stringbuilder as discussed elsewhere.

Masterjun3:
Good review @nattthebear . It makes sense, if I use optimization as a reason to not follow convention, then I should properly optimize everywhere. I will resolve this in a few hours, when I have more time.

Masterjun3:
Those commits should do it.

nattthebear (approved):
This is a great feature.  Thanks for building it.

PR #1897, opened 2024-06-29
Publish form - add alternative online watching url field

adelikat:
Removes the OnlineUrl Display Name since the main encode is never labeled
Adds an alterntative online-watching url field and display name

image


vadosnaprimer (approved):
Thanks!

adelikat:
"This should actually be listed as "required"." 
Not true! I think you mean you want it required IF there is an alt url

PR #1906, opened 2024-06-30
Some refactors to WikiEngine, attempting to switch to `ReadOnlySpan<char>`

YoshiRulz:
~~I'll clean up the diffs when I'm not so sleep-deprived. At this stage, I'd like to know whether you think this is a good idea at all, whether the final signatures are good, and if you'd like any of the added types/methods placed somewhere else. Oh, and if you have any ideas for removing one of the remaining `ToString` calls.~~ I think this is finished now

I still hate the StyleCop settings.

### DIFF
--- a/TASVideos.Common/Extensions/StringExtensions.cs
+++ b/TASVideos.Common/Extensions/StringExtensions.cs
@@ -84,6 +84,19 @@ public static partial class StringExtensions
 			: str[..limit];
 	}
 
+	public static bool EndsWithAny(this ReadOnlySpan<char> str, IEnumerable<string> suffixes)
+	{
+		foreach (var sfx in suffixes)
+		{
+			if (str.EndsWith(sfx))
+			{
+				return true;
+			}
+		}
+
+		return false;
+	}
+
 	public static RegexMatchShim Match(this Regex regex, ReadOnlySpan<char> str)
 	{
 		var iter = regex.EnumerateMatches(str);
@@ -198,6 +211,19 @@ public static partial class StringExtensions
 		return !string.IsNullOrWhiteSpace(str)
 			? SplitCamelCaseRegex.Replace(str, "$1$2$3$4$5 ")
 			: "";
+	}
+
+	public static bool StartsWithAny(this ReadOnlySpan<char> str, IEnumerable<string> prefixes)
+	{
+		foreach (var pfx in prefixes)
+		{
+			if (str.StartsWith(pfx))
+			{
+				return true;
+			}
+		}
+
+		return false;
 	}
 
 	private static readonly Regex SpaceRegex = SpaceCompiledRegex();
--- a/TASVideos.Common/Extensions/StringExtensions.cs
+++ b/TASVideos.Common/Extensions/StringExtensions.cs
@@ -2,6 +2,39 @@
 
 namespace TASVideos.Extensions;
 
+public ref struct RegexMatchShim
+{
+	private const string ErrMsgZeroed = $"this {nameof(RegexMatchShim)} has no value";
+
+	private GroupCollection? _groups = null;
+
+	private readonly Regex _regex;
+
+	private readonly ReadOnlySpan<char> _str;
+
+	private readonly Range _valueRange;
+
+	public GroupCollection Groups
+		=> Success
+			? (_groups ??= _regex.Match(_str.ToString()).Groups)
+			: throw new InvalidOperationException(ErrMsgZeroed);
+
+	public readonly bool Success;
+
+	public readonly ReadOnlySpan<char> Value
+		=> Success
+			? _str[_valueRange]
+			: throw new InvalidOperationException(ErrMsgZeroed);
+
+	internal RegexMatchShim(Regex regex, ReadOnlySpan<char> str, Range valueRange)
+	{
+		_regex = regex;
+		_str = str;
+		_valueRange = valueRange;
+		Success = true;
+	}
+}
+
 public static partial class StringExtensions
 {
 	/// <summary>
@@ -49,6 +82,14 @@ public static partial class StringExtensions
 		return str.Length < limit
 			? str
 			: str[..limit];
+	}
+
+	public static RegexMatchShim Match(this Regex regex, ReadOnlySpan<char> str)
+	{
+		var iter = regex.EnumerateMatches(str);
+		return iter.MoveNext()
+			? new(regex, str, iter.Current.Index..(iter.Current.Index + iter.Current.Length))
+			: default;
 	}
 
 	/// <summary>
--- a/TASVideos.WikiEngine/NodeImplementations/Element.cs
+++ b/TASVideos.WikiEngine/NodeImplementations/Element.cs
@@ -25,12 +25,20 @@ public partial class Element : INodeWithChildren
 		"wbr"
 	];
 	public NodeType Type => NodeType.Element;
-	public List<INode> Children { get; private set; } = [];
-	public IDictionary<string, string> Attributes { get; private set; } = new Dictionary<string, string>();
+
+	public List<INode> Children { get; private set; }
+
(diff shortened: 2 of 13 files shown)


TASVideos.Common/Extensions/StringExtensions.cs
```diff
@@ -2,6 +2,41 @@
 
 namespace TASVideos.Extensions;
 
+public ref struct RegexMatchShim
+{
+#pragma warning disable SA1310
```
adelikat, line 7: No.  Why do this? Just name the variable ErrMsgZeroed and play nice with the naming conventions of this repository

TASVideos.WikiEngine/NodeImplementations/Module.cs
```diff
@@ -29,9 +41,11 @@ public async Task WriteHtmlAsync(TextWriter w, WriterContext ctx)
 		{
 			if (!ctx.AddTdStyleFilter(Parameters))
 			{
-				var div = new Element(CharStart, "div") { CharEnd = CharEnd };
-				div.Children.Add(new Text(CharStart, "Module Error for settableattributes: Couldn't parse parameter string.") { CharEnd = CharEnd });
-				div.Attributes["class"] = "module-error";
+				Element div = new(
+					_charRange,
+					"div",
+					attributes: [new("class", "module-error")],
```
nattthebear, line 47: This creates an extra object for the attributes collection compared to the old version of the code.  This isn't entirely a new problem, as there were some other callsites that wasted objects before, but it's worse now.

TASVideos.WikiEngine/NodeImplementations/Element.cs
```diff
@@ -43,24 +63,38 @@ public Element(int charStart, string tag)
...
+
+	public Element(
+		int charStart,
+		string tag,
+		IEnumerable<KeyValuePair<string, string>> attributes,
+		params INode[] children)
+			: this((charStart, default), tag, attributes, children: children.AsEnumerable()) { }
+
+	public Element(StringIndices range, string tag)
+		: this(range, tag, [], children: (IEnumerable<INode>)[]) { }
```
nattthebear, line 94: Does the use of `[]` here automatically pool a single Enumerable.Empty or are we making new objects every time?

CasualPokePlayer: It just uses [Array.Empty](https://sharplab.io/#v2:D4AQTAjAsAUCAMACEEAsBuWsBiB7XAdACoCmAzgC4AUA2gLoCUmMsAlgHYUkBO7AhgBtkEAGzIwiPLlgBvWIgXIAzMLEhUiUpSoolAHgDSJAJ4A1QQFcSABT6tuelPAA0w+AD53iAGb4G8xTkYRUQAX1hwmCA===).

TASVideos.WikiEngine/NodeImplementations/Module.cs
YoshiRulz: ~~Wasn't it allocating the same number of collections before, just in the `Element` ctor?~~ Ah I see, this is an array + `ToDictionary`. Should I go back to inserting after init, or should I add constructors which take a `Dictionary`?

nattthebear: Passing a `Dictionary` seems good so long as we can make sure we never accidentally share those dictionaries (the passed object should be owned by the element that created it)

YoshiRulz: Sorry, do you mean owned by the `Element` or owned by the call-site? Not that I know how to put either into code—mutation of `Element.Attributes` is relied on by a few call-sites, so it can't simply be changed to `ImmutableDictionary` for example.

nattthebear: Owned by the element; I don't want to accidentally share a single Dictionary across multiple elements, as that would be too confusing for now.

YoshiRulz: There are actually fewer mutations than I thought (must have confused it with `Children`), but even after encapsulating them, I can't think of a good way to enforce that invariant that's not just defensive copying.
I tried `[CallerArgumentExpression]` + checking for `new Dictionary` or `.ToDictionary(`, but while it may have worked, it also would have made every callsite verbose. Plus, half the callsites use the `Attr` method, so they would have to be `new Dictionary(new[] { Attr(..., ...) })` which I believe allocates no less than `new KVP<string, string>[] { Attr(..., ...) }.ToDictionary()`.
Unless you have any ideas, I'll go ahead with replacing `IEnumerable`s with `Dictionary`s, and it will just have to rely on humans not misusing references.

nattthebear:
I feel lukewarm on this PR as a whole.  Avoiding needless copies is good for speedup, and that certainly includes `ReadOnlySpan<char>` usage.  But it makes some complex code even more complex, and small copies don't have a large perf impact.

Do we have profiling evidence from production that wiki rendering is slow enough to be a problem?  What money are we trying to save here anyway; I don't have any access to the cost breakdown of running production?

If wiki rendering is the problem, then we could likely save more speed, and with simpler code, by caching partially hydrated pages.

Masterjun3:
So to me neither the PR title, nor the PR description explain what this does and what its purpose is.
By looking at the diff it seems to me it does 3 things at the same time?
1. Combine the CharStart and CharEnd ints into a single Range variable.
2. Replace strings with ReadOnlySpan<char>. (Why?)
3. Do something with Regex Shim. (What is Shim, what does this do?)

So basically I have the following questions:
- What do these things do?
- What is their purpose? (It doesn't seem to be code-simplification or code-reduction, which is usually what I know from refactors.)
- Do they have to be done all at the same time in the same PR?

YoshiRulz:
I'll try to explain further, though as natt noted, the key change of `ReadOnlySpan<char>` is *inherently beneficial* but of *questionable impact* without performance data.

Merging the start and end indices into a tuple is just a simplification. It also highlights the call-sites where `CharEnd` wasn't being set, as they now use separate constructor overloads. (I originally looked to use the BCL `Range` for this, but that would have required extra checks on init and extra processing to keep compatibility with the `Char{Start,End}` props.)

You didn't ask about the `Element` constructor refactor, but that was to reduce mutation, and also setup for the subsequent commits to be smaller.

The `RegexMatchShim` struct, as the name suggests, is a lightweight wrapper over the BCL's `ReadOnlySpan<char>` RegEx API to make it resemble the `string` one. Again this was to keep the diff small.

You also didn't ask about the first commit. That contains all the changes I would have had to do in the last commit, but doing them first reduced churn. Some of that I think would be a good idea outside of this PR, namely passing an `IEnumerable<string>` to `MakeImage`, using the correct RegEx method, a dedup in `MakeLinkOrImage`, and lifting the ternaries outside `string.Concat`.

So yes, with those few exceptions it does all have to be done in 1 PR, since the later commits depend on the previous, and merging only some without actually switching to `ReadOnlySpan<char>` somewhat defeats the purpose.

TASVideos.WikiEngine/NodeImplementations/Module.cs
```diff
@@ -5,12 +5,24 @@ public class Module : INode
 	public NodeType Type => NodeType.Module;
 	public string Name { get; }
 	public IReadOnlyDictionary<string, string> Parameters { get; }
-	public int CharStart { get; }
-	public int CharEnd { get; set; }
-	public Module(int charStart, int charEnd, string text)
+
+	private StringIndices _charRange = (default, default);
```
YoshiRulz, line 9: This field init was necessary at one point, but in the end only the tuple ctor is left, so it can be removed.

adelikat:
I do really wanted the span usages pulled in, but I think this does too much in one go.

I don't think it is responsible to do these refactors without first setting up some benchmarks where we can show that we have actually improved memory allocations with a proposed change, effort should be put into that first

PR #1907, opened 2024-07-01
LTM: Set system to Flash if game_name contains ruffle

### LINKED ISSUE #1905: LTM movie parser: autodetect Flash from `game_name=ruffle`
Set system to Flash for LTM if there's `game_name=ruffle` in `config.ini`. A lot of people forget to put it into annotations, and I just realized we can stop depending on that at least with ruffle (with PCem we still do since it can emulate different systems).

Randomno:
Resolves #1905

Can be overridden by setting the platform annotation to something else.


vadosnaprimer (approved):

tests/TASVideos.MovieParsers.Tests/TASVideos.MovieParsers.Tests.csproj
```diff
@@ -28,6 +28,11 @@
 		<EmbeddedResource Include="WtfSampleFiles\*.wtf" />
 	</ItemGroup>
 
+	<ItemGroup>
+	  <None Remove="LtmSampleFiles\ruffle-windows-override.ltm" />
+	  <None Remove="LtmSampleFiles\ruffle.ltm" />
+	</ItemGroup>
+
```
Masterjun3, line 35: Why are these lines needed when they aren't needed for all other files?

Randomno: IDE did that, I don't know

YoshiRulz: `<None/>` items are essentially hints to the IDE. The SDK automatically includes anything that's not a source file (`.cs`, `.resx`, etc.) in `@(None)` so they'll show up in the IDE's project tree view. Removing files like this hides them from the tree, but makes no difference to the compilation.

adelikat: just remove the lines, visual studio adds them but we don't want them

Randomno: Removed

adelikat (approved):

PR #1911, opened 2024-07-04
Add/Merge additional params on return urls

Masterjun3:
Resolves the bug where we would redirect to a page with two `?`, breaking the query.

To recreate the bug
- go to Submission Catalog
- click New next to Game to create a new game, enter some stuff, and Save
- you get redirected back, notice you're now on a url like `/Submissions/Catalog/7251?GameId=2356`
- click New next to Game Version to create a new version, enter some stuff, and Save
- you get redirected back again, notice you're now on a url like `/Submissions/Catalog/7251?GameId=2356?GameId=2356&GameVersionId=4820`

That's the bug, there are two `?` in that final link. Only the first `?` counts, so the query sets `GameId` to `2356?GameId=2356`.

~~After this PR the final link will look like `/Submissions/Catalog/7251?GameId=2356&GameId=2356&GameVersionId=4820`.
Notice how there are still two `GameId`s, but it works (and query string expliticly allow multiple settings of the same value).~~

After this PR the final link will look like `/Submissions/Catalog/7251?GameId=2356&GameVersionId=4820`.

### DIFF
--- a/TASVideos/Pages/BasePageModel.cs
+++ b/TASVideos/Pages/BasePageModel.cs
@@ -83,14 +83,27 @@ public class BasePageModel : PageModel
 			? BaseReturnUrlRedirect()
 			: Redirect(page);
 
-	protected IActionResult BaseReturnUrlRedirect(NameValueCollection? additionalParam = null)
+	protected IActionResult BaseReturnUrlRedirect(NameValueCollection? additionalParams = null)
 	{
 		var returnUrl = Request.ReturnUrl();
-		if (additionalParam is not null)
+
+		if (additionalParams is not null)
 		{
-			var returnUrlParams = HttpUtility.ParseQueryString(returnUrl);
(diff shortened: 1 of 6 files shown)


adelikat:
I like this approach, and the code looks good

vadosnaprimer:
I'll test locally and call back.

vadosnaprimer:
If I create game and version, they get properly selected. The baseline goal is created automatically, so I can just select it and save. But if I need to create a different goal and I hit Manage, game version gets unselected.

Masterjun3:
Technically the Manage button is unrelated to this PR. But I found the reason this happens. It's because our javascript doesn't encode the return url properly. I can make a commit that fixes that as well.

vadosnaprimer:
Thanks!

adelikat (approved):

PR #1917, opened 2024-07-12
Select Improver: Remove inital flashing of old select

Masterjun3:
This PR hides the original select by default, and uses the `<noscript>` tag to set the style back to visible, for noscripters.
This PR also improves performance, because the intial render doesn't render the select, as it is already hidden when the html comes from the server.

### DIFF
--- a/TASVideos/wwwroot/css/partials/_customizations.scss
+++ b/TASVideos/wwwroot/css/partials/_customizations.scss
@@ -718,6 +718,6 @@ top-button-bar {
 	}
 }
 
-.d-none-js {
+.d-none-except-noscript {
 	display: none !important;
 }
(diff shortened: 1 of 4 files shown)


TASVideos/wwwroot/css/partials/_customizations.scss
```diff
@@ -716,4 +716,8 @@ top-button-bar {
 			padding: 0.5em;
 		}
 	}
+}
+
+.d-none-js {
```
adelikat, line 721: This name confuses me.  Why can't we just put d-none on something, why do we need a new class?  If we do have this class, can it be better named to indicate when I should use it?  I would assume I would use it from javascript the one usage isn't javascript

Masterjun3: The name can be changed. The purpose of this class is to hide an element only when javascript is enabled.
We can't use the `d-none` by bootstrap, because we overwrite this new style in a later `<noscript>` block, and we don't want to override the bootstrap `d-none`, as that is used in many other places.

We could rename this to `d-js-none`, to align with the bootstrap naming, e.g. `d-sm-none` or `d-lg-none` to hide elements on small or large displays.

PR #1921, opened 2024-07-20
Improve hyperlinks in submissions table

YoshiRulz:


vadosnaprimer (approved):

Masterjun3:
I personally disagree with having the game link to the game page here.
I know it makes sense, but you're on a submission list, you want to see the submission. Other links are distracting.

In fact, other sites link to the specific purpose of a page, and only afterwards do they link to the more general page.
E.g. on GitHub if you're on the page listing commits and you click on a username, you *don't* go to their profile. Instead you go to all commits by them.

For me this doesn't "improve" this table at all. It just switches a bad design with a different bad design. I also expect a lot of people to misclick in the future with this change.

YoshiRulz:
This is separated into 2 commits so that you can drop the second. (I forgot to sign though, so I'll do that myself before merge.)

> E.g. on GitHub if you're on the page listing commits and you click on a username, you don't go to their profile. Instead you go to all commits by them.

OT, but this bothers me every time, I've never expected or wanted that behaviour. I guess they too were tempted to cram in as many features as they could, since only the username links there while the avatar is normal. The filter dropdown on the top-right makes it redundant though.

adelikat:
Rejecting this PR.  What we have is objectively not a good user experience and has been the way it is for about 12 years.  I think it adds negative value to change where the link goes that people have been using for so long.  And the end result isn't objectively better than what we have.  
I propose we make the full row clickable.  I think this requires js, but we have more freedom to use that with our current security headers, and it can gracefully fallback as well

PR #1935, opened 2024-08-10
Some refactors to `Builtins` (WikiEngine)

YoshiRulz:
From #1906.

### DIFF
--- a/TASVideos.WikiEngine/MakeBracketed.cs
+++ b/TASVideos.WikiEngine/MakeBracketed.cs
@@ -132,12 +132,12 @@ public static partial class Builtins
 		return text;
 	}
 
-	public static string NormalizeInternalLink(string input)
+	public static string NormalizeInternalLink(ReadOnlySpan<char> input)
 	{
-		var hashParts = input.Split('#');
-
-		var text = hashParts[0].TrimEnd('/');
-		var ss = text.Split('/');
+		var iAnchorSeparator = input.IndexOf('#');
+		var pathAndQuery = iAnchorSeparator < 0 ? input : input[..iAnchorSeparator];
+		var anchor = iAnchorSeparator < 0 ? "" : input[(iAnchorSeparator + 1)..];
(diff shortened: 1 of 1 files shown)


TASVideos.WikiEngine/MakeBracketed.cs
```diff
@@ -132,12 +132,12 @@ private static string NormalizeUrl(string text)
...
 	}
 
-	public static string NormalizeInternalLink(string input)
+	public static string NormalizeInternalLink(ReadOnlySpan<char> input)
 	{
-		var hashParts = input.Split('#');
-
-		var text = hashParts[0].TrimEnd('/');
-		var ss = text.Split('/');
+		var iAnchorSeparator = input.IndexOf('#');
```
adelikat, line 137: is this a typo? I think it is suppsoed to be isAnchorSeparator ?

TASVideos.WikiEngine/MakeBracketed.cs
```diff
@@ -258,17 +258,19 @@ internal static INode MakeLink(int charStart, int charEnd, string text, INode ch
 		return new Element(charStart, "a", attrs, [child]) { CharEnd = charEnd };
 	}
 
-	private static Element MakeImage(int charStart, int charEnd, string[] pp, int index, out bool unusedParams)
+	private static Element MakeImage(int charStart, int charEnd, IEnumerable<string> pp, out bool unusedParams)
 	{
 		unusedParams = false;
+		var iter = pp.GetEnumerator();
```
adelikat, line 264: I don't think downgrading this from array to IEnumerable is an improvement here.  It makes it more extensible, but this is only used in specific circumstances.  In those cases, I generally prefer to be specific with the types that are being used.  Using iterators, imo, hurts readability compared to a standard for loop, and I don't see any performance benefits

YoshiRulz: Dropped

adelikat (approved):

PR #1937, opened 2024-08-10
Make pub card title not a link when already on that page

YoshiRulz:


vadosnaprimer:
I was already not a fan of a link that's displayed as regular text, but when it also behaves differently makes it even more confusing.

Masterjun3:
I mean this is probably fine? (Btw. this code looks like it will work on `/1234M` but not `/Publications/View/1234`.)

I'd like to understand what purpose this change serves though.
To me it seems intuitive that clicking the title of the card brings you to a link for that card. We have the same behavior on the Forum where you can copy a quick link on every post by clicking its title.
image

Yes people can also copy the URL in their browser URl field, but why remove options?

YoshiRulz:
That's a good point, and one I hadn't considered. (I was only concerned with text selection, since holding Alt is an obscure feature and a bit fiddly.) This only adds complexity.

PR #1941, opened 2024-08-11
Use quotes for playground goal

Masterjun3:


Masterjun3 (commented):
Approved.
image
:(

vadosnaprimer (approved):
I can tho!

PR #1948, opened 2024-08-17
Add various UI stuff

Masterjun3:
"Biggest" part of this PR is adding the submission topic vote count that we had on the old site `75% (9/3/2)` (9 Yes, 3 Meh, 2 No).
I made a VoteCounts class for this. But I didn't know where to put it because it's both Forum Topic stuff and also Submission stuff. So I put it in Common. This can be moved wherever anyway.

So basically the PR adds this to the submission list,
image
and also to the forum topics.
image
<hr>

The entire username now leads to the "last post". No need for the username to lead to the user profile. This can be done in a second click because you're about to go to the user's post which has the profile link there.
image

<hr>
Claim now has an icon.

image
<hr>
Fixed some bugs on the Wiki ViewSource page. On current revisions it was missing the number. And it always said "current" even on non-current revisions.

image
<hr>
Added source links to revisions on Wiki PageHistory page, and also changed the coloring, because in dark mode this was not good:

image
<hr>
Changed footer to button links, because those fit more with our other buttons:

image

### DIFF
--- a/TASVideos.Common/VoteCounts.cs
+++ b/TASVideos.Common/VoteCounts.cs
@@ -1,0 +1,11 @@
+﻿namespace TASVideos.Common;
+
+public class VoteCounts
+{
+	public int VotesYes { get; init; }
+	public int VotesMeh { get; init; }
+	public int VotesNo { get; init; }
+	public bool UserVotedYes { get; init; }
+	public bool UserVotedMeh { get; init; }
+	public bool UserVotedNo { get; init; }
+}
--- a/TASVideos/Extensions/EntityExtensions.cs
+++ b/TASVideos/Extensions/EntityExtensions.cs
@@ -1,4 +1,5 @@
-﻿using TASVideos.Data.Entity.Awards;
+﻿using TASVideos.Common;
+using TASVideos.Data.Entity.Awards;
 using TASVideos.Data.Entity.Forum;
 using TASVideos.Data.Entity.Game;
 using TASVideos.Pages.Forum.Posts;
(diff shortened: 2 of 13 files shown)


vadosnaprimer (approved):

Masterjun3:
Questions were raised about the performance of the added submission topic vote counts query.

The query on Subs-List with *all statuses* and the default 100 page size takes on average:
- 0.100s before this PR.
- 0.110s after this PR.

A bigger difference shows when maximizing the pagesize, loading all my ~7000 submissions at once. There the query takes on average:
- 0.250s before this PR.
- 0.700s after this PR.

RetroEdit (commented):
To me, the links are cleaner than buttons. It semantically and visually distinguishes separates overall site info from the Referrers, which is specific to your current page.

Masterjun3:
The visual distinction could be done in a different way, e.g. by changing the location. This was previously discussed already.
In the end, at least we agreed that all buttons fits to the rest of our page in general.

A different position was already done in #1590 but eventually reverted. A discussion can be held about this. But in any case, this PR only converts the links into buttons.

Masterjun3:
So basically there are two reasons why I replaced the links with buttons.
1. It looks better visually because there was already the "List referrers" button, and that one couldn't be changed to a link because on wiki pages that button is next to other buttons.
2. On mobile, in a cluster of links, buttons give a better visual clarity that it can be clicked, and also makes it easier to tap in general.

The "List referrers" button not fitting with the other buttons is a valid argument. But @RetroEdit I would rather like you make a separate issue ticket for this where stuff can be discussed, instead of discussing in a merged PR. You can link to this discussion in the new issue though.

PR #1952, opened 2024-08-19
Add `[module:PublicationsByPlatform]`

### LINKED ISSUE #1910: Automate the per-platform movie table
https://tasvideos.org/Movies

First tab. We have to add the links manually every time a new platform appears.

Per-class links are not always there, but I think it makes sense to show them as long as they're not empty.

Stars are a flag now, but I'm not sure if it's worth removing it from that table yet. They work as a good way to promote cooler movies right away.

YoshiRulz:
resolves #1910

### DIFF
--- a/TASVideos/WikiModules/PublicationsByPlatform.cshtml
+++ b/TASVideos/WikiModules/PublicationsByPlatform.cshtml
@@ -1,0 +1,14 @@
+﻿@model PublicationsByPlatform
+<table>
+	@foreach (var platform in Model.Platforms.OrderBy(res => res.DisplayName))
+	{
+		<tr>
+			<td>@platform.DisplayName</td>
+			<td><a href="/Movies-@platform.Code">All Publications</a></td>
+			@foreach (var pubClass in Model.PubClasses.OrderBy(t => t.Id))
+			{
+				<td><a href="/Movies-@platform.Code-@pubClass.Link">@pubClass.Name</a></td>
+			}
+		</tr>
+	}
+</table>
--- a/TASVideos/WikiModules/PublicationsByPlatform.cshtml.cs
+++ b/TASVideos/WikiModules/PublicationsByPlatform.cshtml.cs
@@ -1,0 +1,50 @@
+﻿using TASVideos.WikiEngine;
+
+namespace TASVideos.WikiModules;
+
+[WikiModule(ModuleNames.PublicationsByPlatform)]
+public class PublicationsByPlatform(IGameSystemService platforms, IClassService classes) : WikiViewComponent
+{
+	public IReadOnlyList<(string DisplayName, string Code)> Platforms { get; private set; } = null!;
+
+	public IReadOnlyCollection<PublicationClass> PubClasses { get; set; } = null!;
+
+	public async Task<IViewComponentResult> InvokeAsync(IList<string> groupings)
+	{
+		var extant = (await platforms.GetAll()).ToList();
+		List<IReadOnlyList<SystemsResponse>> rows = [];
+		void ProcessGroup(string groupStr)
+		{
+			List<SystemsResponse> row = [];
+			foreach (var idStr in groupStr.Split('-'))
+			{
+				var found = extant.FirstOrDefault(sys => sys.Code.Equals(idStr, StringComparison.OrdinalIgnoreCase));
+				if (found is null)
+				{
+					// ignore, TODO log?
+					return;
+				}
+
+				extant.Remove(found);
+				row.Add(found);
+			}
+
+			rows.Add(row);
+		}
+
+		foreach (var groupStr in groupings)
+		{
+			ProcessGroup(groupStr);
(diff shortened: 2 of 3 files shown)


TASVideos/WikiModules/PublicationsByPlatform.cshtml
```diff
@@ -0,0 +1,13 @@
+﻿@model PublicationsByPlatform
+<table>
```
adelikat, line 2: consider using the standard-table tag (taghelper)

adelikat:
@YoshiRulz : how is this PR coming along? the code looks good

vadosnaprimer:
Turns out we already have modules that have grouping. Example

`[module:displaymovie|flag=verified|systemCode=gb,gbc,gba|obs]`

I believe the same should be done here.

TASVideos/WikiModules/PublicationsByPlatform.cshtml.cs
```diff
@@ -5,11 +5,18 @@ namespace TASVideos.WikiModules;
...
 {
-	public IReadOnlyList<SystemsResponse> Platforms { get; private set; } = null!;
+	public IReadOnlyList<(string DisplayName, string Code)> Platforms { get; private set; } = null!;
 
-	public async Task<IViewComponentResult> InvokeAsync()
+	public async Task<IViewComponentResult> InvokeAsync(IList<string> systemCodes)
 	{
-		Platforms = (await platforms.GetAll()).ToArray();
+		var extant = await platforms.GetAll();
+		Platforms = systemCodes.Select(s => s.Split('-').Select(id => extant.FirstOrDefault(sys => sys.Code.Equals(id, StringComparison.OrdinalIgnoreCase))).ToArray())
```
vadosnaprimer, line 13: Does this add grouping of some of the systems in the whole table, or does it just narrow down the table to only show the systems passed explicitly?

YoshiRulz: The list is drawn from the module's argument now, and the full list from `IGameSystemService` is only used to sanitise it.

vadosnaprimer: The point of the request was to automatically show all platforms with pubs without having to manually add them every time a new platform is published. Reducing displayed systems to only those explicitly passed defeats the purpose of the module.

YoshiRulz:
You can test with
```
[module:PublicationsByPlatform|groupings=A2600-A7800,Jaguar-JaguarCD,GB-SGB-GBC,DOS-DOOM-Windows-Linux,MSX-SVI3x8,NES-FDS,N64-N64DD,DS-DSi,Genesis-32X-SegaCD,SMS-GG-SG1000,SNES-BSX,PCE-PCECD-SGX]
```

TASVideos/WikiModules/PublicationsByPlatform.cshtml
```diff
@@ -0,0 +1,13 @@
...
+	@foreach (var platform in Model.Platforms.OrderBy(res => res.DisplayName))
+	{
+		<tr>
+			<td>@platform.DisplayName</td>
+			<td><a href="/Movies-@platform.Code">All Publications</a></td>
+			<td><a href="/Movies-@platform.Code-Standard">Standard</a></td>
+			<td><a href="/Movies-@platform.Code-Stars">Stars</a></td>
+			<td><a href="/Movies-@platform.Code-Alternative">Alternative</a></td>
```
vadosnaprimer, line 10: Since classes can be edited using site UI (and we recently added a new class), I suggest getting the list of existing classes and iterating though it, instead of hardcoding their then-names.

vadosnaprimer:
Grouping works!!!

Is it a big DB hit if only non-empty lists are shown?

vadosnaprimer (approved):

TASVideos/WikiModules/PublicationsByPlatform.cshtml.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+	public IReadOnlyList<(string DisplayName, string Code)> Platforms { get; private set; } = null!;
+
+	public IReadOnlyCollection<PublicationClass> PubClasses { get; set; } = null!;
+
+	public async Task<IViewComponentResult> InvokeAsync(IList<string> groupings)
+	{
+		var extant = (await platforms.GetAll()).ToList();
+		List<IReadOnlyList<SystemsResponse>> rows = [];
+		void ProcessGroup(string groupStr)
```
Masterjun3, line 16: Is this local function necessary here? It seems to be called only once, and it makes the entire method more difficult to parse and understand, because I have to keep going up and down to see what things are affected where.

YoshiRulz: It's that or `goto`, since C# doesn't have multi-`break`.

PR #1968, opened 2024-08-27
Set `$(EnableConfigurationBindingGenerator)` (reduces reflection)

YoshiRulz:


adelikat:
Does this really reduce reflection? I thought it was necessary if you build your own source generators

YoshiRulz:
Well I can confirm it generates *something*. [The docs](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8/runtime#configuration-binding-source-generator) say that it uses reflection when the Source Generator isn't enabled.

adelikat (approved):

PR #1969, opened 2024-08-27
Enable extra sanity checks on service provider

YoshiRulz:
My understanding is that these are checks for services initialised out of order, and out-of-the-box those would silently fail. In any case the checks currently pass on my machine.


Masterjun3:
According to docs, these are by default already true on Dev builds, and false on other builds.

adelikat:
So we already get this level of checking on dev builds, this is just for production builds.  I'm not sure we need these extra checks at this point.

PR #1975, opened 2024-09-01
Ability to Sync verify Submissions

adelikat:


Masterjun3:
I personally think some of the terms collide here. We have "Console Verify" that people already know, and now there is "Sync Verify". So when something is called "IsVerified" or a filter is called "Show Verified" I think it will cause confusion. It should probably always include the term "Sync".

Masterjun3:
Also, there are some console.logs in the javascript code, maybe they should be removed before merging.

PR #1994, opened 2024-09-27
Improve UX for users looking to replace the movie files in their submissions

YoshiRulz:
screenshot

### DIFF
--- a/TASVideos/Constants.cs
+++ b/TASVideos/Constants.cs
@@ -19,6 +19,7 @@ public static class SystemWiki
 	public const string ActivitySummary = "System/ActivitySummary";
 	public const string AvatarRequirements = "System/AvatarRequirements";
 	public const string ClassEditingHelp = "System/ClassEditingHelp";
+	public const string EditMovieHeader = "System/EditMovieHeader";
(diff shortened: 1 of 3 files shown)


vadosnaprimer:
If people cancel and resubmit to fix a minor mistake, they haven't read Submission Instructions. I think the link we need within the description of that Movie file field is https://tasvideos.org/SubmissionInstructions#UpdatingTheSubmissionFile

Maybe just say there "if you need to do a minor fix to the movie, [here's how](https://tasvideos.org/SubmissionInstructions#UpdatingTheSubmissionFile)"

vadosnaprimer (changes requested):
OK so please put an alert-info instead, at the top of submission edit page, and have a system wiki page whose text will be embedded there, similar to how Submissions/Submit embeds System/SubmitMovieHeader

YoshiRulz:
Sorry for the churn.

I copied what the Submit page had, is that all that's necessary?

vadosnaprimer (approved):
Works perfectly now!

изображение

adelikat (approved):

PR #2003, opened 2024-10-05
CSP round 3(?)

YoshiRulz:
Adds the `default-src` and `require-trusted-types-for` directives to pass Google's tester, plus the directives `font-src`, `frame-src`, and `img-src` since they'd otherwise fall back to `default-src` and that's now too strict. Also adds the unrelated `base-uri` and `form-action` directives just because it's a good idea.

### DIFF
--- a/TASVideos/Extensions/ApplicationBuilderExtensions.cs
+++ b/TASVideos/Extensions/ApplicationBuilderExtensions.cs
@@ -46,8 +46,9 @@ public static class ApplicationBuilderExtensions
 		});
 	}
 
-	public static IApplicationBuilder UseMvcWithOptions(this IApplicationBuilder app, IHostEnvironment env)
+	public static IApplicationBuilder UseMvcWithOptions(this IApplicationBuilder app, IHostEnvironment env, AppSettings settings)
 	{
+		var userAgentReportURL = $"{settings.BaseUrl}/Diagnostics/UserAgentInterventionReports";
 		string[] trustedJSHosts = [
 			"https://cdn.jsdelivr.net",
 			"https://cdnjs.cloudflare.com",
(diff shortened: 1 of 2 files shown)


TASVideos/Pages/Diagnostics/UserAgentInterventionReports.cshtml.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace TASVideos.Pages.Diagnostics;
+
+[RequirePermission(PermissionTo.SeeDiagnostics)]
+public class UserAgentInterventionReportsModel : BasePageModel
+{
+	public void OnPost()
```
adelikat, line 6: You don't need this method

TASVideos/Pages/Diagnostics/UserAgentInterventionReports.cshtml
```diff
@@ -0,0 +1,14 @@
+﻿@model UserAgentInterventionReportsModel
```
adelikat, line 1: Diagnostics is a good place to put this *thumbs up*

TASVideos/Pages/Diagnostics/UserAgentInterventionReports.cshtml.cs
YoshiRulz: Okay but how do I listen for POSTs though.

adelikat:
Oh I see, you want an endpoint that can be called that can then track requests and report statistics? How about not do that one right now, that's a huge lift and needs to be carefully planned as to where and how it will be stored and reported

TASVideos/Extensions/ApplicationBuilderExtensions.cs
```diff
@@ -48,6 +48,31 @@ public static IApplicationBuilder UseStaticFilesWithExtensionMapping(this IAppli
 
 	public static IApplicationBuilder UseMvcWithOptions(this IApplicationBuilder app, IHostEnvironment env)
 	{
+		const string UserAgentReportURL = "https://tasvideos.org/Diagnostics/UserAgentInterventionReports";
```
YoshiRulz, line 51: Ideally this would be `localhost` when running locally and `tasvideos.org` in prod. Is there a better way to do that than checking `env.IsProduction()`?

adelikat:
inject an AppSettings object and there is a BaseUrl property in there

adelikat (approved):

Issue #1878, opened 2024-06-08
Ability to open site search page from middle-click


vadosnaprimer:
Search is a button that brings you to the search page, and if you've entered anything when hitting it, it will show results. But sometimes search needs to be done in a separate tab, and there's no way to open it as a new tab by middle-clicking the search button.

Randomno:
https://github.com/Randomno/tasvideos/tree/middleclick

Don't want to PR this but it works

adelikat:
Doing this via javascript is the wrong way to go.  I'm not sure the right way though, it's a form button due to the text box, you can't middle click that into a new tab, it isn't how HTML works

PR #2004, opened 2024-10-05
Allow middle-clicking the search button with progressive enhancement

YoshiRulz:
Implements #1878.
RetroEdit's earlier draft used JS to add an event handler specifically for MMB. I feel that's the wrong approach, and it should instead use the browser's built-in handling for `<a/>`.
This changeset uses JS to replace the `<button/>` with an `<a/>`, then adds an event handler for `onclick` (which [is a misnomer](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) and is not specific to LMB) which restores the `<button/>`-like submit behaviour.

### DIFF
--- a/TASVideos/wwwroot/js/site-head.js
+++ b/TASVideos/wwwroot/js/site-head.js
@@ -64,6 +64,22 @@ window.addEventListener("DOMContentLoaded", function () {
             rate(btn.dataset.pubId, true);
         };
     });
+
+    // allow middle-clicking the search button to open in new tab
+    const searchBtnElem = document.querySelector(/*ul.navbar-nav */'form[action="/Search/Index"] button[type="submit"]');
+    // is there a function for this? reverse Element.querySelector?
+    let searchFormElem = searchBtnElem.parentElement;
+    while (!(searchFormElem instanceof HTMLFormElement)) searchFormElem = searchFormElem.parentElement;
+    const linkElem = document.createElement("a");
+    Array.from(searchBtnElem.attributes).forEach(attr => linkElem.setAttribute(attr.name, attr.value));
+    linkElem.onclick = e => {
+        e.preventDefault();
+        searchFormElem.requestSubmit();
+    };
+    linkElem.removeAttribute("type");
+    linkElem.setAttribute("href", "/Search/Index");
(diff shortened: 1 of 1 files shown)


TASVideos/wwwroot/js/site-head.js
```diff
@@ -64,6 +64,22 @@ window.addEventListener("DOMContentLoaded", function () {
             rate(btn.dataset.pubId, true);
         };
     });
+
+    // allow middle-clicking the search button to open in new tab
+    const searchBtnElem = document.querySelector(/*ul.navbar-nav */'form[action="/Search/Index"] button[type="submit"]');
+    // is there a function for this? reverse Element.querySelector?
+    let searchFormElem = searchBtnElem.parentElement;
```
adelikat, line 71: what happens if there is not a searchBtnElem on the page? this would be a javascript error.  Also this runs on every page load.  This needs to be fast, I'm not sure that query selector is.  Adding an id would be okay I think.

I also don't like this approach, or possibly any approach, the complexity vs value I think is questionable

YoshiRulz: Is the search bar not on every page? I could easily handle that case if not.

TASVideos/wwwroot/js/site-head.js
```diff
@@ -64,6 +64,22 @@ window.addEventListener("DOMContentLoaded", function () {
             rate(btn.dataset.pubId, true);
         };
     });
+
+    // allow middle-clicking the search button to open in new tab
+    const searchBtnElem = document.querySelector(/*ul.navbar-nav */'form[action="/Search/Index"] button[type="submit"]');
+    // is there a function for this? reverse Element.querySelector?
```
nattthebear, line 70: `Element.prototype.closest`

YoshiRulz:
I just remembered the search has a char limit, which isn't currently enforced client-side but could easily be added to this click handler.

adelikat:
Took inspiration from this PR and implemented in a simpler way, closing

YoshiRulz:
https://github.com/TASVideos/tasvideos/commit/965eb787db36ce5ccc16544f835c436bce5abd10

PR #2009, opened 2024-10-15
Set up `Content-Security-Policy` "reporting" (monitoring)

YoshiRulz:
Follow-up to #2003. For an overview of reporting, see [Google's web.dev](https://web.dev/articles/csp#report-policy-violations) and [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#violation_reporting).

What's missing at the moment is all the handling for POST requests to `/Diagnostics/UserAgentInterventionReports`.

### DIFF
--- a/TASVideos/Extensions/ApplicationBuilderExtensions.cs
+++ b/TASVideos/Extensions/ApplicationBuilderExtensions.cs
@@ -1,4 +1,6 @@
-﻿using Microsoft.AspNetCore.StaticFiles;
+﻿#define SHOULD_INCLUDE_STAGING_CSP
+
+using Microsoft.AspNetCore.StaticFiles;
 using Microsoft.Extensions.FileProviders;
 using TASVideos.Core.Settings;
 using TASVideos.Middleware;
@@ -48,6 +50,7 @@ public static class ApplicationBuilderExtensions
 
 	public static IApplicationBuilder UseMvcWithOptions(this IApplicationBuilder app, IHostEnvironment env, AppSettings settings)
 	{
+		var userAgentReportUrl = $"{settings.BaseUrl}/Diagnostics/UserAgentInterventionReports";
 		string[] trustedJsHosts = [
 			"https://cdn.jsdelivr.net",
 			"https://cdnjs.cloudflare.com",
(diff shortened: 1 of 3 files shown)


TASVideos/Extensions/ApplicationBuilderExtensions.cs
```diff
@@ -63,12 +66,26 @@ public static IApplicationBuilder UseMvcWithOptions(this IApplicationBuilder app
...
+			"report-to for-csp", // browsers using CSP Level 3+ will look up this key in the `Reporting-Endpoints` header and use that URI
+			$"report-uri {userAgentReportUrl}?kind=csp&csp-lvl=lteq2", // browsers from before CSP Level 3 will use this
+			*/
 			"require-trusted-types-for 'script'", // experimental, but Google seems to be pushing it: should block `HTMLScriptElement.innerHTML = "user.pwn();";`, and similarly block adding in-line scripts as attrs
 			$"script-src 'self' {string.Join(' ', trustedJsHosts)}", // `<script/>`s will be blocked unless they're from one of these domains
 			"style-src 'unsafe-inline' 'self' https://cdnjs.cloudflare.com/ajax/libs/font-awesome/", // allow `<style/>`, and `<link rel="stylesheet"/>` if it's from our domain or trusted CDN
 			"upgrade-insecure-requests", // browser should automagically replace links to any `http://tasvideos.org/...` URL (in UGC, for example) with HTTPS
 		];
 		var contentSecurityPolicyValue = string.Join("; ", cspDirectives);
+#if SHOULD_INCLUDE_STAGING_CSP
```
adelikat, line 79: Don't do these.  Is this supposed to be for different environments? If so, the env variable has what you need

YoshiRulz: No this is to skip sending a duplicate header.

Masterjun3: If you want behavior that's easily togglable, we use the AppSettings object (which is conveniently already here) with configuration in the appsettings.config file. Using that also means we don't have to recompile the project for new behaviors.

adelikat (changes requested):
Changing to requested changes for using appsettings rather than a compiler directive

YoshiRulz:
Bruh I removed it.

Masterjun3:
I don't think CSP reporting is something worth implementing currently.
We don't have a setup of anyone watching errors or anything anyway.
And this draft hasn't been touched for 6 months now, so I'm closing it.

PR #2019, opened 2024-10-20
Fix missing images everywhere when running locally

YoshiRulz:
Before/after:
side-by-side
Only has effect in dev environments.

### DIFF
--- a/TASVideos/Pages/Wiki/Render.cshtml.cs
+++ b/TASVideos/Pages/Wiki/Render.cshtml.cs
@@ -4,7 +4,12 @@ using TASVideos.Core.Services.Wiki;
 namespace TASVideos.Pages.Wiki;
 
 [AllowAnonymous]
-public class RenderModel(IWikiPages wikiPages, ApplicationDbContext db, ILogger<RenderModel> logger) : BasePageModel
+public class RenderModel(
+	IWikiPages wikiPages,
+	ApplicationDbContext db,
+	IHostEnvironment env,
+	ILogger<RenderModel> logger
+) : BasePageModel
 {
 	public IWikiPage WikiPage { get; set; } = null!;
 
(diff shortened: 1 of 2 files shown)


YoshiRulz:
Build failure is here:
https://github.com/TASVideos/tasvideos/blob/d72b31b1f173d2965a6c4c28170154f546330643/tests/TASVideos.Test/Pages/Wiki/RenderModelTests.cs#L18
I'm not sure how to get an `IHostEnvironment` at that point. DI via a constructor parameter prevents the test from running.

Masterjun3:
Is this the right place to insert a dev fallback? It relies on the broken links to be redirected to the wiki endpoint.
Instead, couldn't this be implemented more properly in the options when we add the `UseStaticFiles`. Maybe by adding a dev FileProvider that uses the regular file provider and doing the fallback logic there?

adelikat:
I looked and I see no way to do this kind of routing with UseStaticFiles.  The best I found is to put some custom middleware, but thsi would mean that we would either always have the dummy images locally (which i do not want, since I have the correct ones) or the middleware would need to do some kind of check if the image exists, which we get for free with this approach.  This might be a bit hacky but I don't see a less hacky alternative

Masterjun3:
@adelikat I mentioned doing it by overriding the StaticFileOptions.FileProvider in the UseStaticFiles call with our own IFileProvider implementation that does this.
Wouldn't that be a good way to do this?

adelikat:
I would need an example.  I tried to do that approach and it seemed impossible

Masterjun3:
I can give an example in a week if that's okay.

Masterjun3:
@adelikat I posted my example in PR #2062 .

adelikat:
Closing in favor of this approach: #2062

Issue #1658, opened 2023-08-22
News Feed should have a character limit followed by "read more" to go to the actual post.


Spikestuff:
Like what we have for emulation releases are completely fine.
![image](https://github.com/TASVideos/tasvideos/assets/12412764/ff257a85-d530-4b32-a9da-eeaf37137718)

But when it's an actual news topic. It goes on, and it might be better to move that elsewhere than have that entire, small column hold the info.
![image](https://github.com/TASVideos/tasvideos/assets/12412764/0f7b0617-aaf9-400f-abd7-9b19aa81d8d8)

If there was a way to detect a character limit such as using the example above 800 characters before breaking off to a read more, or for a way to detect that it's more than one paragraph, with much more information then breaking it off might look more visually appealing.
![image](https://github.com/TASVideos/tasvideos/assets/12412764/c326369e-cc21-4f30-af77-42249db811b2)

Also you can't click on the link provided without a middle/right click in place (probably by design, but a way to click that would be nice as well).

PR #2020, opened 2024-10-21
Make YouTube, Niconico, etc. embeds load after other content

YoshiRulz:
Pages should appear to load faster, especially on slow browsers/connections, and especially on the front page because there's currently a YouTube player in the news feed (see also #1658), which is in the middle of the page DOM-wise.


Masterjun3 (approved):

Masterjun3:
Oh wow we didn't have that until now?

PR #2022, opened 2024-10-26
Reduce Identity-Cookie size by not putting Roles in it

Masterjun3:
Our access system is based on Permissions, and we add the Permission Claims manually to the Principal (and thus the identity cookie). This means we have no need for the Roles inside it.

The `services.AddIdentity<User, Role>` would add `UserClaimsPrincipalFactory<User, Role>` to the services, but we pre-empt that by putting the "less powerful" base class `UserClaimsPrincipalFactory<User>` to the services.
These are for creating a ClaimsPrincipal, and the original one would add the Roles and their Claims to the Principal.
Instead, the one we now use only adds the usual Claims, like Name and Email.

And this is exactly what we want.


Invariel (approved):

YoshiRulz:
Should mention that reducing the cookie size means reducing the size of the HTTP headers returned on login. Which is the problem that prompted this work.

adelikat (approved):

PR #2038, opened 2024-11-14
[WIP] Output assembly trimming

YoshiRulz:
A big contributor to this untrimmability is `JsonSerializer`. Switching to the Source Generator for that should help.

See also https://github.com/dotnet/efcore/issues/21894.


adelikat:
We want to use JsonSerializer and reflection.  Also trimming isn't advantageous with the way we deploy the site, so I do not think this approach adds value

PR #2039, opened 2024-11-15
Set ST-V movies as Arcade

CasualPokePlayer:
Similar case as NES VS movies, except with Saturn with its ST-V Arcade variant.


vadosnaprimer (approved):

adelikat:
can i get some unit tests for this? or at least an example bk2 so I can do it

CasualPokePlayer:
> can i get some unit tests for this? or at least an example bk2 so I can do it

[diehard.zip](https://github.com/user-attachments/files/17788181/diehard.zip)

PR #2044, opened 2024-11-23
Add more entries to `robots.txt`

YoshiRulz:
It should be apparent why each of these isn't useful to crawl.
- Might want to keep `/Wiki/ViewSource` for archival purposes?
- Might want to block `/Forum/Posts/User/` for humans who aren't logged in?
- The false positive thing: [spec](https://www.rfc-editor.org/rfc/rfc9309.html#name-the-allow-and-disallow-line), [Google's nicer docs](https://developers.google.com/search/docs/crawling-indexing/robots/robots_txt#url-matching-based-on-path-values)

### DIFF
--- a/TASVideos/Middleware/RobotHandlingMiddleware.cs
+++ b/TASVideos/Middleware/RobotHandlingMiddleware.cs
@@ -11,15 +11,11 @@ public class RobotHandlingMiddleware(RequestDelegate request, IHostEnvironment e
 
 		if (env.IsProduction())
 		{
-			// TODO the format for this was recently codified to basically match what Google was using, anyway each entry has an implicit trailing wildcard so there's a small chance of false positives
 			sb.AppendLine("""
 						User-agent: *
 						Disallow: /Movies-
 						Disallow: /MovieMaintenanceLog
 						Disallow: /UserMaintenanceLog
-						Disallow: /Account/
-						Disallow: /Forum/Posts/User/
-						Disallow: /Forum/Topics/Create/
 						Disallow: /InternalSystem/
 						Disallow: /Search
 						Disallow: /*?revision=*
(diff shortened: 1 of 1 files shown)


Masterjun3:
- Why disallow `/Account/`? It makes sense to me that people would google "tasvideos register" to find how to register an account.
- And `/Forum/Topics/Create/` needs authorization so we don't need to block it here. Otherwise I wonder why you don't also block wiki editing, post creation, userfile creation, and all the other creation stuff. Seems too much.
- Why block `/Forum/Posts/User/` ?
- I don't understand the TODO. We know each entry has an implicit trailing wildcard, that's why we can block `/Movies-` e.g.

YoshiRulz:
1. Oh right, I'll revert that.
1. It was in a list of highest-trafficked pages you shared recently. I think those others should be blocked, but if you don't want them listed here, I did have a solution using `rel="nofollow"` in another branch.
1. This was also in said list. It's not really a useful thing to index: crawling `/Forum/Topics/{id}` will include the posters' names with their posts, so in theory they should be searchable that way if someone wanted. But I think searching by poster is prone to abuse, hence my suggestion in OP.
1. The "correct" way is to put a `$` after the path fragment (and then if there are subpages, include a separate `.../*` entry). I haven't bothered.

adelikat:
2. I would recommend not listing this or any page that is auth required as it is not necessary and there are too many others. Robots will not be logged in.  Also, I think usage statistics doesn't need to play a high role here. If it makes sense for something to be crawled, it should be crawled.  If it needs to be more performant as a result, we can address that
4. Then at least put that information in the comment, so that someone who isn't an expert in robot crawling can do the comment.  But I would recommend not bothering with it

Masterjun3 (approved):
I have commited some removals as discussed, so that this PR can be merged.

adelikat (approved):

PR #2047, opened 2024-11-26
Increase allowed submission sync notes length

Masterjun3:
With all the file hashes they can become quite large.


vadosnaprimer (approved):

Masterjun3:
This is not the correct field that had a problem. My bad.

PR #2062, opened 2025-01-03
Fix missing images everywhere when running locally V2

Masterjun3:
This copies what #2019 does, as an example for https://github.com/TASVideos/tasvideos/pull/2019#issuecomment-2564843036 .
It was easier to show this example using a PR.

### DIFF
--- a/TASVideos/Services/DevFallbackFileProvider.cs
+++ b/TASVideos/Services/DevFallbackFileProvider.cs
@@ -1,0 +1,28 @@
+﻿using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Primitives;
+
+namespace TASVideos.Services;
+
+public class DevFallbackFileProvider(IFileProvider originalProvider) : IFileProvider
+{
+	public IDirectoryContents GetDirectoryContents(string subpath)
+	{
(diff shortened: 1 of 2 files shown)


adelikat (approved):

TASVideos/Services/DevFallbackFileProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+	{
+		return originalProvider.GetDirectoryContents(subpath);
+	}
+
+	public IFileInfo GetFileInfo(string subpath)
+	{
+		IFileInfo fileInfo = originalProvider.GetFileInfo(subpath);
+		if (!fileInfo.Exists && subpath.StartsWith("/media/"))
+		{
+			return originalProvider.GetFileInfo("/images/tasvideos_rss.png");
```
YoshiRulz, line 18: Memoise?

Masterjun3: Do you mean we should cache the result of the hard-coded fallback file to avoid calling the method again and again?
To that I answer: Wouldn't caching be the job of the provider's GetFileInfo? If it doesn't do it, why should we? It would certainly cause issues if people would manually move around image files or replace them.

Issue #2068, opened 2025-01-09
Add a staff-only system page with a list of banned users


moozooh:
Banned users currently don't have a role, and hence cannot be tracked via the [role list](https://tasvideos.org/Roles/List) or, presumably, any other means. This is somewhat inconvenient because it requires staff to remember individual usernames of problematic users and track their forum profiles down to restore other forgotten information. Especially inconvenient for e.g. newer moderators who may not have known about these users in the first place, having to rely on other staff members to tell their story.

Since the site already tracks banned users' status internally, I imagine this could all be made visible on a separate, **staff-only page**. The ideal layout would be a table with:
- **username**, 
- **ban date**, 
- **ban expiry date** (where applicable), 
- **moderator comments** (where available) to see the reason for the ban.
 
Even better if it's sortable by any of these columns.

Invariel:
Genuine question: in the spirit of transparency, why restrict it to staff only?  An incredibly common question (in other spaces) is, "Why is/was X banned?"

YoshiRulz:
(I'm working on this.)

Well there's a `ModeratorComments` field but no `BanReason`. Would you want the `ModeratorComments` to be public? If not then it wouldn't have any benefit over going to a user's profile.

bigbass1997:
Moderator comments rarely include detailed reasoning for the ban. If they are used at all, they'll possibly contain some quick notes that other mods can see, such as
* When a ban was initiated
* Length of temp bans
* Behavior to watch out for (even if they aren't currently banned)
* A short sentence or phrase summarizing why they were banned (rare; even in severe cases)

The entire purpose of these notes is to be viewable only by other moderators.

bigbass1997:
> Can we make it a wiki module instead of a page?

Think that would work just as well yeah.

moozooh:
No objections from me, either. Just as long as we can see them all in one place like we can with Discord bans, for example.

moozooh:
> Genuine question: in the spirit of transparency, why restrict it to staff only? An incredibly common question (in other spaces) is, "Why is/was X banned?"

I don't really have a good answer ready for this, other than, perhaps, that it _may_ occasionally expose something sensitive, which we may not want and will have no ability to selectively hide. I see no problem answering such questions directly when asked; whether all of that information should be public at all times is a different question entirely. Perhaps we should discuss this. But for the record, currently, we have pages such as [Permissions](https://tasvideos.org/Permissions) only available to logged-in staff as well.

[...]

PR #2069, opened 2025-01-10
Add `[module:bannedusers]`

YoshiRulz:
Working towards #2068, which from the discussion there and on Discord we agreed should be a module which conditionally renders based on your permissions.

Currently functional (edit: I've since added pagination and sorting):
screencap

The module doesn't render anything unless you have the "Assign Roles" permission, which I believe is what lets you ban users?
And the final column with the moderator comments is only included if you also have the "Edit Users" permission, which it seems is the same people as have "Assign Roles" at the moment.

There's no ban start date field so I didn't add that.

### DIFF
--- a/TASVideos.Data/Entity/User.cs
+++ b/TASVideos.Data/Entity/User.cs
@@ -207,6 +207,9 @@ public static class UserExtensions
 	public static IQueryable<User> ThatHaveCustomLocale(this IQueryable<User> query)
 		=> query.Where(u => u.DateFormat != UserDateFormat.Auto || u.TimeFormat != UserTimeFormat.Auto || u.DecimalFormat != UserDecimalFormat.Auto);
 
+	public static IQueryable<User> ThatAreBanned(this IQueryable<User> query)
+		=> query.Where(u => u.BannedUntil.HasValue && u.BannedUntil > DateTime.UtcNow); // > and < in these methods, but what about ==?
+
 	public static IQueryable<User> ThatAreNotBanned(this IQueryable<User> query)
 		=> query.Where(u => !u.BannedUntil.HasValue || u.BannedUntil < DateTime.UtcNow);
 
--- a/TASVideos/WikiModules/BannedUsers.cshtml
+++ b/TASVideos/WikiModules/BannedUsers.cshtml
@@ -1,0 +1,25 @@
+﻿@model BannedUsers
+
+<partial name="_Pager" model="Model.Users" />
+<standard-table condition="User.Has(PermissionTo.AssignRoles)">
+	<sortable-table-head sorting="@Model.GetPaging()" model-type="typeof(BannedUsers.Entry)" page-override="@Model.CurrentPage" />
+	@foreach (var banned in Model.Users)
+	{
+		<tr>
+			<td><profile-link username="@banned.Name"></profile-link></td>
+			<td><timezone-convert asp-for="@banned.BannedUntil" /></td>
+			<td>
+				@if (banned.LastLoggedIn is not null)
+				{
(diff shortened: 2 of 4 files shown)


TASVideos.Data/Entity/User.cs
```diff
@@ -207,6 +207,9 @@ public static IQueryable<User> ForUser(this IQueryable<User> query, string? user
 	public static IQueryable<User> ThatHaveCustomLocale(this IQueryable<User> query)
 		=> query.Where(u => u.DateFormat != UserDateFormat.Auto || u.TimeFormat != UserTimeFormat.Auto || u.DecimalFormat != UserDecimalFormat.Auto);
 
+	public static IQueryable<User> ThatAreBanned(this IQueryable<User> query)
+		=> query.Where(u => u.BannedUntil.HasValue && u.BannedUntil > DateTime.UtcNow); // > and < in these methods, but what about ==?
```
adelikat, line 211: I don't understand the comment here, if you are asking how something works, would be best not to do it in code but in a discussion

TASVideos/WikiModules/BannedUsers.cshtml
```diff
@@ -0,0 +1,30 @@
+﻿@model BannedUsers
+
+<standard-table condition="User.Has(PermissionTo.AssignRoles)">
+	@*
```
adelikat, line 4: The query needs to run SortedPageOf() instead of ToList(), then this sortable table head should "just work"

TASVideos.Data/Entity/User.cs
YoshiRulz: Just an edge case I noticed in the code below this. Since it's timestamp-based, if it did ever happen, you could just reload the page.

TASVideos/WikiModules/BannedUsers.cshtml
YoshiRulz: It did just work for pagination but not sorting.

TASVideos/WikiModules/BannedUsers.cshtml.cs
```diff
@@ -0,0 +1,25 @@
...
+			.Select(u => new Entry(
+				u.UserName,
+				u.BannedUntil!.Value,
+				u.LastLoggedInTimeStamp,
+				u.ModeratorComments))
+			.SortedPageOf(GetPaging());
+		return View();
+	}
+
+	public record Entry(string Name, DateTime BannedUntil, DateTime? LastLoggedIn, string? ModeratorComments);
```
adelikat, line 24: to get sorting, you have to put the sortable attribute on the properties of this Entry type.  Which means it probably will need to be a class with init properties instead of record

YoshiRulz: I'm only getting "The LINQ expression ... could not be translated."

adelikat: did you make it a class? Records can't be used with the sorting and paging, for some spooky Entity Framework reasons

TASVideos/WikiModules/BannedUsers.cshtml.cs
```diff
@@ -12,14 +12,28 @@ public async Task<IViewComponentResult> InvokeAsync()
...
+		[Sortable]
+		public required string Name { get; init; }
+
+		[Sortable]
+		public required DateTime BannedUntil { get; init; }
+
+		[Sortable]
+		public required DateTime? LastLoggedIn { get; init; }
+
+		public required string? ModeratorComments { get; init; }
```
Masterjun3, line 37: Why `required`?

adelikat (approved):
This is still listed as draft, do you intend more changes? I think the PR is good as is

PR #2079, opened 2025-02-17
Enable Discord embeds for post and topic creations

Masterjun3:
This PR builds upon PR #2077 (and I will rebase once that PR is merged).
It removes the `<` and `>` from post and topic creation links on the discord distributor, meaning they will get an embed.

This is probably a controversal change, because embeds make things much more verbose and maybe cluttered. The upside is that you can read the beginning of posts without even clicking on it.

This is a draft for 2 reasons. Firstly the other PR should be merged, otherwise this one doesn't make sense. And secondly I want to test some things with the embeds. Maybe we show too much info. Maybe there are too many newlines.


Masterjun3:
The embeds are quite big, and I don't like this idea anymore. I'm still trying to get user content into the discord distributor, but I'm at a dead end here, so closing this PR.

PR #2082, opened 2025-02-18
Add fallback for pagination when `Sort` queryparam doesn't match any props

YoshiRulz:
Resolves (some of?) the "The query uses a row limiting operator ('Skip'/'Take') without an 'OrderBy' operator. This may lead to unpredictable results." warnings Masterjun's been investigating.

### DIFF
--- a/TASVideos.Core/Paging/Paginator.cs
+++ b/TASVideos.Core/Paging/Paginator.cs
@@ -94,7 +94,8 @@ public static class Paginator
 	{
 		if (string.IsNullOrWhiteSpace(request?.Sort))
 		{
-			return source0;
+			// per below, can't noop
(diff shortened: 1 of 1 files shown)


Masterjun3 (approved):
This works and reduces some of the warnings we get in our logs.
But I'm not a fan of the usage of "ReferenceEquals" because at no point is there any assurance that SortByParam will always overwrite the query. It's just not good design for future people working on the code, also proven by the fact a comment had to be added to explain what it's even doing.
A boolean variable "wasSorted" as a out parameter or similar would be simple to understand, would need no comments, and would help future people know what they have to be careful of.

YoshiRulz:
I'd argue that a single check at the end is better for maintainability, but there you go.

PR #2087, opened 2025-02-24
Allow zipless Submissions

### LINKED ISSUE #1187: Remove the pre-zipping step from submission upload
It's pointless wasted time for the submitter.  They should just upload their movie file.  We can use a similar method to what we use in userfiles to keep upload bandwidth usage down and upload size below the server's HTTP POST limit.

Masterjun3:
Resolves #1187 .

Previously, users had to compress their submission files into .zip files to submit movies to the site.
With this PR, submissions and submission file replacements don't have to be in the .zip format anymore.

This PR does not change the internal representation of files. The PR still manually zips the files to store them in the db. This can be changed in a later PR, because this change is already big enough.

We avoid the massive filesizes of text-based movie files by doing a client compression step in JS. We already did this in userfiles. However, unlike userfiles, we want to be backwards-compatible for now, meaning .zip files will still be accepted and properly handled like before.
For non-JS users, this PR also handled uncompressed files.

<hr>

**Does not require .zip files:**
- Submission files
- Submission replacement files

**Still requires .zip files:**
- Publication file replacements
- Publication additional movie files

Why do publications still need to be .zip files? It's because the filename of the upload is taken from the uploaded .zip file. Without a .zip file we would need some other way to enter the filename. For submissions the filename is always `submission[id].zip`.

### DIFF
--- a/TASVideos/Extensions/FormFileExtensions.cs
+++ b/TASVideos/Extensions/FormFileExtensions.cs
@@ -1,4 +1,5 @@
-﻿using System.Runtime.CompilerServices;
+﻿using System.IO.Compression;
+using System.Runtime.CompilerServices;
 using System.Security.Claims;
 
 using Microsoft.AspNetCore.Mvc.ModelBinding;
@@ -55,7 +56,7 @@ public static class FormFileExtensions
 	{
 		if (!user.Has(PermissionTo.OverrideSubmissionConstraints) && movie.Length >= SiteGlobalConstants.MaximumMovieSize)
 		{
-			modelState.AddModelError(movieFieldName, ".zip is too big, are you sure this is a valid movie file?");
+			modelState.AddModelError(movieFieldName, "File is too big, are you sure this is a valid movie file?");
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


TASVideos/Extensions/FormFileExtensions.cs
```diff
@@ -55,7 +56,7 @@ public static void AddModelErrorIfOverSizeLimit(
 	{
 		if (!user.Has(PermissionTo.OverrideSubmissionConstraints) && movie.Length >= SiteGlobalConstants.MaximumMovieSize)
```
adelikat, line 57: This maximum movie size is going to be a problem for noscript users trying to upload text based formats like .fm2.  Not sure what to do about it though

adelikat (approved):

TASVideos/Extensions/FormFileExtensions.cs
Masterjun3: I mean if they're well-versed enough in technology that they care about disabling JS, they should be able to .zip a file before upload.

vadosnaprimer (approved):

PR #2089, opened 2025-03-04
Decade TASer, Awards 2024 and Beyond, Awards Folder Restructuring, RSS icon update.

Spikestuff:
We've switched to using a new consistent font being "Inter Tight" information can be found here: https://fonts.google.com/specimen/Inter+Tight
So it's missing the License bit I believe feos was looking into setting up.
This only corresponds to 2024's Awards for now, if I or someone else decides to fix up the previous years.

Awards has been restructured to be based on year instead of just dumped into a single folder.
**This will break how files are currently sorted due to this.**
"xx" files were also deleted, this was maintained on the site side by feos already, preferring the "xxxx" files instead.

All templates should be provided as well. Might be incomplete based on a few things if the site goes down that path.

Also rss image finally got updated. :)


Masterjun3:
I have adjusted the code with the new paths.

adelikat (approved):

PR #2093, opened 2025-03-08
Missing Awards 2024 and Note Amendments.

Spikestuff:
Some awards went missing due to automation not grabbing 'em didn't realize.
Except Arcade-- I just outright missed that one.


Masterjun3 (approved):

vadosnaprimer:
Do we want fonts to be committed to the repo? I haven't fully read their licenses yet
https://fontstruct.com/fontstructions/show/1217225/fceux (just CC-BY)
https://fonts.google.com/specimen/Inter+Tight/license (?)

Masterjun3:
Using fonts doesn't require us to mention anything on the site.
Font license is only important if we redistribute the files (or if we design a new font based on the font, which we don't). The licenses allow redistribution and the necessary license text is always inside the font file itself.

So you can simply commit the font files to the repo if you want.

vadosnaprimer (approved):

vadosnaprimer (approved):

Issue #2060, opened 2024-12-27
RSS feeds don't pass W3C's validator


YoshiRulz:
https://www.rssboard.org/rss-validator/

~~Doesn't seem to cause any problems, so labelling this as an enhancement.~~ Apparently some of the misformattings did cause problems in practice. Most of the initial warnings have been dealt with, but I'll shoot for a third round of fixes now that it's redeployed and I can re-run the validator.

YoshiRulz:
[`<media:content/>`](https://www.rssboard.org/media-rss#media-content) should have a `url` attribute (for direct download URL) or a `<media:player/>` child, not both.

PR #2097, opened 2025-03-12
More RSS fixes

YoshiRulz:
For #2060. I believe there are a couple of problems left, specifically re: links in the item bodies, but I'm not sure how to proceed on those.

### DIFF
--- a/TASVideos/Pages/RssFeeds/Submissions.cshtml
+++ b/TASVideos/Pages/RssFeeds/Submissions.cshtml
@@ -26,7 +26,7 @@
 				<comments condition="@sub.TopicId.HasValue">@(Settings.BaseUrl)/Forum/Topics/@sub.TopicId</comments>
 				<enclosure url="@(subUrl)?handler=Download" />
 				<guid>@subUrl</guid>
-				<pubDate>@sub.CreateTimestamp.ToString("u", CultureInfo.InvariantCulture)</pubDate>
+				<pubDate>@sub.CreateTimestamp.ToString("r", CultureInfo.InvariantCulture)</pubDate>
 			</item>
 		}
 	</channel>
--- a/TASVideos/Pages/RssFeeds/Wiki.cshtml
+++ b/TASVideos/Pages/RssFeeds/Wiki.cshtml
@@ -20,7 +20,7 @@
 				<rss-link>@url</rss-link>
 				<description>@edit.RevisionMessage</description>
 				<comments>@(Settings.BaseUrl)/Forum/</comments>
-				<guid>@url</guid>
+				<guid>@url?revision=@edit.RevisionId</guid>
 				<pubDate>@edit.PubDate.ToString("r", CultureInfo.InvariantCulture)</pubDate>
 			</item>
 		}
--- a/TASVideos/Pages/RssFeeds/Wiki.cshtml
+++ b/TASVideos/Pages/RssFeeds/Wiki.cshtml
@@ -21,7 +21,7 @@
 				<description>@edit.RevisionMessage</description>
 				<comments>@(Settings.BaseUrl)/Forum/</comments>
 				<guid>@url</guid>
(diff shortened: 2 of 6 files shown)


TASVideos/Pages/RssFeeds/Submissions.cshtml
```diff
@@ -11,7 +11,7 @@
 <rss>
 	<channel>
 		<partial model="@((title, link))" name="_RssChannel" />
-		<atom-link href="@(Settings.BaseUrl)/submissions.rss" />
+		<atom-link href="@(Settings.BaseUrl)/Submissions.rss" />
```
adelikat, line 14: Is this needed for the rss spec?  If not we dont' want to do this.  Integrating systems can be case sensitive on urls.  It was lowercase to avoid potentially something

Masterjun3: All our links to the rss feeds are capitalized like this, and rss requires it to be exactly the same.

Masterjun3: Or rather, we have some that are uppercase ( https://tasvideos.org/XMLDataFeed or https://tasvideos.org/NewMovies because they are normalized to uppercase by our wiki engine) and some that are lowercase (the link tags in our main html).
But unfortunately both capitalizations return the same rss, and this atom link needs to be exact.

TASVideos/Pages/RssFeeds/Wiki.cshtml
```diff
@@ -20,8 +20,8 @@
 				<rss-link>@url</rss-link>
 				<description>@edit.RevisionMessage</description>
 				<comments>@(Settings.BaseUrl)/Forum/</comments>
-				<guid>@url</guid>
-				<pubDate>@edit.PubDate.ToString("u", CultureInfo.InvariantCulture)</pubDate>
+				<guid>@url?revision=@edit.RevisionId</guid>
```
adelikat, line 23: This is a decision. Just pointing it out.  I'll trust it makes more sense

TASVideos/Pages/RssFeeds/Submissions.cshtml
Masterjun3: Based on the wiki and the rest of our code, it seems the original intent was for the rss urls to be uncapitalized, so `submissions.rss`.
I would suggest reverting the canonical path changes.

TASVideos/Pages/RssFeeds/Wiki.cshtml
YoshiRulz: As noted in the commit message, these need to be unique within the file or it [fails validation](https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Ftasvideos.org%2FWiki.rss). That doesn't seem to be explicit in the spec, but [it does effectively say](https://www.rssboard.org/rss-specification#ltguidgtSubelementOfLtitemgt) that if the server doesn't ensure uniqueness then clients won't cache correctly.

The spec also notes [guid != link](https://www.rssboard.org/rss-specification#comments), if that was your concern with the revised value.

TASVideos/Pages/RssFeeds/Submissions.cshtml
YoshiRulz: I thought ASP.NET forced all its pages to be PascalCase?

Masterjun3: The actual page is `/RssFeeds/Submissions` but we reroute it to `/submissions.rss`.
(See `ServiceCollectionExtensions.Aliases` for all our route changes.)

TASVideos/Pages/RssFeeds/Wiki.cshtml
adelikat: Its actually more than an alias, it IS the name of the route, no redirect occurs when somone goes to /submissions.rss

adelikat (approved):

PR #2105, opened 2025-03-17
Set `Cache-Control` header for static images

YoshiRulz:
https://github.com/TASVideos/tasvideos/blob/679c5eba46958457a893aae9bcffa248bb3aa192/TASVideos/Extensions/ApplicationBuilderExtensions.cs#L41-L48

Inspired by [this Lighthouse diagnostic](https://developer.chrome.com/docs/lighthouse/performance/uses-long-cache-ttl/) failing.

### DIFF
--- a/TASVideos/Extensions/ApplicationBuilderExtensions.cs
+++ b/TASVideos/Extensions/ApplicationBuilderExtensions.cs
@@ -35,10 +35,24 @@ public static class ApplicationBuilderExtensions
 
 	public static IApplicationBuilder UseStaticFilesWithExtensionMapping(this IApplicationBuilder app, IWebHostEnvironment env)
 	{
+		static void MaybeSetCacheHeader(StaticFileResponseContext ctx)
+		{
+			var fileType = ctx.Context.Response.ContentType ?? string.Empty;
+			if (fileType.StartsWith("image/"))
+			{
+				// re: `public` directive: without it, the browser will assume that all resources downloaded while logged in are secret, and won't cache them.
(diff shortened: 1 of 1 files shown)


adelikat:
Is there not cache control already for static images? Would it be better to do this at the nginx level rather than asp.net? 

If i understand correctly this would either add a kestrel pipeline to all these requests, or do nothing because the server never hits kestrel in the first for static images

Masterjun3:
We don't currently have any cache control configuration on our site (easily tested by checking the response headers in any browser dev tool).

Although adding a one-week would mean that if we update publication screenshots, people might not see it until a week has passed?

Masterjun3:
> the browser will assume that all resources downloaded while logged in are secret, and won't cache them
But it looks like all browsers cache these images anyway. So this PR isn't necessary?

YoshiRulz:
> Although adding a one-week would mean that if we update publication screenshots, people might not see it until a week has passed?

Yep. Which is why I went with a week and not the usual month.

>> the browser will assume that all resources downloaded while logged in are secret, and won't cache them
> 
> But it looks like all browsers cache these images anyway. So this PR isn't necessary?

I don't like the idea of relying on ~~de facto~~ [loosely-defined](https://www.rfc-editor.org/rfc/rfc7234#section-4.2.2) UA behaviour, but it's your call. I noticed this was missing while running Google's page load measurement tool and figured it was free bandwidth savings.

Masterjun3:
According to spec, "public" is only necessary for requests with an "Authorization" header. We don't use that header for authorization, we use cookies.

YoshiRulz:
`ETag` is already being set, which should save bandwidth even without this.

PR #2106, opened 2025-03-19
2024 Awards Maintenance

Spikestuff:
Some final psd files that were missing for templates (and minor adjustments). And clearer x2/x4 fonts for those 2024 Awards (not applied elsewhere).

Secondary automation file provided for if x2/x4 is made in time without the need for relying on the x1 font size.
_Uh--_ let me know if this breaks if anyone tries it out on the previous years. Showcasing it working in 2024 for convenience and most recent sake.

Made a more accessible area for the incompatible (standard) awards located within Automation.
You can use "majoraward.psd" for TAS/TASer/Rookie located in template to save some time-- use the alpha trim afterwards~

One thing to note and probably discuss about:
The [template awards](https://tasvideos.org/Awards) that are used on site is the x1 so the possibility of deleting the x2/x4 equivalent might be on the table?
Unless I've 100% blinded where the larger template variations are being used elsewhere on the site.


adelikat (approved):

Masterjun3:
We currently use the x2 and x4 variants in almost all places where the x1 is used. It is automatically used for high-dpi displays to have a better resolution. On browsers you can manually see the higher scaled versions when you zoom in on the page.

Masterjun3:
Yes it does work, but they are always resized to *look* the same, it's only the internal resolution that is higher.
Check for example my profile https://tasvideos.org/Users/Profile/Masterjun , refresh it with Ctrl+F5, then press F12 and open the Network tab, and then zoom in and you will see network requests for the higher quality images.

Masterjun3:
The higher quality images are used on the Awards Editor pages.

PR #2109, opened 2025-03-22
Set `lang` attr

YoshiRulz:
Sets the page language globally per a Lighthouse diagnostic (which doesn't have a link for some reason even though it's marked as critical? [DIY](https://pagespeed.web.dev)). Also sets the language of translated wiki content, but not forum content. The presence of these attributes may help automated translation tools. Lighthouse says it helps screen readers too.
Any text within a page which is not English prose and should not be translated (MDN's example is wordmarks, so "TASVideos", but maybe usernames and game names) can be given the attribute `translate="no"`. I haven't done that.

While this PR doesn't touch the forum, I think it would be simple enough to
```csharp
Model.Topic.CategoryId switch {
	53 => "ja", // Japanese subforum
	55 => "fr", // French subforum
	58 => "es", // Spanish subforum
	59 => "pt", // Portuguese subforum
	62 => "ru", // Russian subforum
	66 => ""/*unknown*/, // Other Languages subforum (don't have individual language metadata)
	_ => "en"
}
```
...it would just be a bit tedious to go through and add the attr to all the relevant elements.


adelikat:
hardcoded IDs for all the root level pages, yeesh.  And definitely more complexity, I'm not seeing why the value is enough to warrant this.  Surely there is a better approach

YoshiRulz:
Yeah that's just one of the reasons I opted not to touch the forum.

adelikat:
I'm closing this PR.  I suppose there is value in setting the lang attribute but I don't this code merged.  I'm not seeing enough value add for this amount of mess and complexity.  Maybe there's a more elegant solution out there to make it worth the value add, but I'm not seeing it

Issue #640, opened 2021-12-06
[F.Rq.] Gravatar


YoshiRulz:
tl;dr: allows users to change their avatar on many sites at once, and auto-generates unique avatars when a user doesn't provide one. You've probably seen the latter e.g. on GitHub. [API docs](https://en.gravatar.com/site/implement)

update 2024-12: TIL there's an open alternative to Gravatar called [ivatar](https://wiki.libravatar.org/features/). It says it can fall back to Gravatar so maybe it's a direct upgrade?

see also #227

meshuggahtas:
I'm not sure what does the first half means (change their avatar on many sites at once).

Personally I hate auto generated avatars, as it looks like I have picked specifically an idiot avatar I never choose. Just look at my auto generated green cactus gravatar.

Personally I hope this comes with the ability to not use any profile at all.

Since we don't really have a designated designers, I'm not sure about how much effort is required to have a fine selection of usable avatars.

ActuallyTruncated:
I disagree with the closing of #227 in favor of this solution. In fact, I do not think this suggestion is a solution at all - it is more like deciding to keep having this problem, only choosing that it will occur for everyone simultaneously when Gravatar stops serving free bandwidth.

If you think to yourself "nah, that will never happen", consider that we once had ICQ and MSN Messenger integrations.

adelikat:
going with your example, while ICQ and MSN might not be a good idea for integration today, it doesn't mean that going back in time, it would have been a good idea to have a custom messaging system as an alternative.  By the time those things aren't relevant, so was the feature.

Gravatar might not exist, but the concept seems like it can reasonably exist as it solves a current problem for at least as long as the avatar problem stays relevant.  Assuming this is a simple integration (still looking at the docs), should be easy enough to do something else should the day come.

YoshiRulz:
Here's an idea for a UI to handle multiple sources, taken from StackOverflow (in case you add OAuth for login with Discord, or autogenerated "identicons", etc.):
![Screenshot_20220922_005241](https://user-images.githubusercontent.com/13409956/191538281-ee7c077b-85f9-4147-a7c6-fca977957cd3.png)

adelikat:
Can we please make a separate request for some of the improvements.  I have no issue with identicons, probably, but I would like to let this old issue die, the original intent has been delivered

PR #2113, opened 2025-03-27
Replace Gravatar with Libravatar (ivatar)

YoshiRulz:
It's the same but theoretically decentralised, except IDK how to make DNS queries from a Razor Page and approximately 0 people will have that set up, so I just left it and told the user to do it themself.

Mass-rename is a bit noisy but I figured it would be confusing to leave references to Gravatar everywhere.

Should this be rejected: Should take the `//gravatar.com/` change and the `d=` change, and also need to set `r=pg` (latter does not apply to ivatar hence it's not in this PR).

### DIFF
--- a/TASVideos.Common/Constants.cs
+++ b/TASVideos.Common/Constants.cs
@@ -81,7 +81,7 @@ public static class SiteGlobalConstants
 
 	public const int YearsOfBanDisplayedAsIndefinite = 2;
 
-	public const string MainGravatarDomain = "gravatar.com"; // keep in sync with profile-settings.js
+	public const string MainIvatarDomain = "gravatar.com"; // keep in sync with profile-settings.js
 }
 
 public static class ForumConstants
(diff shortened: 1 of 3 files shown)


adelikat:
Why do this

YoshiRulz:
It's decentralised as I said. Someone raised concerns about loss of service in https://github.com/TASVideos/tasvideos/issues/640#issuecomment-1235742509, but I think the greater risk with Gravatar is that it's centralised so they could track UAs across origins. [The docs](https://wiki.libravatar.org/features/) list some other minor differences.

adelikat:
If i deployed this, my avatar would stop working, until I went to settings and changed something?

YoshiRulz:
It worked for me locally, which makes sense because this whole thing it just sugar on top of the avatar URI field. The "Use Gravatar" checkbox would be populated to unchecked, though, since it's now looking for the new base domain.

PR #2114, opened 2025-03-27
Subtract 1 from FPS for Hourglass WTF parser

Randomno:
If I'm understanding correctly, Hourglass adds 1 to the FPS value before saving, and site doesn't account for this. There's only been a handful of submissions since the new site so it hasn't been very noticeable. Seems like this error originated in the [movie file format](https://code.google.com/archive/p/hourglass-win32/wikis/MovieFileFormat.wiki) page on Google Code.

https://github.com/TASEmulators/hourglass-win32/blob/78fa7c6d2be1b5eb46b8d4c0a37e56af616e1a46/src/wintaser/wintaser.cpp#L936
https://github.com/TASEmulators/hourglass-win32/blob/78fa7c6d2be1b5eb46b8d4c0a37e56af616e1a46/src/wintaser/wintaser.cpp#L1071

### DIFF
--- a/TASVideos.Parsers/Parsers/Wtf.cs
+++ b/TASVideos.Parsers/Parsers/Wtf.cs
@@ -22,10 +22,10 @@ internal class Wtf : Parser, IParser
 		br.ReadInt32(); // input frames
 		result.RerecordCount = br.ReadInt32();
 		br.ReadBytes(8); // keyboard type
-		var fps = br.ReadUInt32() - 1;
-		if (fps > 0)
+		var fps = br.ReadUInt32();
+		if (fps > 1)
 		{
-			result.FrameRateOverride = fps;
+			result.FrameRateOverride = fps - 1;
 		}
 
 		return await Task.FromResult(result);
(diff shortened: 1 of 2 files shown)


adelikat:
Are you sure it isn't intended?

Do unit tests need to update?

Randomno:
> Are you sure it isn't intended?
I assume Hourglass has a reason to add 1 to the FPS, but it's not the true FPS value.

Fixed the underflow and the existing tests.

It will ignore the FPS value in the movie file when it's 0 or 1 (Hourglass should never save FPS as 0 anyway).

Masterjun3 (approved):
This PR is correct.
I checked my hourglass files from 2012 and they also had a 61 framerate value in the movie file, when it was really 60 being played back (and recorded into AVI).

According to the hourglass source code, a frame rate value of 0 inside the movie file means it uses the default framerate of 60. I assume this was done because pre-release movie files had this block fully reserved with 0's, so it was for backwards compatibility.
This was before the code was public though, so we don't really need to support that edge case.
This PR is fine as is.

PR #2115, opened 2025-03-27
Link to `/Movies-<platform>` from game pages

YoshiRulz:
I would have linked to `/Platforms` but the anchors there are auto-generated from wiki content, so they don't match.

### DIFF
--- a/TASVideos/Pages/Games/Index.cshtml
+++ b/TASVideos/Pages/Games/Index.cshtml
@@ -22,6 +22,8 @@
 	{
 		ViewData.SetNavigation(id, "/{0}G");
 	}
+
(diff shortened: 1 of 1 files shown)


Masterjun3 (approved):

Masterjun3:
I accepted this, but System/Platform links like this feel very similar to our profile links, game links, submission links, etc., all of which are a TagHelper.
So they're used like
`<profile-link username="@user.UserName"></profile-link>` or
`<pub-link id="@pub.Id">@pub.Title</pub-link>`.
So maybe a System or Platform TagHelper would be the solution that is similar to how the other code is structured.

PR #2118, opened 2025-04-04
Add crawl delay to robots.txt (unofficial, but some bots adhere to it)

Masterjun3:
For bots following this directive, it limits the requests of bots to 60 seconds between requests.
Currently 94% of requests to our site are coming from bots. Maybe this will slow them down a bit.

The Crawl-delay directive is unofficial, but has an [entry on wikipedia](https://en.wikipedia.org/wiki/Robots.txt#Crawl-delay_directive).


adelikat (approved):

YoshiRulz:
How effective was this?

Masterjun3:
We went from an average of 12.79 total requests/sec down to 6.69 total requests/sec once this was deployed.

My data is this metric data from before the deploy, gathered at 2025-04-04 19:15 UTC, at which point the site was deployed for 191 hours.
20250404-191500-utc


And the comparison is this data from after the deploy, gathered at 2025-04-08 19:00 UTC, at which point the site was deployed for 52 hours.
20250408-190000-utc

PR #2125, opened 2025-04-11
Fix double enumerations

YoshiRulz:
I was looking for caching of an `IEnumerable` object, and I found one, which led me to this.

### DIFF
--- a/TASVideos.Core/Services/Awards.cs
+++ b/TASVideos.Core/Services/Awards.cs
@@ -258,7 +258,7 @@ internal class Awards(ApplicationDbContext db, ICacheService cache) : IAwards
 					.ToList()))
 			.ToList();
 
-		var allAwards = userAwards.Concat(publicationAwards);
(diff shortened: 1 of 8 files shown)


YoshiRulz:
To assuage adelikat's concerns about the genericisation commit:
- `MemoryCacheService` passes `data` to a generic extension method `IMemoryCache.Set`, so while the type parameter for that now changes to match the one from the caller, I checked and the extension just assigns it to `ICacheEntry.Value` which is an `object?` prop.
- `WikiTextCache` throws if `data is not WikiResult`.
- `RedisCacheService` and `TestCacheService` both pass `data` to `JsonSerializer.Serialize`. That's obviously working even when the type parameter is `object?`, so narrowing it isn't going to break anything. All the tests which round-trip via `TestCacheService` still pass. And both that and `RedisCacheService` already have a corresponding `JsonSerializer.Deserialize` call which is fully generic.

Strictly speaking, the latter commits shouldn't require it, but it is something to be fixed.

adelikat (approved):

PR #2130, opened 2025-04-12
Improve PublicationsByPlatform module: show counts, hide empty, allow flags, and allow custom ordering

### LINKED ISSUE #1910: Automate the per-platform movie table

Masterjun3:
Resolves #1910 properly.
The `[module:PublicationsByPlatform]` module wasn't useful in its current form.
But now it should allow what @vadosnaprimer wanted in the opening post of that issue ticket.

- Hide empty links
- Show Stars even though it's now a flag not a class

As a bonus, we're already loading the counts, so why not display them too!

Here an example image where I used
`[module:PublicationsByPlatform|flags=commentary|order=standard,stars,commentary,alternative]`
(Note: In my DB `Stars` is still a class, and `Events` does not exist yet, so the example uses the `Commentary`  flag instead.)

Before | After
image

The current `/Movies` page is ordered like `Standard|Stars|Alternative`, so when this PR is merged we can use the following line for a similar ordering:
`[module:PublicationsByPlatform|flags=stars|order=standard,stars,alternative,events]`

### DIFF
--- a/TASVideos/WikiModules/PublicationsByPlatform.cshtml
+++ b/TASVideos/WikiModules/PublicationsByPlatform.cshtml
@@ -4,7 +4,7 @@
 	{
 		<tr>
 			<td>@platformEntry.Platform</td>
-			<td><a href="/Movies-@platformEntry.PlatformCode">All Publications (@platformEntry.Groupings.Sum(c => c.PublicationCount))</a></td>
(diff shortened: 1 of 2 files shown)


vadosnaprimer:
Thank you!!!

vadosnaprimer (approved):

PR #2136, opened 2025-04-27
Add embeds to Bluesky posts on submissions and publications

Masterjun3:
As mentioned in PR #2057 that added Bluesky support, adding embed support is a lot more difficult and can be done at a later time.

This is the later time.

The challenge is that unlike Discord (and Twitter back when Twitter was a thing), Bluesky does not automatically show our open graph meta tags as an embed.
Instead, we have to manually upload an image to Bluesky before making a post, and then use the returned reference blob in the post creation.
This creates the issue that we have to actually have the image data in a byte array to send to Bluesky. This is fine for publications. But for submissions, we have to take the user-submitted encode embed link and make a http request to youtube.

Additionally, Bluesky recommends sending the aspect ratio of the image for proper display. For submissions the youtube embed is always 480x360. For publications we don't send any aspect ratio information, because that would require parsing the image bytes, and that is complicated and maybe even dangerous. Sadly, this means for publications Bluesky will display the image in the default 1x1 aspect ratio box (see image below). It isn't pretty, but it's fine.

**IMPORTANT:** We make a request to a URL based on user data. This is dangerous and needs to be reviewed properly. The link is always prepended with `https://i.ytimg.com/vi/`, so I'm unaware of anything that can be abused here.

<hr>

Examples:

image

image

### DIFF
--- a/TASVideos.Core/Services/ExternalMediaPublisher/Distributors/BlueskyDistributor.cs
+++ b/TASVideos.Core/Services/ExternalMediaPublisher/Distributors/BlueskyDistributor.cs
@@ -1,4 +1,5 @@
 ﻿using System.Globalization;
+using System.Net.Http.Headers;
 using System.Text;
 using System.Text.Json.Serialization;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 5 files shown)


adelikat (approved):

Masterjun3:
I don't know what you want. Our posts lacked a visual indicator.
I followed the [Images embeds](https://docs.bsky.app/docs/tutorials/creating-a-post#images-embeds) part of the Bluesky API tutorial.
You instead seem to want the "Website card embeds" part of the tutorial.

Please explain in what part **my PR** is incomplete in what **I set out to do**.

PR #2139, opened 2025-05-08
.3ct parser

adelikat:
This also adds in a time display assumption that if the time is zero, then we round up to 0.01 with the assumption that at least some amount of time as occurred

Other decisions:
Sets the framecount to the cyclecount and the framerate to the clockrate in order to generate a precise time.  Ideally we would have an even more precise time but that is a more involved change that affects all pub/subs, and I do not want to do that work as part of this PR (or ever, ideally)

Sets the rerecord count to 100.  Its an arbitrary number, and I need something other than zero to better show its intentional rather than a parse error

### DIFF
--- a/TASVideos.Common/ITimeable.cs
+++ b/TASVideos.Common/ITimeable.cs
@@ -42,7 +42,7 @@ public static class TimeableExtensions
 			return timeSpan.ToString(@"h\:mm\:ss\.ff");
 		}
 
-		if (timeSpan.TotalSeconds == 0.0)
+		if (timeSpan.TotalSeconds <= 0.01)
 		{
 			return "00:00.01";
 		}
--- a/TASVideos.Common/ITimeable.cs
+++ b/TASVideos.Common/ITimeable.cs
@@ -42,6 +42,11 @@ public static class TimeableExtensions
 			return timeSpan.ToString(@"h\:mm\:ss\.ff");
 		}
 
+		if (timeSpan.TotalSeconds == 0.0)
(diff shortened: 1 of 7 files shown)


vadosnaprimer:
0 rerecords appear on submission page as "unknown" which is true for a setup like this, so maybe it's better?

adelikat:
for me, unknown tells me that something might be wrong.  But with this format, there are no rerecords by design, I thought a different number would make that clearer

adelikat:
fine.

TASVideos.Common/ITimeable.cs
```diff
@@ -42,6 +42,11 @@ public static string ToStringWithOptionalDaysAndHours(this TimeSpan timeSpan)
 			return timeSpan.ToString(@"h\:mm\:ss\.ff");
 		}
 
+		if (timeSpan.TotalSeconds == 0.0)
+		{
+			return "00:00.01";
+		}
+
 		return timeSpan.ToString(@"mm\:ss\.ff");
```
Masterjun3, line 50: Instead of `if (timeSpan.TotalSeconds == 0.0)` this should be `if (timeSpan.TotalSeconds <= 0.01)`, beacuse otherwise a movie time of 0.002s would still be shown as 00:00.00.

Masterjun3 (changes requested):

vadosnaprimer (approved):

Masterjun3 (approved):

PR #2142, opened 2025-05-10
Fix small bugs

Masterjun3:
When unpublishing a publication, we should delete its wiki page, like we do when a submission is deleted.

Also the page size was 0 on our `/Logs` page, because in the PagingDefaults attribute only the sort was given, and the int's default value is 0.


adelikat (approved):

YoshiRulz:
This was re: https://tasvideos.org/Forum/Posts/535990. Presumably that still needs cleaning up manually?

PR #2147, opened 2025-06-15
Mark `.conf` (DOSBox) and `.xml` (EmuHawk) as previewable userfiles

YoshiRulz:


vadosnaprimer:
In general .conf is a generic extension that may be just binary, what would appear in the preview for such a case?

YoshiRulz:
Mojibake string:
https://github.com/TASVideos/tasvideos/blob/92684e8b2dbe0b33c7e69ddd0e0029046c0815c8/TASVideos/Pages/UserFiles/Info.cshtml.cs#L42
Though there's a second check on whether the preview is shown, not sure how that works:
https://github.com/TASVideos/tasvideos/blob/92684e8b2dbe0b33c7e69ddd0e0029046c0815c8/TASVideos/Pages/UserFiles/Info.cshtml.cs#L29

vadosnaprimer (approved):
I personally prefer displaying actually important stuff like xml and conf (which is used for the new dosbox core in hawk), and for weird formats that are internally binary it's fine to sacrifice their nice look (or lack thereof). Bizhawk being easy to use is one of our missions, and it's much easier if contents of those 2 are instantly visible.

PR #2148, opened 2025-06-16
Reorganise navbar, adding, moving, and renaming links

YoshiRulz:
Preview of how it would look (modulo Wiki and Admin menus):
screenshot

### DIFF
--- a/TASVideos/Pages/Shared/_Layout.cshtml
+++ b/TASVideos/Pages/Shared/_Layout.cshtml
@@ -43,7 +43,7 @@
 				</picture>
 			</div>
 			<a id="icon" asp-page="/Index">
-				<picture>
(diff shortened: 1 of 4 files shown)


vadosnaprimer:
I discussed this with staff. Nobody supports these changes as they are, and there's so many of them, and (as it often happens) zero explanation of which exact usage problems are being solved here. Every such change would need a separate discussion, and again the problem being solved has to be stated clearly for every single case. Readjusting everything to one's personal taste is not very productive, unless it's a part of a brainstorming session aiming to solve existing problems that are already agreed upon.

I don't be listing everyone's take on this PR, if people want they can repost them here. Closing until there's definitive discussion of problems to solve in each individual case.

PR #2155, opened 2025-06-27
Navbar bugfixes

YoshiRulz:
The first commits from #2148 which don't affect the links or their order.


YoshiRulz:
before
after
comparison
(That's not the hover effect; the user's selection is now shown.)

The search box bug is that the top bar was unnecessarily cramped for the mid-sized media query:
before
after
comparison

PR #2156, opened 2025-06-28
Merge "Home" and logo in navbar

YoshiRulz:
This link is redundant with the logo and wordmark to its immediate left. Removing it frees a bit of space for the other items—in my testing here it eliminated the third row at specific narrow widths, but that was done on the live site i.e. without https://github.com/TASVideos/tasvideos/commit/67a91a7f2e3b55c6f014d8c24bb4fa5da350f6b9.
Obviously the red boxes aren't part of the site, I'm just showing the clickable regions for comparison.
before
after
comparison

Cherry-picked from #2148.

### DIFF
--- a/TASVideos/Pages/Shared/_Layout.cshtml
+++ b/TASVideos/Pages/Shared/_Layout.cshtml
@@ -43,7 +43,7 @@
 				</picture>
 			</div>
 			<a id="icon" asp-page="/Index">
-				<picture>
+				<picture title="To Homepage">
 					<source srcset="/images/logo-light.webp .5x,
 							/images/logo-light-2x.webp 1x,
 							/images/logo-light-4x.webp 2x"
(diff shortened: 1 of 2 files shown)


bigbass1997:
I'm not sure what problem there is with having three rows, since the site does handle that arrangement okay.

But in any case, the Home link does serve a purpose: it becomes highlighted when you are on the homepage, which helps indicate to unfamiliar visitors where they are. Plus, while subjective, I think it's reasonable to believe that not everyone is aware you can click the logo to go to the homepage.

Masterjun3:
I agree with @bigbass1997 and I think removing the Home button isn't right.
Youtube, Twitter, Wikipedia, and a bunch of social media sites all have a "Home" button despite being right next to the site logo that brings you to the same page.
Our logo and site title don't particularly *look* clickable anyway.

PR #2166, opened 2025-07-08
Strip queryparams from submission YT embed link

YoshiRulz:
this is not retroactive; subs can be fixed with just a null edit


adelikat:
The build failed, is there unit tests that need to be updated?

Issue #1845, opened 2024-05-11
[F.Rq.] "Source map" for HTML to make contributing easier


YoshiRulz:
My experience with contributing to the site, being a novice to ASP.NET, has involved a lot of grepping the codebase for tag/attribute names in order to find the source file where the thing I'm looking at originates. I've since learned that the directory structure under `/TASVideos/Pages` roughly corresponds to page paths, but that doesn't account for 'partials'.
I propose that every source file which "renders" output should also include its own path as a simple HTML comment, such that more people might be able to submit fixes or at least precisely identify where problems are.
I've found that `Html.ViewContext.ExecutingFilePath` produces the right path (relative to `$(ProjectDir)`), but it would be a pain to manually include that in every page and partial. Does anyone know of a better way?
Have a couple related SO links: [1](https://stackoverflow.com/questions/4248125/how-can-you-find-which-view-called-which-partial-view-in-asp-net-mvc), [2](https://stackoverflow.com/questions/6311463/how-to-get-the-current-view-name-in-asp-net-mvc-3)

nattthebear:
I agree that the file structure is difficult to follow.  ASP.NET is heavily convention based and I haven't kept up on the conventions.

> I propose that every source file which 'renders' output should also include its own path as a simple HTML comment

Does that mean the final emitted markup served to the end user would have these comments in it?  Would that inflate our document size too much?

YoshiRulz:
Just as a ballpark figure I'd say the worst-affected pages would grow by at most 1 KiB *before compression*. This could be locked to only development environments if that's a problem, given that contributors are expected to test locally anyway.

Masterjun3:
One idea is to add one line to `_ViewStart.cshtml` that does something similar.
`<!-- @(Html.ViewContext.View.Path) -->`
Result:
![image](https://github.com/TASVideos/tasvideos/assets/22375320/1dc4786b-40c5-4e08-b69b-5ba7585f46ea)

But this [only works for views, not for partials](https://learn.microsoft.com/en-us/aspnet/core/mvc/views/layout?view=aspnetcore-8.0#running-code-before-each-view).

Masterjun3:
Okay actually, we can override (or I think it's called to shadow) the `Activate` method of the `RazorPageActivator` and use that in the services.

![image](https://github.com/TASVideos/tasvideos/assets/22375320/d8377e88-81ef-4f87-9281-cf5248b9767d)

[...]

PR #2174, opened 2025-07-28
Include source file paths as HTML comments in development environments

YoshiRulz:
Per https://github.com/TASVideos/tasvideos/issues/1845#issuecomment-2115759800, the general solution can only print at the start of files. Rather than manually add comments to the ends of files, I think selectively adding "pause"/"resume" around nested partials serves the same purpose with almost no work. This PR only adds that to the outermost partial, to be refined later.

The below examples are from the homepage and a publication page.
Screenshot from 2025-07-29 05-19-49


YoshiRulz:
If I find some output where it's not obvious which partials have closed, yes.

adelikat:
@YoshiRulz : can you fix the merge conflict?

adelikat (approved):

PR #2175, opened 2025-07-28
Improve UX of topic watch button

YoshiRulz:
I added text to the button, like the staff-only buttons beside it have. edit: And changed the wording from "watching" to "subscribing".


vadosnaprimer:
I like Masterjun's suggestion

изображение

vadosnaprimer (approved):

PR #2176, opened 2025-07-28
When editing a wiki page, include the history button in the footer

YoshiRulz:


TASVideos.Core/Extensions/WikiHelper.cs
```diff
@@ -149,7 +150,7 @@ public static bool IsGameResourcesPage(this IWikiPage page)
 		return IsGameResourcesPage(page.PageName);
 	}
 
-	public static bool IsHomePage(string? pageName)
+	public static bool IsHomePage([NotNullWhen(true)] string? pageName)
```
adelikat, line 153: whhy the NotNullWhen attributes? what does that gain us?

adelikat (approved):

PR #2178, opened 2025-07-29
reduce memory allocations in wikiengine

adelikat:
Before

| Method       | Mean     | Error     | StdDev    | Gen0    | Gen1    | Allocated |
|------------- |---------:|----------:|----------:|--------:|--------:|----------:|
| ParseMarkup  | 1.144 ms | 0.0227 ms | 0.0213 ms | 33.2031 |  7.8125 | 277.98 KB |
| RenderToHtml | 1.315 ms | 0.0257 ms | 0.0240 ms | 44.9219 | 13.6719 |  368.6 KB |
| RenderToText | 1.184 ms | 0.0200 ms | 0.0187 ms | 42.9688 | 13.6719 | 352.01 KB |

After
| Method       | Mean     | Error     | StdDev    | Gen0    | Gen1    | Allocated |
|------------- |---------:|----------:|----------:|--------:|--------:|----------:|
| ParseMarkup  | 1.093 ms | 0.0158 ms | 0.0132 ms | 27.3438 |  5.8594 | 226.67 KB |
| RenderToHtml | 1.280 ms | 0.0225 ms | 0.0211 ms | 37.1094 | 11.7188 | 317.29 KB |
| RenderToText | 1.217 ms | 0.0222 ms | 0.0208 ms | 35.1563 | 11.7188 | 300.71 KB |

### DIFF
--- a/TASVideos.WikiEngine/NewParser.cs
+++ b/TASVideos.WikiEngine/NewParser.cs
@@ -851,7 +851,7 @@ public partial class NewParser
 			var e = new Element(_index, "code");
 			if (lang != "")
 			{
-				e.Attributes["class"] = string.Concat("language-", PrismNames.FixLanguage(lang));
+				e.Attributes["class"] = "language-" + PrismNames.FixLanguage(lang);
 			}
 
 			e.Children.Add(new Text(_index, EatSrcEmbedText()) { CharEnd = _index });
--- a/TASVideos.WikiEngine/NewParser.cs
+++ b/TASVideos.WikiEngine/NewParser.cs
@@ -115,7 +115,7 @@ public partial class NewParser
 	private string EatToBracket()
 	{
 		var from = _index;
-		var ret = new StringBuilder();
+		var start = _index;
 		for (var i = 1; i > 0;)
 		{
 			if (Eof())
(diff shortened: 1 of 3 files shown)


TASVideos.WikiEngine/NewParser.cs
```diff
@@ -856,7 +851,7 @@ private void ParseStartLine()
 			var e = new Element(_index, "code");
 			if (lang != "")
 			{
-				e.Attributes["class"] = "language-" + PrismNames.FixLanguage(lang);
+				e.Attributes["class"] = string.Concat("language-", PrismNames.FixLanguage(lang));
```
nattthebear, line 854: Does this specific change gain us anything?  Seems like it's replacing `string + string` with `string.Concat(string, string)` which I would expect to be essentially identical.

nattthebear (approved):
These look like really nice changes, I especially like the rework of the `Eat` family of methods.

I don't have time right now to do any strenuous regression testing, but I'm not too worried as any brokenness would likely be pretty obvious and quickly fixable.

TASVideos.WikiEngine/NewParser.cs
CasualPokePlayer: They are identical, they compile to the same code in the end (+ operator calls string.concat if both strings are not constant).

PR #2186, opened 2025-08-05
Move `[InternalsVisibleTo]` to project files

YoshiRulz:
Apparently this has existed since .NET 5, I just missed it when I last was looking into assembly metadata.


TASVideos/TASVideos.csproj
```diff
@@ -56,5 +56,6 @@
 
 	<ItemGroup>
 		<ProjectReference Include="..\TASVideos.Api\TASVideos.Api.csproj" />
+		<InternalsVisibleTo Include="TASVideos.RazorPages.Tests" />
```
YoshiRulz, line 59: This one isn't working ~~for some reason~~ because of `<GenerateAssemblyInfo>false</GenerateAssemblyInfo>` above. I'll leave it as `AssemblyInfo.cs`.

adelikat:
That one exception to the rule really bugs me here.  If I pull this, I now have 2 places to check and maintain, so I feel like this adds value

adelikat:
We still use GitVersionTask, do we not?

adelikat:
Ok, remove that and convert everything the same, and I'll approve

adelikat:
@YoshiRulz : are you planning on finishing this?

adelikat (approved):

PR #2191, opened 2025-08-14
Support top-level dir in `.bk2`/`.tasproj`

YoshiRulz:
Per TASEmulators/BizHawk#4416. Leaving this as a draft until release.


YoshiRulz:
We didn't end up doing this, but you can take the changes anyway if you'd like to accept manually-created archives as EmuHawk does.

vadosnaprimer:
What does "manually-created archives as EmuHawk does" mean exactly?

YoshiRulz:
If you extract, edit, archive (rather than overwriting a single file with 7-Zip or similar), and the archive you create includes a top-level dir, EmuHawk will detect that and parse the movie.

PR #2193, opened 2025-08-16
Add in SubBSNESv115+ clock rates

CasualPokePlayer:


adelikat (approved):

Morilli:
This seems accurate but doesn't match any of the reported framerates from bizhawk itself:

https://github.com/TASEmulators/BizHawk/blob/654544b3290b3fcfe156327a359d16b449117cb9/src/BizHawk.Client.Common/movie/PlatformFrameRates.cs#L27

https://github.com/TASEmulators/BizHawk/blob/6284b56e81ebcb8470c87b0dca38c24749bc4406/src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/BsnesCore.cs#L130

Should we update those in bizhawk as well to match with this? Might be a bit annoying as `VsyncNumerator` is an int but at least the one in `PlatformFrameRates` could be changed easily.

Masterjun3:
These values do come from BizHawk though. When I make a SubSnes movie (Ntsc or Pal) in BizHawk, the .bk2 headers give me the values that were added in this PR. (That's also how I made the test files.)

Masterjun3:
I found it, the values come from here:
https://github.com/TASEmulators/BizHawk/blob/9c0e06f3fc7fd737a60fc3d1e3bea8b9567d208d/src/BizHawk.Emulation.Cores/Consoles/Nintendo/BSNES/SubBsnesCore.cs#L20-L23

Morilli:
Right, it seems `ClockRate` is different and used solely for movie header purposes. Although video dumping appears to use `VsyncNumerator` and `VsyncDenominator` and rtc logic also relies on these values. But perhaps I should take this to the bizhawk repo instead.

CasualPokePlayer:
VsyncDenominator seems to match up exactly with how BSNES internally does things (although it's technically more an average, the - 2 you see in PlatformFramerates is actually done as - 4 every other frame in BSNES). 21477272 is just rounded down 21477272.72727 (for making it an integer for VsyncNumerator anyways, it can be multiplied by 11, resulting in 236250000/3931026, which reduces down to 39375000/655171)

PR #2197, opened 2025-08-18
Add unit tests for link tag helpers

YoshiRulz:
Testing these:
https://github.com/TASVideos/tasvideos/blob/ffe8999d7e52e2016b52d92684ec4447916707c4/TASVideos/TagHelpers/WikiLinkTagHelpers.cs#L8-L91

### DIFF
--- a/tests/TASVideos.RazorPages.Tests/TestableHtmlGenerator.cs
+++ b/tests/TASVideos.RazorPages.Tests/TestableHtmlGenerator.cs
@@ -22,24 +22,6 @@ using Microsoft.Extensions.WebEncoders.Testing;
 
 namespace TASVideos.RazorPages.Tests;
 
-file static class ConfigureSubstitute
-{
-	public static T For<T>(Action<T> configure)
-		where T : class
-	{
-		var mock = Substitute.For<T>();
-		configure(mock);
-		return mock;
-	}
-
-	public static IOptions<T> ForIOptions<T>(T? optionsObj = null)
-		where T : class, new()
-	{
-		optionsObj ??= new();
-		return For<IOptions<T>>(mock => _ = mock.Value.Returns(optionsObj));
-	}
-}
-
 // ReSharper disable EmptyConstructor
 internal class TestableHtmlGenerator(
 	IOptions<MvcViewOptions> options,
@@ -47,12 +29,18 @@ internal class TestableHtmlGenerator(
 		Substitute.For<IAntiforgery>(),
 		options,
 		metadataProvider,
-		ConfigureSubstitute.For<IUrlHelperFactory>(
-			mock => _ = mock.GetUrlHelper(Arg.Any<ActionContext>())
-				.Returns(callInfo => new FakeUrlHelper(callInfo.ArgAt<ActionContext>(0)))),
+		CreateUrlHelperFactory(),
 		new HtmlTestEncoder(),
 		new DefaultValidationHtmlAttributeProvider(options, metadataProvider, new()))
 {
+	private static IUrlHelperFactory CreateUrlHelperFactory()
+	{
+		var urlHelperFactory = Substitute.For<IUrlHelperFactory>();
+		urlHelperFactory.GetUrlHelper(Arg.Any<ActionContext>())
+			.Returns(callInfo => new FakeUrlHelper(callInfo.ArgAt<ActionContext>(0)));
+		return urlHelperFactory;
+	}
+
 	private sealed class FakeUrlHelper(ActionContext context) : UrlHelperBase(context)
 	{
 		public override string Action(UrlActionContext actionContext)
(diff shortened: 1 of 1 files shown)


Masterjun3:
So basically, if you want to use `Page`, you actually have to give it a page that exists. For example `/123M` is not a page that exists. It's actually `/Publications/View` and we only reroute it like
`options.Conventions.AddPageRoute("/Publications/View", "{id:int}M")`.

Also the output TagName needs to be set to `a`, otherwise it stays at the original one like `pub-link`.
For example:
```diff
--- a/TASVideos/TagHelpers/WikiLinkTagHelpers.cs
+++ b/TASVideos/TagHelpers/WikiLinkTagHelpers.cs
@@ -11,7 +11,9 @@ public class PubLinkTagHelper(IHtmlGenerator generator) : AnchorTagHelper(genera

        public override void Process(TagHelperContext context, TagHelperOutput output)
        {
-               Page = $"/{Id}M";
+               output.TagName = "a";
+               Page = "/Publications/View";
+               RouteValues.Add("Id", Id.ToString());
                base.Process(context, output);
        }
 }
 ```

Though I'm wondering why we don't just set the href directly with `output.Attributes.Add("href", $"/{Id}M");` like it was done before this PR?

**Comment by YoshiRulz** [comment]
Thank you, that was the small detail I was missing. And since it doesn't work for wiki pages, I've reverted the change to `<wiki-link/>`, but the rest are now working.

File: TASVideos/TagHelpers/Buttons/AddLinkTagHelper.cs [hunk 1]
```diff
@@ -6,10 +6,13 @@ namespace TASVideos.TagHelpers;
 
 public class AddLinkTagHelper(IHtmlGenerator generator) : AnchorTagHelper(generator)
 {
+	public override void Process(TagHelperContext context, TagHelperOutput output)
+		=> ProcessAsync(context, output).Wait();
```
adelikat, line 10: This is bad. Why are we doing this? .Wait() on async calls is asking for deadlocks

TASVideos/TagHelpers/Buttons/AddLinkTagHelper.cs
```diff
@@ -6,10 +6,13 @@ namespace TASVideos.TagHelpers;
...
+
 	public override async Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
 	{
 		output.TagName = "a";
-		await base.ProcessAsync(context, output);
+		base.Process(context, output);
```
adelikat, line 15: Similarly, why are we not doing async here?

TASVideos/TagHelpers/Buttons/AddLinkTagHelper.cs
YoshiRulz: The first commit is to fix an inconsistency in subclasses of `AnchorTagHelper`. [In `TagHelper`, `ProcessAsync` calls `Process`](https://github.com/dotnet/aspnetcore/blob/a813ad450371e4c72578ac2d2d39a57e49dad23a/src/Razor/Razor/src/TagHelpers/TagHelper.cs#L39-L55), which is fine. `AnchorTagHelper` overrides `Process` *only*, which is fine. But those subclasses were overriding `ProcessAsync` only, so if `Process` is called (which I guess was never happening, thankfully) then it would be the implementation from `AnchorTagHelper`, which is obviously not going to produce the same output.

adelikat: Can we not implement both Process and ProcessAsync? The .Wait() is a no go for me, need another way

YoshiRulz:
Moved that to a separate PR, and rewrote the tests here to be async.

TASVideos/TagHelpers/WikiLinkTagHelpers.cs
```diff
@@ -5,36 +5,42 @@
...
-public class PubLinkTagHelper : TagHelper
+public class PubLinkTagHelper(IHtmlGenerator generator) : AnchorTagHelper(generator)
 {
 	public int Id { get; set; }
 
 	public override void Process(TagHelperContext context, TagHelperOutput output)
 	{
 		output.TagName = "a";
-		output.Attributes.Add("href", $"/{Id}M");
+		Page = "/Publications/View";
```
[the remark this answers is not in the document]
YoshiRulz: My earlier comment:
> The idea was to use the same mechanism as `<a asp-page/>`. After this PR I wanted to add more link helpers, but existing calls mostly use that rather than `<a href/>`. Also, I'm hoping that `<a asp-page/>` would help with tracking page links (for generating referrer lists).

The unit tests should catch regressions, but I also did a bit of manual testing.

TASVideos/TagHelpers/WikiLinkTagHelpers.cs
```diff
@@ -5,36 +5,42 @@
 
 namespace TASVideos.TagHelpers;
 
-public class PubLinkTagHelper : TagHelper
+public class PubLinkTagHelper(IHtmlGenerator generator) : AnchorTagHelper(generator)
```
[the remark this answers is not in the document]
YoshiRulz: Yes?
https://github.com/TASVideos/tasvideos/blob/c68804dd297c352057de6f84ae5f4fdbd259faf2/TASVideos/Pages/Shared/_DisplayMiniMovie.cshtml#L10
Do you mean the constructor parameter? It's being passed through to the base class.

Issue #2151, opened 2025-06-22
[Feature Request]  Give higher priority to time in TVC video titles


InputEvelution:
Publications on TASVideosChannel that can't fit into YouTube's title field cut off with ..., as seen here:

![Image](https://github.com/user-attachments/assets/ab9568dc-14e1-43cf-8cb1-3d679f253a6c)

As the time of the TAS goes at the end of the title, this is the first element to be excluded from view, despite typically being the aspect the entire movie is optimised around. Additionally, some TAS communities (e.g. SM64, Celeste) typically credit every single person involved in the optimization of a TAS regardless of how small or long ago their contribution was, which exacerbates this problem further.

The suggestion here is to give higher priority to the time of the TAS in some way or another. One way would be to shorten the author list (and potentially even the game name) with [...] instead, leaving the time intact, which I believe was done previously on the old site. Another, simpler solution could be to reorder the title so that the time comes after game and category but before the list of authors, which would improve readability of movie titles with massive numbers of authors in general.

InputEvelution:
That could also be a solution, though I'm not sure how others might feel about removing author attribution from the title altogether in these situations.

Masterjun3:
I've read this feature request, and it's not clear to me what the goal of this ticket is.
- Is the goal to truncate specifically the author list, like said in the title of the ticket?
- Or is the goal to always show the time in the title.

See for example this video https://www.youtube.com/watch?v=aGl8_qPXv6k where the title is currently:
`[TAS] GBA Simple 2960 Tomodachi Series Vol. 3: The Itsudemo Puzzle - Massugu Soroeru Straws by Da...`

Implemented as described in this issue ticket would mean the author list would be truncated here.

So what's the focus here? The author list being too long? The time always showing? Something else entirely?
If the focus is the time, this ticket's opening post and the title need to be adjusted.

PR #2199, opened 2025-08-24
Put time before authors in publication titles

InputEvelution:
Addresses #2151.

### DIFF
--- a/TASVideos.Data/Entity/Publication.cs
+++ b/TASVideos.Data/Entity/Publication.cs
@@ -123,8 +123,8 @@ public class Publication : BaseEntity, ITimeable
 		Title =
 			$"{System.Code} {gameName}"
 			+ (!string.IsNullOrWhiteSpace(goal) ? $" \"{goal}\"" : "")
-			+ $" by {string.Join(", ", authorList).LastCommaToAmpersand()}"
-			+ $" in {this.Time().ToStringWithOptionalDaysAndHours()}";
+			+ $" in {this.Time().ToStringWithOptionalDaysAndHours()}"
+			+ $" by {string.Join(", ", authorList).LastCommaToAmpersand()}";
 	}
 }
 


Masterjun3 (dismissed):
This is a change with very huge implications. It changes how titles for our movies are generated. We had the same formatting [for at least 19 years](https://web.archive.org/web/20060615005324/http://tasvideos.org/) now. This changes it.

Also, if we change publication titles, at least we also need to change the submission titles too. That's why I have this on "Request changes".

InputEvelution:
I've made adjustments so that the order is only different in the titles for YouTube videos. Though, I'm not sure how to test this.

Masterjun3:
You added a new DB column. That would require adding a new DB migration, and we would somehow let this new title be generated for all existing publications (which I kind of don't know how to do).

InputEvelution:
What course of action would you recommend?

TASVideos/Pages/Publications/Urls/Edit.cshtml.cs
```diff
@@ -49,6 +52,17 @@ public async Task<IActionResult> OnGet()
 		}
 
 		Title = title;
+		var youtubetitle = await db.Publications
+			.Where(p => p.Id == PublicationId)
+			.Select(p => p.YouTubeTitle)
+			.SingleOrDefaultAsync();
```
vadosnaprimer, line 58: I feel it should be possible to construct yt title on the fly without adding new DB column. By grabbing the correct pub here and just picking its fields that we need to construct the title. This is the only place where it's needed I think, so it doesn't have to exist in the Publication class for everyone (unlike normal title).

InputEvelution:
Ugh, I'll figure out what's going wrong with the build after syncing fork in the morning.

adelikat:
@InputEvelution : don't worry about it, the key part is the youtube title generation logic.  I have some refactors that are going to make this conflict worse.  I'll handle integrating this logic in, and handle conflicts, thanks for this PR

PR #2203, opened 2025-08-30
Hack in support for `Accept-Encoding: zstd`

YoshiRulz:
Per [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Encoding#browser_compatibility) this is supported in Chrome >= 123 (and thus in most browsers), as well as FF >= 126, but not Safari. Browsers which don't support Zstandard simply won't ask for it in `Accept-Encoding`, and will get either gzip or no compression.
There's no chance of this making it into .NET 10, but it's only 20 LOC to do it ourselves. The ZstdSharp library was already included transitively.

### DIFF
--- a/TASVideos/appsettings.Production.json
+++ b/TASVideos/appsettings.Production.json
@@ -1,6 +1,7 @@
 {
   "BaseUrl": "https://tasvideos.org",
   "EnableGzipCompression": "False",
+  "EnableZstdCompression": "False",
   "MinimumHoursBeforeJudgment": 72,
   "SubmissionRate": {
     "Submissions": 3,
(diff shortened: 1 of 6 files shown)


TASVideos/appsettings.Production.json
```diff
@@ -1,6 +1,7 @@
 {
   "BaseUrl": "https://tasvideos.org",
   "EnableGzipCompression": "False",
+  "EnableZstdCompression": "False",
```
YoshiRulz, line 4: Is this meant to be disabled in prod?

Masterjun3:
For production we do compression using nginx. That's why compression is disabled in the application code.

PR #2207, opened 2025-09-05
Override synchronous `Process` in `AnchorTagHelper` subclasses

YoshiRulz:
See https://github.com/TASVideos/tasvideos/pull/2197#discussion_r2323773674 for explanation.
I've attempted to implement these 1) without simply calling `ProcessAsync` from `Process`, which adelikat didn't like, and 2) while minimising code duplication.


YoshiRulz:
It's a "just in case" which I already ran into in practice while writing unit tests.

Looks like the [unit tests for `ScriptTagHelper`](https://github.com/dotnet/aspnetcore/blob/7c53dac57386a1a286b4d382f4aa36bbcf5829a1/src/Mvc/Mvc.TagHelpers/test/ScriptTagHelperTest.cs) are all async. I suspect that they would see odd failures if they made those synchronous for some reason.

PR #2211, opened 2025-09-14
adapts pull request #2199 logic 

adelikat:
adapts https://github.com/TASVideos/tasvideos/pull/2199
and puts it into the YoutubeSync class so it is consistent across all youtube sync operations

### DIFF
--- a/TASVideos.Core/Services/Youtube/YouTubeSync.cs
+++ b/TASVideos.Core/Services/Youtube/YouTubeSync.cs
@@ -67,7 +67,12 @@ internal class YouTubeSync(
 
 		var obsoleteStr = video.ObsoletedBy.HasValue ? "[Obsoleted] " : "";
 		var displayStr = !string.IsNullOrWhiteSpace(video.UrlDisplayName) ? $"[{video.UrlDisplayName}] " : "";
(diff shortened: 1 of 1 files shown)


TASVideos.Core/Services/Youtube/YouTubeSync.cs
```diff
@@ -67,7 +67,12 @@ public async Task SyncYouTubeVideo(YoutubeVideo video)
 
 		var obsoleteStr = video.ObsoletedBy.HasValue ? "[Obsoleted] " : "";
 		var displayStr = !string.IsNullOrWhiteSpace(video.UrlDisplayName) ? $"[{video.UrlDisplayName}] " : "";
-		string title = $"[TAS] {obsoleteStr}{displayStr}{video.Title}";
+
+		string[] titleSeparators = [" in ", " by "];
```
vadosnaprimer, line 71: What is these appear in game name?

InputEvelution:
I'm really sorry for not testing or thinking through this code more thoroughly.

At this point the only idea I have left would be to re-generate the entire title from scratch within YouTubeSync, but I don't know if that would be desirable or even feasible.

PR #2212, opened 2025-09-15
Add CelesteTAS movie file parser

InputEvelution:
This adds a parser for movies created with https://github.com/EverestAPI/CelesteTAS-EverestInterop , allowing Celeste TASers to submit movies to the site without having to convert them to work with libTAS first.

Timing is reported in IGT, as such the platform framerate on the site will need to be 1000 rather than 60 in order to correctly parse movie length.

CelesteTAS movies uses .tas extensions, which were noted as being generic enough to potentially cause issues down the line. For this reason, users will need to change the movie's file extension to ".celtas" when submitting.


vadosnaprimer:
Do they have a spec for movie format?

InputEvelution:
It looks like Celeste Studio actually has a feature to integrate multiple .tas files into a single one, so this PR could be reworked to rely on that instead of using a zip.

PR #2241, opened 2025-12-04
Add Registration Bot Trap

Masterjun3:
This adds a registration field called "Home Page" that is invisible. When filled out it cancels the registration attempt and IP bans the user.


vadosnaprimer:
It was mentioned that browsers may autofill this field. Is that realistic and is there a way out?

Masterjun3:
I don't know about autofill.
For a way out, those who have the IP Ban permission can remove the IP again.

adelikat (approved):

PR #2248, opened 2026-01-11
Fix handling of closing tags inside of noparse and code blocks.

### LINKED ISSUE #2247: Forum BBCode: `[noparse]` does not prevent parsing of closing tags
This forum text works as expected, using `[noparse]` to escape what would otherwise be interpreted as BBCode:

```
example: [noparse][b]hello[/b][/noparse]
```

It renders as:

> example: [b]hello[/b]

But suppose we wanted to make the example itself bold:

```
example: [b][noparse][b]hello[/b][/noparse][/b]
```

The intended rendering is:

> example: **[b]hello[/b]**

But the actual rendering is:

> example: **[b]hello**[/noparse][/b]

The problem is that `[noparse]` does not prevent the inner closing tag `[/b]` from being [handled](https://github.com/TASVideos/tasvideos/blob/de99c4233e3457c3abb526b5355eb5cea85c90e5/TASVideos.ForumEngine/BbParser.cs#L313). When there's no matching outer opening `[b]`, it's fine: the `[/b]` just [gets copied to the output](https://github.com/TASVideos/tasvideos/blob/de99c4233e3457c3abb526b5355eb5cea85c90e5/TASVideos.ForumEngine/BbParser.cs#L342). But when there is an outer opening `[b]`, the inner `[/b]` matches it, and terminates both the `[b]` and the `[noparse]`.

nattthebear:
Closes #2247.

Normally, child tags inside of `[noparse]` are ignored.  But when a closing tag matched an opening tag outside of the `[noparse]`, we still parsed it.  So in this instance, the `[/b]` would close the `[b]`, also closing the `[noparse]` inside of it, and leading to a mess:

```
[b][noparse]Hello [/b][/noparse]
```

The logic that allows closing tags to "skip over" other tags and close distant things was overly aggressive and should not have been applied here.  To fix this, we don't allow skipping over to close when inside a tag with no children allowed.  This is a breaking change, but the new behavior seems preferable.

I don't have a full localdev setup right now, so I tested this only with a few unit tests, covering both the case we want to fix, and one regression scenario that seemed important.

I also admit that I haven't written C# in a while, and am not up to date on much that's going on in this language 😬.  My brain is usually full of the Typescript that I get paid to write 😆.  So the code feels a bit awkward and probably doesn't use any of the new features appropriately.

### DIFF
--- a/TASVideos.ForumEngine/BbParser.cs
+++ b/TASVideos.ForumEngine/BbParser.cs
@@ -244,13 +244,7 @@ public partial class BbParser
 
 	private Element? FindMatchingForPop(string name)
 	{
-		var canPierce = true;
-		if (KnownTags.TryGetValue(_stack.Peek().Name, out var state) && state.Children == TagInfo.ChildrenAllowed.No)
-		{
-			canPierce = false;
-		}
-
-		if (canPierce)
+		if (ChildrenExpected())
 		{
 			return _stack.FirstOrDefault(elt => elt.Name == name);
 		}
--- a/tests/TASVideos.ForumEngineTests/BbParserTests.cs
+++ b/tests/TASVideos.ForumEngineTests/BbParserTests.cs
@@ -143,4 +143,24 @@ public class BbParserTests
 		var actual = await ParseBbcodeString(input);
 		Assert.AreEqual(expected, actual);
 	}
+
+	[TestMethod]
+	public async Task CloseConditionalChildYes()
+	{
+		const string input = "[b][url=http://foobar]wow[/b][/url]";
+		const string expected = "<b><a href=\"http://foobar\" rel=\"noopener external\">wow</a></b>[/url]";
+
+		var actual = await ParseBbcodeString(input);
+		Assert.AreEqual(expected, actual);
+	}
+
+	[TestMethod]
+	public async Task CloseConditionalChildNo()
+	{
+		const string input = "[b][url]http://foobar[/b][/url]";
+		const string expected = "<b><a href=\"http://foobar[/b]\" rel=\"noopener external\">http://foobar[/b]</a></b>";
+
+		var actual = await ParseBbcodeString(input);
+		Assert.AreEqual(expected, actual);
+	}
 }


TASVideos.ForumEngine/BbParser.cs
```diff
@@ -242,6 +242,25 @@ private bool TagHasValidParent(string nextTag, TagInfo nextTagInfo, out Element
 		return nextTagInfo.RequiredParent == null || nextTagInfo.RequiredParent == popTo.Name;
 	}
 
+	private Element? FindMatchingForPop(string name)
+	{
+		var canPierce = true;
+		if (KnownTags.TryGetValue(_stack.Peek().Name, out var state) && state.Children == TagInfo.ChildrenAllowed.No)
```
SandTAS: The logic of this line mostly duplicates the `ChildrenExpected` function, except in the case of `TagInfo.ChildrenAllowed.IfParam`. Here, `TagInfo.ChildrenAllowed.IfParam` tags are considered pierceable whether or not they have a param. Is that intended? Should this condition rather be a call to `ChildrenExpected`?

In any case, I recommend adding test cases involving `IfParam` tags, with and without params set. For example, `[b][post]12345[/b][/post][/b]` should become `<b><a href="/Forum/Topics/12345[/b]">Thread #12345 ...</a></b>`, that seems clear (ignoring that the link doesn't make sense). With a param, `[b][post=12345]abc[/b]def[/post][/b]`, it depends on whether you consider the `[post]` pierceable or not: either `<b><a href="/Forum/Topics/12345">abc</a></b>def[/post][/b]` or `<b><a href="/Forum/Topics/12345">abc[/b]def</a></b>`. (The extraneous `[/post][/b]` makes the first interpretation seem obviously wrong here, but it may also be that someone wrote that first `[/b]` with the intention of also closing the inner `[post]`, something like `in [b][post=12345]abc[/b] you can read…`.)

SandTAS: > `[b][post]12345[/b][/post][/b]` should become `<b><a href="/Forum/Topics/12345[/b]">Thread #12345 ...</a></b>`, that seems clear

Now that I think about it more, it's not so clear. The parser wouldn't be able to look up a thread when the ID is malformed like that. Maybe `[url]` is a better example of an `IfParam` tag, since its format is less constrained and a URL might legitimately contain constructs with `[` and `]`.

SandTAS: Looks nice! Thanks for the discussion.

PR #2249, opened 2026-01-15
Upgrade to .NET 10

Masterjun3:
This updates our project to use .NET 10.
The Nuget packages are also updated to .NET 10, and I took the opportunity to update all other Nuget packages as well (there were no breaking changes anywhere).

This PR introduces new requirements for the solution and the projects:
- The .NET 10 SDK is required now (obviously).
- For Visual Studio users, this now requires the new Visual Studio 2026 instead of 2022.

I will update our GitHub Wiki articles to the new requirements once this is merged.

There is also a new EF Core migration for our DB in there. This adds a few indexes and some foreign keys that should have been there in the first place.
If this migration fails on our production db, it means our data is broken and we have some entries that are supposed to be deleted (like e.g. a `role_claims` entry with a `role_id` that does not exist).

This PR like it is now probably has some new warnings:
- There are OpenApi changes that deprecate some methods we use.
- I also assume because of the update of the `Microsoft.CodeAnalysis.CSharp` nuget package that we will have some new warnings from that.
- Maybe also the Tests.

I can take a look at the warnings, but I wanted to make the PR first so that all the tests run on GitHub.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -25,7 +25,7 @@
     <PackageVersion Include="Microsoft.IdentityModel.JsonWebTokens" Version="8.15.0" />
     <PackageVersion Include="Microsoft.IO.RecyclableMemoryStream" Version="3.0.1" />
     <PackageVersion Include="Microsoft.Web.LibraryManager.Build" Version="3.0.71" />
-    <PackageVersion Include="MSTest" Version="3.11.1" />
+    <PackageVersion Include="MSTest" Version="3.8.3" />
     <PackageVersion Include="Nager.Country" Version="4.3.5" />
     <PackageVersion Include="Namotion.Reflection" Version="3.4.3" />
     <PackageVersion Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="10.0.0" />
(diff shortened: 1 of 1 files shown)


.github/workflows/dotnet-core.yml
```diff
@@ -40,9 +40,12 @@ jobs:
...
     - name: Test
-      run: dotnet test --no-restore --verbosity normal --collect:'Code Coverage' --filter:"FullyQualifiedName!~TASVideos.E2E.Tests"
-    - name: Generate coverage report
-      run: |
-        dotnet tool run dotnet-coverage merge 'tests/*/TestResults/**/*.coverage' --output-format=cobertura --output=merged.coverage.xml
-        reportgenerator -reports:merged.coverage.xml -targetdir:. -reporttypes:MarkdownSummaryGithub
-        cat SummaryGithub.md >>"$GITHUB_STEP_SUMMARY"
+      run: dotnet test --no-restore --verbosity normal -p:TestingPlatformCommandLineArguments="--filter FullyQualifiedName!~TASVideos.E2E.Tests --settings .runsettings"
+
+    # The following steps were for code coverage, but do not work anymore since .NET 10 changed the dotnet test command from VSTest to Microsoft.Testing.Platform (MTP).
```
YoshiRulz: OH, yeah don't use that. Removing `global.json` should bring back the old runner.

Masterjun3: That didn't work, see the [github actions test](https://github.com/TASVideos/tasvideos/actions/runs/21047328701/job/60524960384) in the commit before I added global.json. The errors are these:
```
"/home/runner/work/tasvideos/tasvideos/TASVideos.sln" (VSTest target) (1) ->
"/home/runner/work/tasvideos/tasvideos/tests/TASVideos.Data.Tests/TASVideos.Data.Tests.csproj" (VSTest target) (12) ->
/home/runner/.nuget/packages/microsoft.testing.platform.msbuild/2.0.2/buildMultiTargeting/Microsoft.Testing.Platform.MSBuild.targets(263,5): error : Testing with VSTest target is no longer supported by Microsoft.Testing.Platform on .NET 10 SDK and later. If you use dotnet test, you should opt-in to the new dotnet test experience. For more information, see https://aka.ms/dotnet-test-mtp-error [/home/runner/work/tasvideos/tasvideos/tests/TASVideos.Data.Tests/TASVideos.Data.Tests.csproj]
```
In particular the line
> Testing with VSTest target is no longer supported by Microsoft.Testing.Platform on .NET 10 SDK and later.

PR #2251, opened 2026-01-19
Simplify a bunch of test assertions, and update MSTest

YoshiRulz:


YoshiRulz:
A bunch more warnings, then cleaning up the diff. This was just my day's progress.

tests/TASVideos.Core.Tests/Services/AwardsTests.cs
```diff
@@ -103,11 +103,11 @@ public async Task ForPublication_PublicationAward_ReturnsAward()
 
 		Assert.IsNotNull(actual);
 		var list = actual.ToList();
-		Assert.AreEqual(1, list.Count);
+		Assert.HasCount(1, list);
 		var actualPubAward = list.Single();
```
YoshiRulz: This could be refactored to `Assert.IsTrue(actual is [var actualPubAward]);`, or I could have made a helper, but I chose to do neither because this PR is already big enough.

PR #2252, opened 2026-01-19
Resolve remaining warnings from updating to .NET 10

YoshiRulz:


TASVideos.Api/Endpoints/TagsEndpoints.cs
```diff
@@ -55,12 +55,12 @@ await tagService.GetById(id)))
 			};
 		})
 		.WithSummary("Creates a new tag")
-		.WithOpenApi(g =>
+		.AddOpenApiOperationTransformer((g, _, _) =>
```
YoshiRulz: Not sure about this, just copied the example.

tests/TASVideos.IntegrationTests/TASVideosWebApplicationFactory.cs
```diff
@@ -134,10 +134,12 @@ FROM pg_stat_activity
 			terminateCommand.Parameters.AddWithValue(_testDatabaseName!);
 			terminateCommand.ExecuteNonQuery();
 
+#pragma warning disable CA2100
```
adelikat: is there not a code way to let it know that this code is safe?

Issue #2245, opened 2026-01-04
The Account/EmailConfirmationSent page is broken


Masterjun3:
The page https://tasvideos.org/Account/EmailConfirmationSent (note, can only be accessed if logged out) is currently not working at all.
The issue might be the recaptcha not being wired up correctly.

VelpaChallenger:
It's a production-only issue, I cannot reproduce it locally, not even after enabling reCaptcha. Still, I found an issue with the onSubmit callback so I created #2262.

PR #2262, opened 2026-02-14
Potential fixes for #2245

VelpaChallenger:
I really don't think this is related to the resend button not working in production but in any case it's still an issue. CSP also doesn't seem to be the issue as I'm able to reproduce those in my local environment but the button still works. So the investigation continues.

### DIFF
--- a/TASVideos/Pages/Account/EmailConfirmationSent.cshtml
+++ b/TASVideos/Pages/Account/EmailConfirmationSent.cshtml
@@ -6,7 +6,7 @@
 }
 
 @await Html.RenderWiki(SystemWiki.EmailConfirmationSentMessage)
-<form client-side-validation="true" method="post">
+<form client-side-validation="true" method="post" id="email-confirmation-sent-form">
 	<row>
 		<column md="6">
 			<fieldset>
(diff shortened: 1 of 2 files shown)


Masterjun3:
Note that this does not check the Recaptcha value on the server at all, so it will go through even if it's invalid.

VelpaChallenger:
> Note that this does not check the Recaptcha value on the server at all, so it will go through even if it's invalid.

In that case, we could remove it altogether, right? I believe the form check is intended to run on the client (based on the commit comments, since it wasn't part of my changes). I'm not very familiar with recaptcha, but otherwise I believe maybe this is what would be needed for a full client/server check?

```csharp
		var encodedResponse = Request.Form["g-recaptcha-response"];
		var isCaptchaValid = await reCaptchaService.VerifyAsync(encodedResponse);

		if (!env.IsDevelopment() && !isCaptchaValid)
		{
			ModelState.AddModelError("", "TASVideos prefers human users.  If you believe you have received this message in error, please contact admin@tasvideos.org");
		}
```

It's code from `Register.cshtml.cs` (most of the feature was based on that page).

Masterjun3:
Closing this PR in favor of #2270 .

PR #2266, opened 2026-02-19
Disable rule SA1412 "Store files as UTF-8 with byte order mark"

YoshiRulz:
There is no reason to be using this backwards-compatibility feature in $CURRENTYEAR. So devs are getting their code flagged even if it's perfectly functional and readable, and the change asked of them is completely invisible, not even affecting whitespace.

I have not removed the BOMs from every file because it's not worth the time or effort.


adelikat (approved):

YoshiRulz:
In case you need it in future, it's down here in Rider:
Screenshot from 2026-03-01 01-39-44

PR #2287, opened 2026-03-17
Further distinguish hyperlinks and submenus (non-hyperlinks) in the navbar

YoshiRulz:
Follow-up to #2283, though the earlier iterations of the navbar suffered more or less from the same problem.
Live site:
Screenshot_20260317_134310
With this changeset:
Screenshot_20260317_134329
The arrows are slightly larger and on the baseline, and for desktop users, the hover effect is a dashed line instead of solid because I couldn't think of anything else.

Other changes I considered but have not included:
- italics for submenus (looked bad IMO)
- `cursor: context-menu;` for submenus (I think it's intended only for context menus? also it would only help desktop users, and they already have the underline on hover)
- even higher `font-weight` for hyperlinks / lower `font-weight` for submenus (it would need to be like 500 vs 800 to be distinct, which looked bad IMO)
- drop shadows (might help since the contrast is a bit low, but the `color` isn't fully opaque so a naive implementation won't work)

### DIFF
--- a/TASVideos/TagHelpers/NavbarTagHelpers.cs
+++ b/TASVideos/TagHelpers/NavbarTagHelpers.cs
@@ -51,7 +51,7 @@ public class NavDropdownTagHelper : NavItemBase
 		}
 
 		output.Content.SetHtmlContent(
-			$"<a href='#' class='{addClass}' data-bs-toggle='dropdown'>{Name ?? ""}<span class='caret'></span></a>");
+			$"<a href='#' class='{addClass}' data-bs-toggle='dropdown'>{Name ?? ""}</a>");
 
 		output.Content.AppendHtml($"<div class='dropdown-menu'>{content}</div>");
 	}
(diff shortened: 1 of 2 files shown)


TASVideos/TagHelpers/NavbarTagHelpers.cs
```diff
@@ -51,7 +51,7 @@ public override async Task ProcessAsync(TagHelperContext context, TagHelperOutpu
 		}
 
 		output.Content.SetHtmlContent(
-			$"<a href='#' class='{addClass}' data-bs-toggle='dropdown'>{Name ?? ""}<span class='caret'></span></a>");
+			$"<button class='{addClass}' data-bs-toggle='dropdown'>{Name ?? ""}</button>");
```
YoshiRulz:
```suggestion
			$"<button type='button' class='{addClass}' data-bs-toggle='dropdown'>{Name ?? ""}</button>");
```
Not sure why it was working without this, probably detected and fixed by Bootstrap.

Masterjun3:
I'm closing this PR. The changes are not justified, and there are problems as I mentioned in the previous comments.

YoshiRulz:
Can you at least take the first 2 commits? edit: plus the `<button type='button'/>` fix mentioned in a comment edit edit: #2311

PR #2288, opened 2026-03-18
First redesign of `/UserFiles` landing page

YoshiRulz:
It doesn't really come across without a lot more sample data, but it will look something like this:
Screenshot_20260318_233454
...which compacts the page by a lot.

### DIFF
--- a/TASVideos/Pages/UserFiles/Index.cshtml.cs
+++ b/TASVideos/Pages/UserFiles/Index.cshtml.cs
@@ -1,10 +1,12 @@
+using System.Text;
+
 namespace TASVideos.Pages.UserFiles;
 
 [AllowAnonymous]
 public class IndexModel(ApplicationDbContext db, IExternalMediaPublisher publisher) : BasePageModel
 {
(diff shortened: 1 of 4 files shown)


TASVideos/Pages/UserFiles/Index.cshtml.cs
```diff
@@ -3,32 +3,47 @@ namespace TASVideos.Pages.UserFiles;
...
 				g.DisplayName,
 				g.UserFiles.Select(uf => uf.UploadTimestamp).ToList()))
 			.ToListAsync();
+		GroupedGames = gamesWithMovies
+			.GroupBy(g => char.IsAscii(g.GameName[0])
+				? char.IsAsciiLetterOrDigit(g.GameName[0])
+					? char.IsAsciiDigit(g.GameName[0])
+						? '#'
+						: char.ToUpperInvariant(g.GameName[0])
+					: '$'
```
YoshiRulz:
```suggestion
					: '@'
```
forgot to update this

Masterjun3 (approved):

PR #2299, opened 2026-04-29
Fix System editing

Masterjun3:
System editing was fully broken.

We ran into two simultaneous bugs here, originating from the commit that converted a lot of classes into records.

First bug: The original bound property (with [BindProperty]) was a `SystemsResponse`, which was a record with an IEnumerable inside. In the .cshtml we didn't even put the IEnumerable into a form, so the model binder just set it to null. In the OnPost the model was then invalid because that field was "required".

Second bug: The OnPost model is invalid, so it tries to return the page. However, the model binder set the non-nullable IEnumerable to null (which might be a big downside to using records). The .cshtml tries to iterate the null value and crashes.

I fixed these bugs by making a new record type just for the editing flow, and putting the IEnumerable separately.

(Btw, technically we still have a third bug: After OnPost returns, the page loses the information about the "InUse" and "SystemFrameRates" properties. But that's non-critical, and I don't know how to best fix that anyway.)

### DIFF
--- a/TASVideos/Pages/Systems/Create.cshtml
+++ b/TASVideos/Pages/Systems/Create.cshtml
@@ -26,7 +26,6 @@
 	</row>
 	<form-button-bar>
 		<submit-button edit="false"></submit-button>
-		<submit-button></submit-button>
 		<cancel-link asp-page="Index"></cancel-link>
 	</form-button-bar>
 </form>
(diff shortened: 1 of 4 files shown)


vadosnaprimer (approved):

Masterjun3:
This is basic functionality that we have tests for. But none of the tests use the model binder, so I suppose it would have only been caught by an E2E test? But I didn't see any POST tests in there, so I don't know how to write tests like that properly.

PR #2302, opened 2026-05-01
Always use InvariantCulture when passing DateTime around for ExceededLimit

CasualPokePlayer:


Masterjun3 (approved):
Usually in this code base we use "is not null" instead of ".HasValue", but it's fine.

PR #2304, opened 2026-05-01
merge doom+windows+linux into pc

vadosnaprimer:
https://tasvideos.org/Forum/Posts/542364

remove my old edginess from test file

unsure if I should also rename the ltm test files to say pc instead of linux


Masterjun3:
Are the merged SystemCodes unused now? Maybe removing them would be wise so no future parsers use them?

vadosnaprimer:
DOOM and Windows are now PC on the site, but Linux game versions still need to be moved, and I was hoping for automated way because we want to also update pub/sub titles. Not sure about causing yt upsates for them tho...

CasualPokePlayer:
> DOOM and Windows are now PC on the site, but Linux game versions still need to be moved, and I was hoping for automated way because we want to also update pub/sub titles. Not sure about causing yt upsates for them tho...

That shoudn't matter for removing entries in the SystemCodes class (as that's just a parser detail), it's not like the site automatically syncs platforms with that (that's just rather handled with a standard db)

Masterjun3:
Oh yeah, I meant "SystemCode unused" in the sense of the *code* not having any references to them anymore. If we want no future parsers to use those systems, their entries should be removed.
All existing movies are unaffected by editing the SystemCode class around as we see fit.

Masterjun3 (approved):

PR #2311, opened 2026-05-03
Make navbar submenus (non-hyperlink items) into `<button/>`s

YoshiRulz:
From #2287:
> Them being `<a/>`s in the first place *is* the a11y concern, because [links](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#onclick_events) [aren't](https://webaim.org/techniques/hypertext/#keyboard) [buttons](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/button_role#accessibility_concerns). Starting with that as a first principle, you can then style them to look similar, which Bootstrap does.

### DIFF
--- a/TASVideos/TagHelpers/NavbarTagHelpers.cs
+++ b/TASVideos/TagHelpers/NavbarTagHelpers.cs
@@ -51,7 +51,7 @@ public class NavDropdownTagHelper : NavItemBase
 		}
 
 		output.Content.SetHtmlContent(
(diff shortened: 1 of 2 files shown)


Masterjun3:
Remove the underline style change and I can merge this.

YoshiRulz:
It's setting the decoration to a solid underline to match the `<a/>`s.

Masterjun3:
Oh wait you're right I thought this was a different change! Sorry.

Masterjun3 (approved):

PR #2312, opened 2026-05-03
Improve diff UX

YoshiRulz:
Before and after:
Screenshot from 2026-05-04 00-23-14

### DIFF
--- a/TASVideos/wwwroot/css/diff.scss
+++ b/TASVideos/wwwroot/css/diff.scss
@@ -37,6 +37,7 @@
 		font-weight: bold;
 		padding: 0 0.5em;
 		text-align: right;
+		user-select: none;
 	}
 
 	.contracter {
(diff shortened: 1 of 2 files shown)


vadosnaprimer:
How does it show complex stuff like https://tasvideos.org/Wiki/PageHistory?path=MovieRules&fromRevision=65&toRevision=66

YoshiRulz:
Better in some ways, worse in others.
comparison

vadosnaprimer:
Back when we were tuning the current setting it became clear it's always a tradeoff. I don't think tweaking it further is gonna improve things overall. I don't mind e0ae742e87ed806da2843cde2a42d49e9b0b6edc but it's not like we ever need to copypase diffs.

Masterjun3:
Line numbers unselectable is good!
The change in algorithms might need to be discussed or tested more. In fact, in issue #987 in comment https://github.com/TASVideos/tasvideos/issues/987#issuecomment-1019345557 feos already tested stuff and decided Efficiency was better than Semantic.
For what it's worth, I agree. The Semantic one creates too many small edits for my liking.

YoshiRulz:
Before I looked at the code and saw what library it used, my idea was to add a post-processing step which "slid" the `<ins/>` window back to a newline where possible (see example OP), so I could revisit that.

Masterjun3 (approved):

PR #2322, opened 2026-05-08
Add additional check against chaining redirects

bigbass1997:
Fixed a minor oversight when creating redirects.

If you already have a `B -> C` redirect:
* the form would prevent you from creating an `A -> B` redirect
* but it _wouldn't_ prevent you from creating a `C -> D` redirect

This PR adds a check to prevent the latter scenario.

### DIFF
--- a/TASVideos/Pages/Wiki/Redirects/Create.cshtml.cs
+++ b/TASVideos/Pages/Wiki/Redirects/Create.cshtml.cs
@@ -35,7 +35,13 @@ public class CreateModel(ApplicationDbContext db) : BasePageModel
 
 		if (await db.WikiRedirects.AnyAsync(r => r.PageNameFrom == WikiRedirect.PageNameTo))
 		{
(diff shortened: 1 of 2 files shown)


bigbass1997:
Oh, also, I changed the error message slightly to better clarify (imo) what the user should avoid.

Masterjun3:
This wasn't quite an oversight. There is technically nothing wrong with redirect chains. But it makes sense to prevent them, if we want to.

Btw, did you make sure this also is checked on the Edit.cshtml.cs page?

bigbass1997:
Ah alright, I just figured since the other check existed that we might want to avoid chaining them. I didn't test it, but I could see a potential issue if a loop was accidentally created. Like A -> B -> C -> A.

And no I forgot to check the edit page, give me a moment.

Masterjun3:
> but I could see a potential issue if a loop was accidentally created. Like A -> B -> C -> A

But that is already prevented! I wrote in the previous PR how these loops are prevented.

bigbass1997:
That's good then! I guess there's no technical harm in chains, but might be helpful in the long-term for editors to avoid them anyways.

Masterjun3 (approved):

PR #2325, opened 2026-05-12
Refactor wiki redirects into service and add wiki redirect tests

Masterjun3:
This PR adds a wiki redirect service so that we can test it, and then adds tests.


Masterjun3:
Reason for draft: I noticed a problem. Moving wiki pages and leaving a redirect now can cause an inconsistent DB state (when the redirect succeeds but the move fails).
Maybe I'll have to work with transactions here.

adelikat (approved):

PR #2327, opened 2026-05-12
Change topicfeed module to default to topic's title

### LINKED ISSUE #2326: News Module only pulls from subject for headings, problem if a post is the OP
Posts on the News Module use the subject line for headings. This normally works fine. However, if the first post of a thread is displayed on the module, no title is displayed at all as opening posts cannot have post subjects. Recommended action: pull from thread title if subject does not or cannot exist.

bigbass1997:
Resolves #2326 

The topicfeed module lists the posts in one or more topics. It titles each post using the subject line. However, the first post in a topic cannot have a subject line causing the module to display nothing.

This PR changes the module to default to the topic's title if the subject line is blank.

### DIFF
--- a/TASVideos/WikiModules/TopicFeed.cshtml
+++ b/TASVideos/WikiModules/TopicFeed.cshtml
@@ -10,8 +10,8 @@
 		var collapseId = $"collapse-content-{post.Id}";
 		<div class="card-body">
 			<div>
-				<h5 condition="!Model.HideContent"><a href="/Forum/Posts/@(post.Id)">@post.Subject</a></h5>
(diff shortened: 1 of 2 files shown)


Masterjun3 (approved):
@bigbass1997 I have commited some changes to your PR.

One problem was you using an external library `NUglify.Helpers` for something as simple as checking for null or whitespace, which is built in. Avoiding dependence always makes sense.

Secondly, calling external methods in a DB Query of all things is quite brittle.
The generated SQL just returned both the subject and the title and then it instructed the C# code to execute the external method on all results. Which is fine like this, but any additional conditions on the subject or a minimally more complex query would lead to a "could not be translated" Exception.
It's more robust here to return both values and handle sanity checks afterwards.

bigbass1997:
Sorry didn't even notice the external library; saw a similar `IsNullOrWhiteSpace` used all over the codebase, and `IfNullOrWhiteSpace` was seemingly an available option.

Issue #1539, opened 2023-03-31
A missing framerate entry causes submission to break (and also breaks all pages trying to show submission)


Masterjun3:
When a submission is created, (and our parser doesn't override the framerate) we match the framerate exactly to the system and region.

https://github.com/TASVideos/tasvideos/blob/b693cc7aefe18a42b899181401d3d20e3682e490/TASVideos.Core/Services/QueueService.cs#L435-L440

If this doesn't exist, the field is empty and breaks the title generation.
<hr>

To avoid breaking the site, we have to either set a default framerate if the framerate wasn't found, or we have to make sure that fitting framerate entries are always created when creating a new system.

vadosnaprimer:
I can't remember why we didn't have a suitable framerate in that case. What submission was it?

Masterjun3:
I think the region was set to "World", but we only had "NTSC" and "PAL" regions, or something like that.

vadosnaprimer:
So if we want to use `World` or `Any` keywords for framerates that are not TV based, we should not only create them but to also update all the affected movie parsers?

PR #2348, opened 2026-06-27
region fixes for framerate creation

vadosnaprimer:
see #1539

- when override is used during movie parsing, don't uppercase region code (because `World` should not be)
- when user creates framerate, use region enum dropdown instead of arbitrary string

TODO: figure out editing a framerate (should we accept the string and then try to match it to enum values, or is it better to change the field itself to enum globally?)

unrelated tiny tweak for GUI: fix SHA-1 and MD5 displayed names


vadosnaprimer:
the TODO is now done and tested

adelikat (approved):

PR #2350, opened 2026-07-06
Layout improvements to Game Versions table

YoshiRulz:
See https://github.com/TASVideos/tasvideos/pull/2350#issuecomment-4915136617 for a screenshot of the latest revision.
~~Before and after~~:
Screenshot from 2026-07-07 04-18-02
~~Slightly easier to read, but I'm sure it can be improved further. Maybe some colour? I thought of [this xkcd](https://xkcd.com/1286/).~~

### DIFF
--- a/TASVideos/Pages/Shared/_GameVersionTableRow.cshtml
+++ b/TASVideos/Pages/Shared/_GameVersionTableRow.cshtml
@@ -1,0 +1,35 @@
+@using TASVideos.Data.Entity.Game
+@model TASVideos.Pages.Games.Versions.ListModel.VersionEntry
+
+<tr>
+	<td>
+		<span class="text-nowrap"><i condition="@Model.Type == VersionTypes.Unknown" class="fa fa-circle-question"></i> @Model.Type</span>
+	</td>
+	<td>
+		@Model.Name
+		@if (!string.IsNullOrWhiteSpace(Model.SourceDb) && ViewData.ContainsKey("ShowSourceDB"))
+		{
+			<br /><span>SourceDB: @Model.SourceDb</span>
+		}
+	</td>
+	<td>@Model.TitleOverride</td>
+	<td>@Model.Region</td>
+	<td>@Model.Version</td>
+	<td>@Model.System</td>
+	<td style="overflow-wrap: anywhere">
+		<small condition="!string.IsNullOrEmpty(Model.Sha1)">SHA-1: @Model.Sha1</small><br />
+		<small condition="!string.IsNullOrEmpty(Model.Md5)">MD5: @Model.Md5</small>
+	</td>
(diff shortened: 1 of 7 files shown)


vadosnaprimer:
All I'm seeing is hashes are now monospace (and uppercase).

TASVideos/wwwroot/css/partials/_customizations.scss
```diff
@@ -720,3 +720,16 @@ top-button-bar {
 		padding-left: 1rem;
 	}
 }
+
+.game-version-checksums {
+	& th {
+		font-weight: inherit;
+		text-align: end;
```
YoshiRulz: ins `opacity: 0.8;` or so

YoshiRulz (commented):
I also aligned the labels horizontally, though I now think that was a mistake and only lines *within* a value (i.e. when it's wrapped) should be aligned with each other.
edit: Mockup of what it would mean to misalign the labels again, hopefully it's clear what I'm going for with this:
617734265-66736b49-b342-40c3-ac93-228fa158eb08
If all else fails, I could just use a slightly different colour for MD5...

TASVideos/Pages/Shared/_GameVersionTableRow.cshtml
```diff
@@ -0,0 +1,36 @@
...
+			<br /><span>SourceDB: @Model.SourceDb</span>
+		}
+	</td>
+	<td>@Model.TitleOverride</td>
+	<td>@Model.Region</td>
+	<td>@Model.Version</td>
+	<td>@Model.System</td>
+	<td><table class="game-version-checksums">
+		@* re: wbr, we could simply insert a U+200B ZWSP in the middle of the string, but then it would be there when copied *@
+		<tr condition="!string.IsNullOrEmpty(Model.Sha1)"><th>SHA1:</th><td><samp>@Model.Sha1![..20]<wbr>@Model.Sha1![20..]</samp></td></tr>
```
vadosnaprimer: It's called `SHA-1` not `SHA1`.

vadosnaprimer:
How does it look on a narrow screen?

YoshiRulz:
Showcases narrower viewport, new layout/alignment, new colouring, and new sort order:
Screenshot from 2026-07-08 23-09-32

vadosnaprimer:
What does the color mean?

YoshiRulz:
It has no semantic meaning (it's calculated from the hash), it's only to help you distinguish (= ignore) values without reading them. For example, the top value is cyan, and on my screen I can see that colour doesn't appear again, so there mustn't be a duplicate.

vadosnaprimer:
I don't think we should even have duplicates. If it's a version, no matter the name it remains the same file and therefore is supposed to be the same version. If different databases call it differently, we can add that to version notes. Or is this feature there to help detecting those duplicates in the first place?

PR #2357, opened 2026-07-12
Limit `float`'d image sizes on wiki pages and game entries

YoshiRulz:
Works with Bootstrap's "Containers" (stepped fixed-width body, which was the crux of the problem).

### DIFF
--- a/TASVideos/wwwroot/css/partials/_customizations.scss
+++ b/TASVideos/wwwroot/css/partials/_customizations.scss
@@ -1,7 +1,6 @@
 /// <reference path="../bootstrap.scss" />
 /// <reference path="colors.scss" />
 /// <reference path="variables.scss" />
-@use "sass:map";
 
 :root {
 	--tasvideos-bg: #{$tasvideos-bg};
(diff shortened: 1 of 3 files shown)


vadosnaprimer:
Screenshots?

vadosnaprimer:
> Screenshots?

Screenshots?

YoshiRulz:
It can't really be captured in a screenshot. I can demo it in Discord if your IDE is broken.

Masterjun3:
I'll check this out in a few days. But meanwhile if you want, for screenshots you can do a before and after for mobile widths (~410px) and desktop window widths (~1280px).

vadosnaprimer:
> It can't really be captured in a screenshot. I can demo it in Discord if your IDE is broken.

Screenshots:

maximum width, full image, looks great

изображение

width getting smaller and smaller, image too, looks ok, tho ideally the image would only downscale during the smallest width step

изображение

shrink a little bit more, image is back to normal size but it forgot to jump above the text

изображение

just a little bit smaller

изображение

this and smaller and the text finally fully moves below, tho the image remains right aligned

изображение

Ideal workflow for this whole thing would be image sitting on the right if the text has enough space (probably possible to set min width for text), and if it all stops fitting into 2 columns, the image would instead go above text. And if width keeps getting smaller, it's ok to start downscaling the image. Unfortunately I don't know how to make it all happen, nor if it's even possible.

YoshiRulz:
https://github.com/TASVideos/tasvideos/blob/28e8d0da14efe90c8089e1e4ae4d95a9cc8d19eb/TASVideos/wwwroot/css/partials/_customizations.scss#L240-L243

TASVideos/wwwroot/css/partials/_customizations.scss
```diff
@@ -228,6 +229,22 @@ body {
...
+	@media (min-width: $width) {
+		// doesn't look good in motion, but this does its job of keeping the accompanying text wide enough to read comfortably
+		.container :is(.embedright, .embedleft) {
+			max-width: calc($width * 2 / 5);
+		}
+	}
+}
+// this is an override for the "extra small" size, which isn't in the map -_-
+@media (max-width: #{map.get($container-max-widths, "sm") - 1px}) {
+	.container .embedright, .container .embedleft {
```
YoshiRulz:
```suggestion
	.container :is(.embedright, .embedleft) {
```
to match above

Masterjun3:
Hmm, so your approach looks bad in mobile because it scales up images if they're too small, even 16x16 icons are suddenly massive and don't look good.
Also, I understand trying to use bootstrap things in the scss code because we rely on bootstrap, but I feel like the task is simple enough to not need this complex looking code.
I'm making a commit that should make this a bit simpler I think.

Masterjun3:
After this commit, it now looks like this (all the different sizes):
image

image

Masterjun3 (approved):
If nobody has complaints with the latest changes (see the images), I think this PR is ready to merge.

vadosnaprimer (approved):
Tested in dev env, looks great! I don't mind image scaling if text is always good.

YoshiRulz (commented):
I agree the new approach gives better results than what I had.

PR #2364, opened 2026-07-19
Show warning when embedding an image from one of the banned hosts

YoshiRulz:
Screenshot 2026-07-20 at 02-19-30 Editing SandBox - TASVideos
It only appears when rendering within the wiki/forum preview pane, not on the wiki page / thread itself.

### DIFF
--- a/TASVideos.Common/Constants.cs
+++ b/TASVideos.Common/Constants.cs
@@ -82,6 +82,25 @@ public static class SiteGlobalConstants
 	public const int YearsOfBanDisplayedAsIndefinite = 2;
 
 	public const string MainIvatarDomain = "seccdn.libravatar.org"; // keep in sync with profile-settings.js
+
+	// Hardcoded for now, we can make a database table if this becomes a maintenance burden
+	public static readonly string[] BannedAvatarSites = [
+		"cdn.discordapp.com",
+		"media.discordapp.net",
+		"membres.lycos.fr",
+		"rphaven.org",
+		"usuarios.lycos.es"
+	];
+
+	public static bool LinkIsBannedImageHost(string url)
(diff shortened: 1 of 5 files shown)


adelikat:
Why are we recommending catbox.moe? I have never heard of that (not that I would have heard about the best image hosting sites).  Still we need to decide what sites to recommend.

And is this feature desired?

YoshiRulz:
Which services to recommend absolutely needs to be decided. I checked the 2 places the banned host list is mentioned (`/Profile/Settings` and `/SiteRules`), and neither included a positive example. So I just picked the most popular, Imgur, and then catbox.moe I'd heard of as an Imgur alternative after some TOS change or something. But I now see they have recently announced a max. 2 year lifespan on uploads, making them a terrible choice.

I totally forgot to mention, the raison d'etre for making this was [this post](https://tasvideos.org/Forum/Posts/543916) where the user tried to embed from `media.discordapp.net/attachments`.

vadosnaprimer:
We don't have to decide which sites will be reliable for the next few years... because we can't. We don't control the future, so it's pointless to recommend something as "reliable". Helpful info would be showing which sites we ban instead.

Masterjun3:
Using Html.Raw with values from a completely unrelated part of the code base is one of the most bad-practice pieces of code I've seen.
This is just inviting some bad html syntax errors down the line at best, and some severe site vulnerability at worst.

If it were up to me, I'd fully shut down this approach. 

The risks of Html.Raw are not worth this feature at all.

YoshiRulz:
Moved the logic to the server side (in the forum/wiki parsers), and replaced the positive suggestions with a link to the negative list.

PR #2366, opened 2026-07-20
Allows for submissions to have an alternative optimization metric

### LINKED ISSUE #2217: Add entry field for alternative timing methods to submissions and publications, option to choose which is displayed in title
Currently, all movies on TASVideos are displayed purely in terms of the input file's parsed time (usually boot-up until final input). While this is a useful general-purpose solution, it comes with a number of drawbacks that are exacerbated for games with prominent TAS/RTA speedrun communities:

- A TAS may appear to be slower than the RTA world record to an uninformed viewer due to the longer time (e.g. SMB1 warps)
- A TAS that aims for in-game time will only display its real time completion time despite that being irrelevant, making it appear as though TASVideos doesn't want these kinds of submissions (e.g. Celeste)
- A TAS that improves the time on a real console may be slower on an emulator, disincentivizing submission since it will appear to lose time over the predecessor (e,g, SM64 1 key)

Currently, other timings are only able to be mentioned in the sub or pub description, _maybe_ in the encode if the publisher wants to include it. These issues are particularly problematic on TVC, as not only are viewers likely to decide whether they will engage with content based on first impressions alone, but they are also unlikely to check the description for more information even after watching, and tags such as "Aims for in-game time" are not preserved anywhere.

In this proposed solution, the Submit Movie page would have an additional field where the user could select a timing method from a drop-down list (IGT, RTA, Console, etc.), manually enter a time, and check a box if they would prefer it to be displayed on the submission/publication. (See this mockup from Samsara for reference:)

Image

This time would then be verified by a Reviewer or Judge, and if preferred, replace the automatically parsed time in the title.

calvarado194:
Closes #2217 . Currently a draft, not ready for review.

* Adds two new fields to submission form: Optimization Metric and Final Time/Score. Optimization Metric is the chosen metric that the movie has been optimized for, currently supporting TAS Timing, RTA Timing, In-game Timer and Maximum Score. Final Time/Score is the value of that metric by the end of the input file. Defaults to TAS Timing.
* A submission with a timing metric other than TAS Timing will use that metric for its submission title and subsequent publication title should it be published. A submission that aims for score will display both final score and TAS Time.

TODO:
* Add tests
* Validate that the metric and value are being passed correctly from sub to pub (I think I need to find a way to give myself perms for publication in dev environment lol).
* Should an update to the submission's metric affect the publication? How would we go about updating existing publications?


TASVideos.Data/Entity/OptimizationMetric.cs
```diff
@@ -0,0 +1,36 @@
...
+	[Display(Name = "In-game Timer")]
+	InGameTiming,
+
+	[Display(Name = "Maximum Score")]
+	HighScore,
+}
+
+public static class OptimizationMetricExtensions
+{
+	extension(OptimizationMetric criteria)
```
adelikat: Nice

TASVideos/Pages/Submissions/Edit.cshtml.cs
```diff
@@ -34,6 +34,15 @@ public class EditModel(
 	public List<SelectListItem> AvailableClasses { get; set; } = [];
 	public List<SelectListItem> AvailableRejectionReasons { get; set; } = [];
 
+	public List<SelectListItem> AvailableOptimizationGoals { get; set; } = Enum.GetValues<OptimizationMetric>().ToList().ConvertAll<SelectListItem>(g =>
```
adelikat: You will want to use a .Select() here instead of a .ConvertAll().  In addition generally the Enum.Getvalue() part is put into a private static value so the reflection piece of the transform is run once instead of Every page request

YoshiRulz:
```suggestion
	public static readonly List<SelectListItem> AvailableOptimizationGoals { get; set; } = Enum.GetValues<OptimizationMetric>().Select(g => new SelectListItem {
		Text = g.EnumDisplayName(),
		Value = g.ToString(),
		Selected = false,
	}).ToList();
```
You might just be able to do `asp-items="Html.GetEnumSelectList<OptimizationMetric>()"` instead.

vadosnaprimer:
> Should an update to the submission's metric affect the publication?

Those have always been separate, even if one inherited data from another.

> How would we go about updating existing publications?

Just manually maybe. Or if there's a lot of them and we don't want to waste youtube quota, a script could be used (or DB hacking), or it could be moved to cataloging side which doesnt force a youtube update (but that'd be non-ideal maybe, because submission form needs this anyway).

ALSO I'd like to suggest this idea for discussion:

- Instead of replacing time with score or showing them both automatically, maybe it's better to just include final score into branch label? We already do it for countable maximum goals, like "all 120 stars" or "all 96 exits", so "maximum score 99999" should be fine too?

Masterjun3:
This PR adds quite a lot of duplications to the publication and submission GenerateTitle methods. I consider it almost necessary to add tests for this new complexity. The existing GenerateTitle tests are in these files:
- `tests\TASVideos.Data.Tests\Entity\PublicationTests.cs`
- `tests\TASVideos.Data.Tests\Entity\SubmissionTests.cs`

It would also help with the feedback because nowhere currently can people see an actual example of what this looks like. Examples are good.

calvarado194:
Added the requested tests! Sorry, I was not sure where those tests were located.

Here's a couple examples of how the titles would look, for each relevant type of title:
```
IGT example
#1111: CelesteCommunity's PC Celeste in 21:23.330 (IGT) (sub page)
PC Celeste by CelesteCommunity in 21:23.330 (IGT) (pub page)
[TAS] PC Celeste in 21:23.330 (IGT) by CelesteCommunity (youtube)

Score Example
#1111: ikuyo's PC Sayonara Wild Hearts "maximum score" (2,116,350) in 48:23.136 (sub page)
PC Sayonara Wild Hearts "maximum score" (2,116,350) by ikuyo in 48:23.136 (pub page)
[TAS] PC Sayonara Wild Hearts "maximum score" (2,116,350) in 48:23.136 by ikuyo (youtube)
```

TAS timing titles keep their current structure.

By the way, I noticed there are no tests specific to the YouTube title generation, regardless of using this feature or not. Should I add them on this PR or do I make a separate one?

TASVideos.Core/Services/QueueService.cs
```diff
@@ -1016,6 +1023,8 @@ public record SubmitRequest(
 	string? EncodeEmbeddedLink,
 	IList<string> Authors,
 	string? ExternalAuthors,
+	OptimizationMetric Metric,
```
adelikat, line 1026: Based on the compiler warning above, it looks like Metric needs to be nullable here, and the cshtml file needs to be adjusted accordingly, if needed

adelikat (changes requested):
Requesting changes on the possible null reference assignment.  If it is just noise, I can help you suppress the warning, but it could be that the page fails to render on existing submissions where the metric has not been set

adelikat (approved):

PR #2367, opened 2026-07-20
Handle redirects in `module:brokenlinks`

### LINKED ISSUE #2365: BrokenLinks module doesn't account for redirects
If a link exists to a page that doesn't exist, but a redirect from that location exists, the `BrokenLinks` wiki module will still report it as broken.

Example: https://tasvideos.org/WelcomeToTASVideos contains a link to `Contact`. https://tasvideos.org/Contact doesn't exist, but a redirect from there to `Staff` does. So the link functions as expected, but the module claims the link is broken.

YoshiRulz:
Fixes #2365.

### DIFF
--- a/TASVideos/WikiModules/BrokenLinks.cshtml.cs
+++ b/TASVideos/WikiModules/BrokenLinks.cshtml.cs
@@ -56,7 +56,7 @@ public class BrokenLinks(ApplicationDbContext db, IWikiPages wikiPages) : WikiVi
 			.Select(c => c.Name.ToLowerInvariant())
 			.ToList();
 
-		var redirects = db.WikiRedirects.Select(entry => entry.PageNameFrom).ToHashSet();
+		var redirects = await db.WikiRedirects.Select(entry => entry.PageNameFrom).ToHashSetAsync();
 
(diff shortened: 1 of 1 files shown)


vadosnaprimer (approved):
Approving the concept, no idea about the code.

TASVideos/WikiModules/BrokenLinks.cshtml.cs
```diff
@@ -56,11 +56,14 @@ public async Task<IViewComponentResult> InvokeAsync()
 			.Select(c => c.Name.ToLowerInvariant())
 			.ToList();
 
+		var redirects = db.WikiRedirects.Select(entry => entry.PageNameFrom).ToHashSet();
```
adelikat: please use ToHashSetAsync()

YoshiRulz:
```suggestion
		var redirects = await db.WikiRedirects.Select(entry => entry.PageNameFrom).ToHashSetAsync();
```
?
If I don't `await` here and store the `Task` in the local, I can't `await` it later because it's read in a non-`async` lambda.

adelikat (approved):

PR #2368, opened 2026-07-22
Fixes Publication Edit and Publish Request not using YouTube title formats

calvarado194:
Issue was detected in several recent publications where the YouTube video was not launched with its expected format.


TASVideos/Pages/Publications/Urls/Edit.cshtml.cs
```diff
@@ -71,15 +71,6 @@ public async Task<IActionResult> OnPost()
 	{
 		var publication = await db.Publications
```
adelikat, line 72: Did you test this code path? Generally you need to .Include() a number of subtables from publications in order to call GenerateTitle successfully with the correct values, there's an extension method that will include the needed tables.  This either crashes or generates values with missing data, i belive

calvarado194: I ran the existing test suite and all tests succeeded. Should I run coverage to make sure this specific path is being tested?

adelikat: When querying a publication object that .GenerateTitle() is called on, you must use .IncludeTitleTables()

or else it will crash (or worse, silently fail) during the title generation.

adelikat (changes requested):
requesting changes, to 
use  .IncludeTitleTables(), see comment

adelikat (approved):

PR #2373, opened 2026-08-01
Hint bots to not download UserFiles

YoshiRulz:
Masterjun said this is NOT the cause for the recent downtime events, but it's still a waste of resources.


Masterjun3:
I don't know if we want this or not, but I wanted to point out that the robots.txt change also disallows bots downloading submission files, publication files, and additional publication files. Those all use the same handler.

adelikat:
I think I specifically want all of those files to not be crawled.  We could also add the nofollow to the pub and sub handler links

adelikat (approved):

PR #2374, opened 2026-08-02
Trim spaces in `[user: UserName]` links

YoshiRulz:
This was harmless when rendering wiki pages (since `/Users/Profile/%20UserName` resolves for some reason), but broke RSS.

### DIFF
--- a/TASVideos.WikiEngine/MakeBracketed.cs
+++ b/TASVideos.WikiEngine/MakeBracketed.cs
@@ -132,7 +132,7 @@ public static partial class Builtins
 
 		if (text.StartsWith("user:"))
 		{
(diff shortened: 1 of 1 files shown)


adelikat:
what is an example of a link that is out in the wild that breaks RSS?

My concern is that maybe users have trailing spaces in their name, would those linsk break (you can't register a name like that, but you could in the old system so we may have some, I can't remember if that's true)

YoshiRulz:
The publications feed. See convo at https://discord.com/channels/280806848909541376/738421558783246467/1533191201610535203.

IIRC *you* manually cleaned up the last few grandfathered usernames.

adelikat (approved):

PR #2386, opened 2026-09-10
Add reactions to forum posts

Masterjun3:
This PR adds the usual reactions feature of modern social media.

- A user can react with any one of these: 👍👎😄🎉😕❤️🚀👀. (The same as GitHub reactions.)
- No moderation feature as of now.
- Reactions are not anonymous, and can be viewed on `/Forum/Posts/1234/Reactions`, but there is currently no direct link to that url.
- Posts by TASVideoAgent and TASVideosGrue can't be reacted to.
- Only supported with JavaScript enabled. It makes a server call, *and replaces the HTML with the returned HTML from the server*, for reactivity. This is dangerous so we need to be careful that this works as expected. Although since it's just a server request, I believe this is safe.

Images:
image
After clicking on the smiley icon:
image
image

### DIFF
--- a/TASVideos.Common/Constants.cs
+++ b/TASVideos.Common/Constants.cs
@@ -82,6 +82,8 @@ public static class SiteGlobalConstants
 	public const int YearsOfBanDisplayedAsIndefinite = 2;
 
 	public const string MainIvatarDomain = "seccdn.libravatar.org"; // keep in sync with profile-settings.js
+
+	public static readonly string[] AllowedReactions = ["👍", "👎", "😄", "😕", "❤️", "🎉", "🚀", "👀"];
 }
 
 public static class ForumConstants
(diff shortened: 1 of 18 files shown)


vadosnaprimer (approved):

Masterjun3:
Ah well, everything is permission-based, so we always have the option to remove the permission from the user, and then they can't react anymore.

I always wonder, like, how frequently in the 8 years since our Discord server existed did mods have to remove specific reactions?
I know it's pointless to argue against moderation features, as it's always useful to have more. It just feels exhausting when it's expected to put hours into a feature that will barely be used, "just in case". And then you compare those hours with just deleting entries in the database by hand on a case-by-case basis.

Maybe that's why nobody wants to add new features ever.

Edit: I think I have a good idea for a minimal but good enough solution, stay tuned for commits!
Edit2, Note to self: Lock reactions on locked topics.
Edit3: Converting this to a draft. I'm going to finish this within the next 2 weeks, maybe even 1 week.

Masterjun3:
I updated the opening post with the changes I made. This PR is now ready.
Summary:
- The moderation feature. Every reaction can be deleted on the page that shows them all.
- Locked topics also have locked reactions (except for moderation).
- Removed the thing where you couldn't react on TASVideoAgent and TASVideosGrue posts.
- Added a few links here and there to show the reactions (by user, or by post).

